PR #371, opened 2012-11-29
Implementation of activate.bat portability

juntalis:
I saw that there was [another pull request implementing similar changes](https://github.com/pypa/virtualenv/pull/370/files), so I figured there wouldn't be any harm in putting forward my own. I think this implementation is a bit more beneficial, since it effectively results in the same environment variables as the current script, and because it doesn't rely on the use of temporary files or make any assumptions on what environment variables are currently set for the end user.


juntalis:
Cleaning up my stagnant pull requests.

PR #384, opened 2013-01-13
Add control PYTHONPATH in virtualenv

rodrigopmatias:
I'm write the control in activate.sh for PYTHONPATH this variable is usede for set the site-packages in user enviroment, this impact in virtualenv.


mohanraj-r:
I still find this as one of the frustrating points of using a virtualenv. Doc says that now `--no-site-packages` is redundant as its the default behavior 

> Not having access to global site-packages is now the default behavior.

But it does not mention anything about `PYTHONPATH` and the how its value affects this behavior. I wish  the effect of `--no-site-packages` would be to `unset PYTHONPATH` as that would be the correct behavior. This has caused of a lot of pain for me and by the looks of posts in stack overflow for others as well. 

How can we solve this?

dholth:
It's weird to have site-packages in PYTHONPATH. Normally Python can find its own site-packages automatically.

Many Python developers do not use PYTHONPATH at all. Instead, put a .pth file in the site-packages directory of the Python installation that you intend to use. Here is a blog post: http://bob.ippoli.to/archives/2005/02/06/using-pth-files-for-python-development/

Ivoz:
@mohanraj-r `PYTHONPATH` is a 'global' environment variable that adds to any and every python's path; not just a virtualenv's, or a global one. As such it is an independent thing from a virtualenv'ed python.

PR #433, opened 2013-06-13
Prevent accidental execution instead of sourcing

alanfranz:
This prevents a very common issue with newbies - trying to execute "activate" instead of sourcing it and thinks some magic would just happen.


BrownTruck:
Hello!

```
This prevents a very common issue with newbies - trying to execute "activate" instead of sourcing it and thinks some magic would just happen.

---

*This was migrated from pypa/virtualenv#433 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

alanfranz:
I think the idea is still sound, but I don't have the time to reenable the PR for all the shells, so I'll just close it.

PR #588, opened 2014-03-31
Add support for non-standard directory layouts

DragonSA:
# Context
On FreeBSD, pypy has been modified to follow the same layout as python, so as to better facilitate the installation of python packages on different python implementations (without having to rework the underlying ports).  

Since virtualenv hard codes the directory layout, this breaks virtualenv under pypy on FreeBSD, and on any python implementation that does not conform to virtualenv's assumptions about directory layout.  

# Changes
This pull request removes the hard codings and instead replaces them with paths provided by sysconfig.  sysconfig was used in favour of distutils.sysconfig as the later does not provide an interface for getting 'script' or 'bin' paths.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -22,6 +22,7 @@ from distutils.util import strtobool
 import struct
 import subprocess
 import tarfile
+import sysconfig
 
 if sys.version_info < (2, 6):
     print('ERROR: %s' % sys.exc_info()[1])
@@ -110,7 +111,7 @@ REQUIRED_MODULES = ['os', 'posix', 'posixpath', 'nt', 'ntpath', 'genericpath',
                     'fnmatch', 'locale', 'encodings', 'codecs',
                     'stat', 'UserDict', 'readline', 'copy_reg', 'types',
                     're', 'sre', 'sre_parse', 'sre_constants', 'sre_compile',
-                    'zlib']
+                    'sysconfig', 'zlib']
 
 REQUIRED_FILES = ['lib-dynload', 'config']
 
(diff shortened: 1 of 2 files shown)


koobs:
@DragonSA If this turns out to need any improvements in FreeBSD's Python or virtualenv packaging, let me know

DragonSA:
Thanks koobs.  Alas I never found the time to fix the patch.  Still on my TODO list.  I don't think we'll get much issue until FreeBSD starts supporting pypy in the bsd.python.mk framework.

koobs:
@DragonSA Re-base patch for master? EPROGRESS

DragonSA:
In summary: ENOPROGRESS

On FreeBSD we don't release pypy using it's release script.  The reasons are:
 * Supporting building multiple pypy versions within one port (i.e. a normal pypy, sandboxed pypy, etc) - not used except for those who do a custom build
 * Allow pypy to fit into a standard $PREFIX structure (i.e. in bin/ include/ and lib/)
 * Tentative plans to make switching from cPython to PyPy as a port's "python" implementation more seamless (i.e. just change pointers to bin, include, lib) [although this change may not be so required for that]

To fix we need to either:
 1) Change the way we install PyPy on FreeBSD, or
 2) Change virtualenv to automatically detect the library paths

This patch attempts to do the later however it wasn't complete and introduced an error.  I no longer have time to work on it.

bartekrutkowski:
Is there anything happening with this bug? This makes the virtualenv unusable with any Pypy version on FreeBSD (and, I assume, other systems that are installing Pypy similar way to Python).

DragonSA:
I now think there are two problems here: PyPy on FreeBSD doesn't install as specified by upstream (i.e we shouldn't be mangling the paths - my fault) and virtualenv should be autodetecting these paths.  I plan to fix PyPy on FreeBSD, and if time permitting, virtualenv, but my time is limited :-(.

DragonSA:
This has been fixed in FreeBSD's pypy implementation.

PR #589, opened 2014-04-02
Cleaned up activate.fish.

jonathanmarvens:
Cleaned up __activate.fish__ to be more *idiomatic* fish and removed a bit of unnecessary code (such as support for "Aspen magic directories", since *AFAIK* and *[AFAICT](http://aspen.io/virtual-paths 'Aspen: Python Web Framework with Simplates')*, this is no longer the case).

\- Jonathan


Ivoz:
I've applied this in develop.

Ivoz:
@jonathanmarvens turns out removing the double quotes for variables was a really bad thing. They prevent some parameter expansion that can crash the script. Hope you dogfooded your changes before submitting them at the time :/

PR #611, opened 2014-05-18
Fixed file paths with spaces causing pip to choke on *nix systems

blakfeld:
Spaces in the file path causes virtualenv pip to choke. I added a bit to re-write the shebang if the system is not windows, and the path contains a space. I basically just modified the "fixup_scripts" function.

Tested locally, seems to work pretty well :)

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -992,6 +992,10 @@ def create_environment(home_dir, site_packages=False, clear=False,
             to_install.append('pip')
         install_wheel(to_install, py_executable, search_dirs)
 
+    if not is_win:
+        if ' ' in os.path.abspath(bin_dir):
+            fix_path_whitespace(home_dir, bin_dir)
+
     install_activate(home_dir, bin_dir, prompt)
 
 def is_executable_file(fpath):
@@ -1497,6 +1501,10 @@ def install_activate(home_dir, bin_dir, prompt=None):
         files['activate'] = ACTIVATE_SH.replace('__VIRTUAL_ENV__', home_dir_sh)
 
     else:
+
+        # On non-Windows boxes, escape any spaces
+        home_dir = home_dir.replace(' ', '\ ')
+
         files = {'activate': ACTIVATE_SH}
 
         # suppling activate.fish in addition to, not instead of, the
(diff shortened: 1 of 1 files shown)


kevinmarks:
How has this bug from 2011 been allowed to persist for so long? I have lost hours debugging this on Mac OS X

ludovicriffault:
Any updates on this please? I still have the problem :(

blakfeld:
Sadly I don't think they have any intention of merging this or fixing the problem.

pfmoore:
I'm not a Unix user but my understanding is that not all Unix systems support the sort of quoting you're doing. (This has been debated at length on at least one other PR). If this PR is to have any chance (and it won't be me that commits it - as I say I don't have the Unix knowledge) it'll need to be confirmed as a correct fix on various Unix systems

blakfeld:
Hmm, maybe, I'll have to go hunt that down. All I'm doing is escaping the space with a backslash. That should be supported by all the major shells (sh, bash, and zsh for sure).

pfmoore:
Shebang lines are supported by the kernel, not by the shell. And the kernel can be pretty simple.

I believe there is one Unix flavour that insists on hash, bang, no more than 32 arbitrary characters, newline. And then uses the characters between ```#!``` and the newline as a completely uninterpreted filename (so backslashes are treated as actual backslashes, spaces as actual spaces, etc). On that version, no form of excaping is needed, but equally no form works. Compare that to other versions which allow arguments and terminate the executable name with a space - some of those have escaping conventions. There is no valid way of satisfying both of those sets of rules simultaneously.

As I say, my knowledge is limited, so I wouldn't take the above as guaranteed - but I believe that's the gist of the problem (basically it's insoluble in a platform-independent way).

blakfeld:
Oh awesome, I had no idea that was handled that far down the stack. Thank you for taking the time to type that up, that was incredibly informative!

ludovicriffault:
Thanks for answers! Any of you know a workaround to get this fixed? I really need to use pip with virtualenv in a directory with whitespace.

pfmoore:
@ludovicriffault You can probably hand-edit the shebang line on the scripts to a form that your local system handles - it's not ideal but it should work.

BrownTruck:
Hello!

```
Spaces in the file path causes virtualenv pip to choke. I added a bit to re-write the shebang if the system is not windows, and the path contains a space. I basically just modified the "fixup_scripts" function.

Tested locally, seems to work pretty well :)

---

*This was migrated from pypa/virtualenv#611 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

BrownTruck:
This Pull Request has been automatically migrated to pypa/virtualenv#910 to reparent it to the ``master`` branch. Please see the new pull request for any new discussion.

PR #634, opened 2014-08-01
A collection of fixes for when prefix != exec_prefix.

shidesthewise:
Patch version from Hans Meine from the duistutils mail list.
https://mail.python.org/pipermail/distutils-sig/2008-April/009336.html
http://mail.python.org/pipermail/distutils-sig/attachments/20080414/16fb135e/attachment-0001.diff

Add copy_inc_files function to fix issue of lost pyconfig.h.

When sys.exec_prefix and sys.prefix differ the pyconfig.h file
is in $(exec_prefix)/include/python$(version) and that file is
missing in the virtualenv because env/include/python$(version)
is a symlink to $(real_prefix)/include/python$(version).
If --always-copy is used then the two include directories merge.
I added copy_inc_files to symlink (or copy) the individual contents
of prefix include and exec_prefix include into env/include.

Update expected output to match actual output.

Setuptools and pip install line has been on one line instead of two
for awhile.

Add link from test to tests so python setup.py test work as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,9 @@ import re
 import shutil
 import sys
 
+if sys.version_info[:2] < (2, 6):
+    sys.exit('virtualenv requires Python 2.6 or higher.')
+
 try:
     from setuptools import setup
     from setuptools.command.test import test as TestCommand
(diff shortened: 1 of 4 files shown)


shidesthewise:
Because this is old and based on 1.11 and virtualenv is now in a 13.X series, I'll re-submit.

PR #638, opened 2014-08-06
Split out this setup.py fix from https://github.com/pypa/virtualenv/pull/635

shidesthewise:
Fix default args for py.test to be "tests" instead of using the
command name "test" also as the directory name to find tests.


shidesthewise:
Because this is old and based on 1.11 and virtualenv is now in a 13.X series, I'll re-submit.

PR #641, opened 2014-08-18
resolve PATHEXT on Windows

### LINKED ISSUE #640: The executable python does not exist
Running `virtualenv myvenv --python=python` fails with "The executable python (from --python=python) does not exist" even if python is on the PATH:

```
$ which python
/c/Python27/python
```

The issue is wrong executable detection on Windows. `python` is equivalent to `python.exe` on Windows but this is not recognized by virtualenv. It works only if you use `virtualenv myvenv --python=python.exe`.

schlamar:
This fixes #640 for me.

It works now with `virtualenv --python=C:\Python27\python` and `virtualenv --python=python` on Windows.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1580,6 +1580,20 @@ def fix_lib64(lib_dir, symlink=True):
         else:
             copyfile('lib', lib64_link)
 
+def resolve_exe(exe):
+    if os.path.exists(exe):
+        return exe
+
+    if os.name != 'nt':
+        return None
+
+    # check PATHEXT
+    exts = os.environ['PATHEXT'].split(os.pathsep)
+    for ext in exts:
+        if os.path.exists('%s%s' % (exe, ext)):
+            return '%s%s' % (exe, ext)
+    return None
+
 def resolve_interpreter(exe):
     """
     If the executable given isn't an absolute path, search $PATH for the interpreter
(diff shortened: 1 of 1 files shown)


virtualenv.py
```diff
@@ -1593,10 +1607,12 @@ def resolve_interpreter(exe):
...
-            if os.path.exists(os.path.join(path, exe)):
-                exe = os.path.join(path, exe)
+            check_exe = resolve_exe(os.path.join(path, exe))
+            if check_exe:
+                exe = check_exe
                 break
-    if not os.path.exists(exe):
+    exe = resolve_exe(exe)
+    if not exe:
         logger.fatal('The executable %s (from --python=%s) does not exist' % (exe, exe))
```
pfmoore: By overwriting ```exe``` here you've made the error message misleading (as it no longer states the same argument as the user supplied).

virtualenv.py
```diff
@@ -1580,6 +1580,20 @@ def fix_lib64(lib_dir, symlink=True):
...
 
+def resolve_exe(exe):
+    if os.path.exists(exe):
+        return exe
+
+    if os.name != 'nt':
+        return None
+
+    # check PATHEXT
+    exts = os.environ['PATHEXT'].split(os.pathsep)
```
pfmoore: Do we really want to use ```PATHEXT``` here? If what you've found isn't ```python.exe``` it's unlikely to actually be a valid Python installation, after all...

BrownTruck:
Hello!

```
This fixes #640 for me.

It works now with virtualenv --python=C:\Python27\python and virtualenv --python=python on Windows.

---

*This was migrated from pypa/virtualenv#641 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

BrownTruck:

Issue #92, opened 2011-03-15
Add option to always copy instead of symlink


carljm:
Originally reported by Roland Schulz at https://bugs.launchpad.net/virtualenv/+bug/442780

It would be nice to have an option -copy or -nosymlink to force coping instead of symlinking. It can be very easily be done by just:
delattr(os,"symlink")

I wanted to create a virtual env in a network mounted directory. Thus
symlinks don't work because they point to the file on the computer where one
creates the virtual env which are not necessary available on other
machines/nodes.

PR #663, opened 2014-10-29
Autodetect if symlinks are supported on target file system

techtonik:
This detects symlink suppot on destination filesystem and not only in Python interpreter or OS. If symlinks are not supported, falls back to copy behavior.

This fixes some case reported in #92 and #409, without specifying `--always-copy` explicitly.

Ideal case would be to invert `virtualenv` symlinking behavior - symlink only when explicitly requested.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1100,6 +1100,27 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
         ## Maybe it should delete everything with #!/path/to/venv/python in it
         logger.notify('Not deleting %s', bin_dir)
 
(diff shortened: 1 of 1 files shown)


BrownTruck:
Hello!

```
This detects symlink suppot on destination filesystem and not only in Python interpreter or OS. If symlinks are not supported, falls back to copy behavior.

This fixes some case reported in #92 and #409, without specifying --always-copy explicitly.

Ideal case would be to invert virtualenv symlinking behavior - symlink only when explicitly requested.

---

*This was migrated from pypa/virtualenv#663 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

PR #699, opened 2015-01-05
[Rewrite] Support for virtualenv in legacy virtualenv

ionelmc:
Notes:

* 2.6 support is broken cause virtualenv and pip don't seem to work with `python2.6 -m ...`
* Now global site paths are generated by the Flavor. If this is too fragile it remains to be seen. PyPy is still not supported.
* I need to add more integration tests.


ionelmc:
There is need to refactor the Flavors a bit. I'm looking into doing that after getting this in. Currently I'm considering to have 3 dimensions (in contrast to the current 2-dimensional builder-flavor design):
* OS-dimension: Windows, Linux, OSX
* Target-dimension: CPython, PyPy, Jython or IronPython - the "Target" would interface with the target interpreter (replacing the `_get_base_python_info`) and do _some_ of the file layout stuff.
* BootstrapTechnique-dimension: Legacy or *3.3-venv*

ionelmc:
Also, 3.2 support is also broken. No one wants support for this right?

ionelmc:
Gonna rework this a bit

ionelmc:
See #700

PR #700, opened 2015-01-05
Rewrite virtualenv in legacy virtualenv resolve bin

ionelmc:
This one is simpler, resolves the python bin prior to getting any `info` about it.


ionelmc:
See #702.

PR #702, opened 2015-01-06
[Rewrite] Support for virtualenv in legacy virtualenv via `resolve bin`

ionelmc:


ionelmc:
This PR adds support for 2.6 and copies directories (like `include` or packages) from the source python. This makes other supporting other runtimes possible down the road and fixes few issues caused by the lack of bootstrap packages (like distutils) and include dirs.

virtualenv/builders/legacy.py
```diff
@@ -114,8 +115,8 @@ def _get_base_python_info(self):
         # Get information from the base python that we need in order to create
         # a legacy virtual environment.
         return json.loads(
-            subprocess.check_output([
-                self.python,
+            check_output([
+                self._get_base_python_bin(),
```
pfmoore: So, you're running a subprocess in ```_get_base_python_bin``` to get the argument for a subprocess call here. On Windows, the subprocess call overhead is significant - have you tested the performance of this against the current virtualenv?

I'm not saying I know of any *other* way to do it, but if there's a significant performance regression, we should make it clear. After all, virtualenv on Windows is already a bit slow (nothing like how it was before we used wheels, but that's the point - I'd hate to lose that gain).

virtualenv/flavors/posix.py
```diff
@@ -10,20 +10,32 @@ class PosixFlavor(BaseFlavor):
...
     @property
     def activation_scripts(self):
         return set(["activate.sh", "activate.fish", "activate.csh"])
 
-    def python_bins(self, version_info):
+    def python_bins(self, base_python):
+        version_info = base_python["sys.version_info"]
         return [
-            "python{0}".format(".".join(map(str, version_info[:i])))
+            "python%s" % ".".join(map(str, version_info[:i]))
```
pfmoore: Why change from using ```format()```? Seems like an unrelated (and unneeded) change.

virtualenv/flavors/posix.py
```diff
@@ -10,20 +10,32 @@ class PosixFlavor(BaseFlavor):
...
         ]
 
-    def lib_dir(self, version_info):
+    def lib_dir(self, base_python):
         return posixpath.join(
             "lib",
-            "python{0}".format(
-                ".".join(map(str, version_info[:2]))
-            ),
+            "python{0}.{1}".format(*base_python["sys.version_info"])
```
pfmoore: ... particularly as you still use ```format()``` here.

virtualenv/builders/legacy.py
ionelmc: I suppose we could special case this later on for Windows to avoid the secondary subprocess, but as it is now it's badly needed cause there's no reliable way to figure out where the bin really is without running it.

In my tests there wasn't noticeable slowdown.

pfmoore:
Added some basic comments. I would prefer if 2.6 support and directory copying were separate PRs, as I can't say what parts of this PR relate to which aspect (and I'm not too familiar with what's needed for 2.6 support, so it's harder for me to review that aspect).

Having said that, much of the code seems to be about getting information directly from the "base Python". I don't really see how that relates to *either* of the above goals. Maybe all of the ```_get_base_python_bin``` could be its own PR, "get more accurate information about the base Python"? (If I've misunderstood the purpose of that function, well that's sort of my point, it could do with being separated out into something with a clearer, more specific, purpose).

virtualenv/builders/base.py
```diff
@@ -44,7 +50,35 @@ def __init__(self, python, flavor, system_site_packages=False,
     def check_available(cls, python):
         raise NotImplementedError
 
+    def _get_base_python_bin(self):
```
pfmoore: This seems to be called multiple times. You should cache the result as a subprocess call is a huge overhead.

ionelmc: Glad you asked :grimacing: - this is done in the final PR.

virtualenv/builders/base.py
```diff
@@ -44,7 +50,35 @@ def __init__(self, python, flavor, system_site_packages=False,
...
+                    from distutils import sysconfig
+
+                if (sys.platform.startswith("win") or sys.platform == "cli" and os.name == "nt"):
+                    bindir = getattr(sys, "real_prefix", sys.prefix)
+                else:
+                    bindir = sysconfig.get_config_var("BINDIR")
+
+                print(json.dumps(bindir))
+                """)
+            ]).decode(locale.getpreferredencoding()),
```
pfmoore: Will this work on Windows, where getpreferredencoding could well be something like CP850, but Python could be installed in a directory with Unicode characters? Especially in Python 2.5, where per-user installs could be under the user's home directory (and so will be more likely to contain non-ASCII characters). I'd suggest encoding the JSON data in such a way that it is pure-ASCII before sending it to the caller, and then you can decode it back to Unicode on the receiving end.

ionelmc:
@pfmoore `_get_base_python_bin`'s purpose if basically busting out of a virtualenv. Without that you cannot create a virtualenv from a virtualenv cause the paths (sys.prefix, sys.exec_prefix) would be wrong. 

`_get_base_python_bin` will resolve the target to the real bin, thus, when the interpreter is probed for info (`_get_base_python_info`) it will have accurate information.

virtualenv/builders/base.py
pfmoore: Well, I'm not reviewing that PR... ;-) I was assuming that the expectation was that *this* PR would be merged, not that the comments would be handled in the "big" PR. Is that not how you see it?

But regardless, glad that you had it covered. Ta.

pfmoore:
OK, cool. That was about what I thought (I'd written similar code in the past myself, so it seemed familiar). Adding ```_get_base_python_bin``` seems like a great candidate for a standalone PR, in that case. Much better that way than lumping it in with 2.6 support.

virtualenv/builders/base.py
ionelmc: You're probably right. My plan is to add integration tests with non-ascii paths later on, we'll see this, and many other places having failures.

gaborbernat:
I'll close this now as feels hard to know what still needs or not. If you would like anything kept feel free to open new PR.

PR #703, opened 2015-01-06
Drop support for Python 2.6

dstufft:


dstufft:
Bleh, I don't think we can actually do this reasonably without dropping support for 2.6 in pip too.

PR #722, opened 2015-03-02
activate.sh: drop $_OLD_VIRTUAL_PATH, use remove_from_path instead

blueyed:
Using `$_OLD_VIRTUAL_PATH` will overwrite any changes to $PATH inbetween
activating and deactivating the virtualenv.  It also does not behave
well with `exec $SHELL` (reloading your shell).

`remove_from_path` is taken from pyenv/rbenv's pyenv-which/rbenv-which
shim (https://github.com/sstephenson/rbenv/blob/e851250/libexec/rbenv-which#L18-L27).


Ivoz:
@blueyed I like the idea of this, however concerned that this might not be portable shell code? We don't have the luxury of demanding bash.

BrownTruck:
Hello!

```
Using $_OLD_VIRTUAL_PATH will overwrite any changes to $PATH inbetween
activating and deactivating the virtualenv.  It also does not behave
well with exec $SHELL (reloading your shell).

remove_from_path is taken from pyenv/rbenv's pyenv-which/rbenv-which
shim (https://github.com/sstephenson/rbenv/blob/e851250/libexec/rbenv-which#L18-L27).

---

*This was migrated from pypa/virtualenv#722 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

blueyed:
@BrownTruck / @dstufft
Before rebasing/resubmitting this I'd like to have some feedback about if this is feasible / useful, and would be considered to be merged after all.

@Ivoz 
I could investigate to make this POSIX compatible.

BrownTruck:

Issue #407, opened 2013-03-13
PS1 isn't reset if empty


shish:
If one activates a virtualenv while PS1 is empty (eg a non-login shell), then deactivates it, the environment name will be left in the prompt. This can then cause lots of confusion as scripts assume "PS1 is non-empty" == "interactive shell"

```
$ export PS1=
. myEnv/bin/activate
(myEnv) deactivate
(myEnv) # env name is still here after deactivate
```

```
$ export PS1=moo
moo . myEnv/bin/activate
(myEnv)moo deactivate
moo # env name is gone
```

PR #723, opened 2015-03-02
activate.sh: handle empty (but set) vars

blueyed:
Fixes https://github.com/pypa/virtualenv/issues/407


Ivoz:
Cheers, this wasn't a clean merge, so I applied it manually.

Issue #365, opened 2012-11-26
bin/activate doesn't remove alias after running deactivate


Mumage:
If you source bin/activate, then run deactivate, the alias 'python -m pydoc' persists.  In the interest of leaving everything exactly as it was before sourcing bin/activate, perhaps it would be good to unalias the alias as part of the deactivate function?

blueyed:
Fixed in #724 - it should have been `unalias` instead of `unset` in `deactivate`.

PR #724, opened 2015-03-02
activate.sh: use unalias for unsetting the pydoc alias

blueyed:
Fixes https://github.com/pypa/virtualenv/issues/365

### DIFF
--- a/virtualenv_embedded/activate.sh
+++ b/virtualenv_embedded/activate.sh
@@ -2,7 +2,7 @@
 # you cannot run it directly
 
 deactivate () {
-    unset pydoc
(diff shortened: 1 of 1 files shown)


Ivoz:
Turns out putting `unaliases` is a bad idea, I've changed `pydoc` to be a short bash function that can be unset.

blueyed:
@Ivoz 
> Turns out putting `unaliases` is a bad idea

Why?

Ivoz:
we are calling `deactivate` in the script, which will try to unalias `pydoc`, before the `pydoc` alias is defined.

PR #731, opened 2015-03-16
Upgrade setuptools to 14.3

dstufft:


dstufft:
Error already exists, unrelated to this PR.

PR #747, opened 2015-04-21
Fixed pip/setuptools installation on Jython

agronholm:
This clears one roadblock from getting virtualenv to work with Jython. There are more, but they should probably be fixed on Jython's side.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -955,6 +955,7 @@ def install_wheel(project_names, py_executable, search_dirs=None):
         call_subprocess(cmd, show_stdout=False,
             extra_env = {
                 'PYTHONPATH': pythonpath,
+                'JYTHONPATH': pythonpath,  # for Jython < 3.x
                 'PIP_FIND_LINKS': findlinks,
                 'PIP_USE_WHEEL': '1',
                 'PIP_PRE': '1',


rbtcollins:
Jython doesn't honour the PYTHONPATH variable? :(. Well it is what it is, but I think we need a comment explaining when we can remove the workaround...

agronholm:
There already is -- I explicitly mentioned Jython 2.x. I just committed a change in the Jython 3 repository that makes Jython use PYTHONPATH instead.

virtualenv.py
```diff
@@ -955,6 +955,7 @@ def install_wheel(project_names, py_executable, search_dirs=None):
         call_subprocess(cmd, show_stdout=False,
             extra_env = {
                 'PYTHONPATH': pythonpath,
+                'JYTHONPATH': pythonpath,  # for Jython 2.x
```
rbtcollins: Cool - so here please say 

 # For Jython < 3.x

instead. Thanks!

agronholm:
I've updated the PR.

agronholm:
So where do we stand with this?

agronholm:
This is the ONLY thing remaining that prevents virtualenv from working properly on Jython 2.7. Please merge this now!

rbtcollins:
So, it looked fine with that tweak to me, but I'm not a committer - sorry.

PR #752, opened 2015-05-07
Fix the document structure

ztane:
PyPI/docutils give up on bad structure here: https://pypi.python.org/pypi/virtualenv/12.1.1


Ivoz:
Cheers, I've applied this to everything in the most modern changes.rst

PR #778, opened 2015-07-24
Fixed fish_prompt not displaying $status

Lucretiel:
Many `fish_prompt`s display the `$status` of the previous command (`$status` being fish's equivalent of `$?` in POSIX shells). However, in order for this to work correctly, `$status` must be retrieved right away; otherwise, the `$status` will be of some command within the `fish_prompt` function. This pull request updates an activated virtualenv's `fish_prompt` to call `_old_fish_prompt` first, and store the resultant prompt string in a local variable, to ensure that `$status` is captured correctly.

### DIFF
--- a/virtualenv_embedded/activate.fish
+++ b/virtualenv_embedded/activate.fish
@@ -51,10 +51,12 @@ if test -z "$VIRTUAL_ENV_DISABLE_PROMPT"
     
     # with the original prompt function copied, we can override with our own.
     function fish_prompt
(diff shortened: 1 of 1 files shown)


Ivoz:
Hi there lucretiel, I've recently applied another change to `activate.fish` which has been cleaned up a little. I'd appreciate it if you'd be able to write this patch on top of the new version of the file.

Lucretiel:
Sounds good. I thought of a better way to implement part of it so I'll add that update as well.

Issue #169, opened 2011-08-30
python-config should be linked


lunaryorn:
virtualenv should provide `pythonX.Y-config` of the python interpreter used for the virtual environment as `python-config` in the virtual environment.

Usage case:  Building of extension modules inside virtualenvs.  Some extension modules use autotools as build system.  Autotools uses python-config to find the extension module.  This fails, if the "python" inside the virtualenv is Python 3, but "python-config" coming from the system Python is Python 2.

bchretien:
:+1: This can be handy to find the proper include directory containing `Python.h`.

wookayin:
+1
This is needed, to deal with annoying troubles on building (native) extensions.

mentalbrew:
+1 also. This issue dates to 2011. Any idea if this is going to be addressed?

StefanD986:
I posted this on the `python-ideas`  mailinglist as possible feature for the standard library `venv` module.

@lunaryorn , @madprog @josharian @hoangelos @wiseman @offbyone @mentalbrew @bchretien @hoodie @wookayin and anyone who's stumbling on this issue:
If this is still relevant for you, please give it a *thumbs up* in the discussion here (You can log in with your github account): https://mail.python.org/archives/list/python-ideas@python.org/thread/QTCPOM5YBOKCWWNPDP7Z4QL2K6OWGSHL/

PR #783, opened 2015-08-16
Adding python-config script to virtualenv bin dir

madprog:
Should solve most of pypa/virtualenv#169


amchoukir:
I can confirm that this patch fixes issues building [Valloric/ycmd](https://github.com/Valloric/ycmd)

Ivoz:
It seems you've forgotten to add/commit the file `python-config` into git with this PR.

madprog:
Hello Matt,
The python-config file is gzipped and base64-ed, and inlined at the line 2239 of virtualenv.py, just like activate_this.py or distutils.cfg
Where should it be copied else?

Ivoz:
@madprog the source for all such files needs to live in [virtualenv_embedded](https://github.com/pypa/virtualenv/tree/develop/virtualenv_embedded). They should then be automatically included into `virtualenv.py` by running `bin/rebuild-script.py`.

madprog:
Here is the commit with the missing file.

Ivoz:
@madprog now I can see the contents of the file. It seems to me likely you copied this from a python 2 source. It uses the python 2 only print statement, so this is unsuitable to merge at the moment. Additionally, it doesn't cover the additional options available in a python3 `python-config`, you can see those [here for instance](https://github.com/python/cpython/blob/3.5/Misc/python-config.in).

madprog:
@Ivoz Indeed it was at first copied from a python2-config. At that time (end of 2011), python3 was not *that* mainstream ;)
Please now find the python-config that you linked, updated so that it also runs with python2.
I compared the outputs using this zsh script:
```bash
#!/usr/bin/zsh

for i in prefix exec-prefix includes libs cflags ldflags extension-suffix help abiflags configdir; do
    echo "\x1b[32m$i\x1b[0m"

    PYTHON2_CMD="diff -u <(python2-config --$i 2>&1) <(python2 python-config --$i 2>&1)"
    PYTHON3_CMD="diff -u <(python3-config --$i 2>&1) <(python3 python-config --$i 2>&1)"

    if ! $SHELL -c "$PYTHON2_CMD" >/dev/null; then
        echo "\x1b[31mpython2\x1b[0m"
        $SHELL -c "$PYTHON2_CMD"
    fi

    if ! $SHELL -c "$PYTHON3_CMD" >/dev/null; then
        echo "\x1b[31mpython3\x1b[0m"
        $SHELL -c "$PYTHON3_CMD"
    fi
done
```
They are not exactly the same on my system: the order of the flags is different (but similar), some may be duplicated, and new flags are supported for python2, but I think that it "will do". What do you think?

Ivoz:
@madprog I have applied this through #798, thanks for most of the work :)

PR #785, opened 2015-08-19
virtualenv fish shell bgcolor to black

AndreasHassing:
White on blue in fish shell is extremely unreadable - not sure if this is just the case for my particular installation, but this seems to look better.


Ivoz:
Hi there @AndreasHassing , you never supplied the source file of the change (in `virtualenv_embedded/activate.fish`)

AndreasHassing:
Oh yeah @Ivoz, thanks for pointing it out to me. I'll fix it once I'm back on the laptop and notify you.

Ivoz:
Also note `activate.fish` was updated recently, you may have to rebase changes.

Ivoz:
@AndreasHassing I changed the format to be hopefully more sensible in bcf6c085a8ac68c maybe that might suite the needs of this PR?

Ivoz:
I'm hoping the current code when released will look ok (better than last release). Thanks for suggesting it get improved and your contribution anyway. If you think it could still use help next release please do yell :)

PR #797, opened 2015-09-18
upgrade wheel to 0.25

davidszotten:
fixes compatibility issue with python 3.5. see
https://bitbucket.org/pypa/wheel/pull-requests/53


Ivoz:
Cheers mate, since wheel 0.26 I have upgraded to that instead. Should be out next release

jck:
pip fails to create and cache wheels for many packages in python 3.5 virtualenvs. please release a new version soon.

PR #806, opened 2015-10-10
Fix bad links to virtualenvwrapper in docs

willingc:
Fix two links to doug hellman's blog since current links are broken.


willingc:
Thanks @lvoz

Issue #53, opened 2011-03-14
Whitespace in root path of virtualenv breaks scripts


vbabiy:
I'm not really sure if this is a distribute/setuptools/virtualenv but,

If I install virtualenv in

**/var/lib/hudson/home/jobs/Minification WebHelpers/workspace/python/2.4**

then run ./bin/easy_install:

**bash: ./bin/easy_install: "/var/lib/hudson/home/jobs/Minification: bad interpreter: No such file or directory**

Seems like something does not obey whitespace in path names correctly.

---
- Bitbucket: https://bitbucket.org/ianb/virtualenv/issue/38
- Originally Reported By: Domen Kožar
- Originally Created At: 2010-05-07 14:57:32

diimdeep:
So annoying, should be fixed asap.

pfmoore:
See the link @Ivoz posted, this is a Unix limitation. #611 might work for some Unix variants, if they support backslash escapes in a shebang line, but it's not clear which versions do (and the code just blindly does it without checking - which admittedly won't make the problem *worse*, but won't help either if it's not supported...)

jashephe:
It's indeed true that this is a result of the way that unix handles shebang lines, but if #611 fixes the problem for some systems and doesn't worsen the problem for others, would that still be an improvement?

pfmoore:
If that's true then yes. But I can'rt comment on #611 as I'm not a Unix developer. There might be cases where it makes things worse, I just don't know. Sorry I can't be more help.

jashephe:
Fair enough.  #611 probably needs to be more carefully looked at and tested for fringe cases.

acidjunk:
Even worse: on windows it breaks on the default Jenkins path with the same error:
FATAL: whitespaces are not allowed in Python interpreter's path: C:\Program Files (x86)\Jenkins\shiningpanda\jobs\c3418983\virtualenvs\d41d8cd9

fbidu:
I was just affected by this issue. Following the instructions I found [on StackOverflow](http://stackoverflow.com/a/15475004), I managed to make `pip` work by just setting the first line to `#!/usr/bin/env python`

merwok:
Changing the shebang of installed scripts to “env python” means that they will work only in an activated virtualenv.  The scripts were generated with explicit absolute paths so that they would always use the Python in the venv, and thus find the installed packages needed by the scripts.

pfmoore:
[...]

PR #811, opened 2015-10-18
escape string in __VIRTUAL_ENV__ path for bash

davidbstein:
Fixes issue pypa/virtualenv#53 in bash and shell only. Uses the printf `%q` "shell quote" formatter. Falls back to quoting some characters that universally need to be escaped. 

Fallback is needed because while bash's `printf` command has the `%q` formatter, the builtin function printf may not.

(edit): removed irrelevant link


Ivoz:
Hi there, you need to actually commit changes for virtualenv_embedded/activate.sh as well.

Did you understand *not* to do so from reading the documentation somewhere? If so we need to clarify it.

davidbstein:
Hi Matt -

Completely my fault, have had a fix locally for a while and pushed it
upstream without checking contributing.rst - docs are quite clear. Sorry.

-David

On Sun, Oct 18, 2015 at 2:07 AM Matt Iversen <notifications@github.com>
wrote:

> Hi there, you need to actually commit changes for
> virtualenv_embedded/activate.sh as well.
>
> Did you understand *not* to do so from reading the documentation
> somewhere? If so we need to clarify it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pypa/virtualenv/pull/811#issuecomment-148994319>.
>
-- 

Sent from a phone.

Ivoz:
Hi @davidbstein upon looking further, I don't see how this solves #53 at all?

pfmoore:
Regardless of whether this PR does fix the problem, would it be worth adding a test that demonstrates the issue, so we'd at least be able to see the problem via a test?

davidbstein:
thanks for the quick review!

@Ivoz - AFAICT, #53 is caused by the bad entry in PATH, not the malformed shebang - easy_install and pip work again with this change  (testing on Ubuntu and OSX - escaping the PATH fixes the issue, escaping the shebang line does not). Fixing the shebang line seems like a good idea - I'll try to dig into a good way to do that.

@pfmoore - tests seem like a great idea. Will add one.

Ivoz:
@davidbstein so after manually applying your patch to test,

I created a new virtualenv called "h a", confirm the patch is in the new virtualenv:
```
(h\ a) ivo·ivosung ~/code/pypa/virtualenv/h a/bin develop $ cat activate | grep 'VIRTUAL_ENV='
# VIRTUAL_ENV="/home/ivo/code/pypa/virtualenv/h a"
VIRTUAL_ENV=$(printf %q "/home/ivo/code/pypa/virtualenv/h a") || VIRTUAL_ENV=$(echo "/home/ivo/code/pypa/virtualenv/h a" | sed 's/[ ()$;]/\\&/g')
```
Yes, we've managed to escape it in the `$PATH`:
```
(h\ a) ivo·ivosung ~/code/pypa/virtualenv/h a/bin develop $ echo $PATH
/home/ivo/code/pypa/virtualenv/h\ a/bin:/home/ivo/bin:/hom...
```
But, We can't escape the hashbang problem:
```
(h\ a) ivo·ivosung ~/code/pypa/virtualenv/h a/bin develop $ pip -h
/home/ivo/.pyenv/plugins/pyenv-pip-rehash/libexec/pip: ./pip: "/home/ivo/code/pypa/virtualenv/h\: bad interpreter: No such file or directory
 126 (h\ a) ivo·ivosung ~/code/pypa/virtualenv/h a/bin develop $ easy_install -h
/home/ivo/.pyenv/plugins/pyenv-pip-rehash/libexec/easy_install: ./easy_install: "/home/ivo/code/pypa/virtualenv/h: bad interpreter: No such file or directory
```

So we've put an escape in the `$VIRTUAL_ENV` env var & `$PATH`, but paths could be resolved without that anyway; and we haven't fixed execution of scripts as shown above. So we still haven't fixed scripts being broken, as is title of #53

Ivoz:
I'm pretty sure that escaping paths in `$PATH` is in fact *incorrect*.

I'm guessing that this is only working for you (or, *appearing* to work), because now that your patch with the backslash escape has made the virtualenv path invalid, when you type `$ pip`, `$ easy_install` etc you are finding working executable, but they are the global executables - we obviously don't want those. If you type `$ ./bin/pip` you will still find the command breaking from the space in the hashbang.

PR #827, opened 2015-12-03
Fix PATH in sh subshells in fish shell

ViViDboarder:
If in a subshell from within `fish`, the PATH variable will be converted to the proper format for sh, but none of the custom path-like variables will. This causes errors because it sets the PATH to an invalid value when activating.

This patch makes activate.fish always store the old path using `:` deliminators and then converts it back to `fish` format later.

### Issue reproduction
Simplest example:

`PATH` converts correctly
```
fish $ echo $PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $PATH'
  /bin:/usr/local/bin
```

`OLD_PATH` does not
```
fish $ set -x OLD_PATH $PATH
fish $ echo $OLD_PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $OLD_PATH'
  /bin/usr/local/bin
```

There you will see the `OLD_PATH` is not a valid path for sh.

### DIFF
--- a/virtualenv_embedded/activate.fish
+++ b/virtualenv_embedded/activate.fish
@@ -1,10 +1,23 @@
 # This file must be used using `. bin/activate.fish` *within a running fish ( http://fishshell.com ) session*.
 # Do not run it directly.
 
+function _bashify_path -d "Converts a fish path to something bash can recognize"
+    set fishy_path $argv
+    set bashy_path $fishy_path[1]
+    for path_part in $fishy_path[2..-1]
+        set bashy_path "$bashy_path:$path_part"
+    end
+    echo $bashy_path
+end
+
+function _fishify_path -d "Converts a bash path to something fish can recognize"
+    echo $argv | tr ':' ' '
+end
+
 function deactivate -d 'Exit virtualenv mode and return to the normal environment.'
     # reset old environment variables
     if test -n "$_OLD_VIRTUAL_PATH"
-        set -gx PATH $_OLD_VIRTUAL_PATH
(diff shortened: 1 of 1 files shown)


ViViDboarder:
This could likely be applied to the rewrite branch as well.

ViViDboarder:
I have updated this patch to localize all changes related to fish shell into the activate.fish instead of relying on other shell scripts to know anything about fish.

iest:
I ran into this issue as well — took a while and a lot of frustration to debug. Be great to get this merged 🎉

BrownTruck:
Hello!

```
If in a subshell from within fish, the PATH variable will be converted to the proper format for sh, but none of the custom path-like variables will. This causes errors because it sets the PATH to an invalid value when activating.

This patch makes activate.fish always store the old path using : deliminators and then converts it back to fish format later.

Issue reproduction

Simplest example:

PATH converts correctly

fish $ echo $PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $PATH'
  /bin:/usr/local/bin


OLD_PATH does not

fish $ set -x OLD_PATH $PATH
fish $ echo $OLD_PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $OLD_PATH'
  /bin/usr/local/bin


There you will see the OLD_PATH is not a valid path for sh.

---

*This was migrated from pypa/virtualenv#827 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*
```

ViViDboarder:
Closed after rebase

Issue #93, opened 2011-03-15
ensure tcl works within virtualenv


carljm:
Originally reported by Shawn Wheatley at https://bugs.launchpad.net/virtualenv/+bug/449537

Environment:
Tested on Windows XP SP3 and Windows 7
Python 2.6.3
Virtualenv 1.3.4

Within a fresh, just activated VirtualEnv, I get the following error trying to run a simple Tkinter window:

H:\My Documents\pythondev>virtualenv H:\testvirtualenv
New python executable in H:\testvirtualenv\Scripts\python.exe
Installing setuptools...................done.

H:\My Documents\pythondev>H:\testvirtualenv\Scripts\activate.bat
(testvirtualenv) H:\My Documents\pythondev>python
Python 2.6.3 (r263rc1:75186, Oct 2 2009, 20:40:30) [MSC v.1500 32 bit (Intel)]
on win32
Type "help", "copyright", "credits" or "license" for more information.

> > > import Tkinter
> > > Tkinter._test()
> > > Traceback (most recent call last):
> > >   File "<stdin>", line 1, in <module>
> > >   File "C:\Python26\Lib\lib-tk\Tkinter.py", line 3749, in _test
> > >     root = Tk()
> > >   File "C:\Python26\Lib\lib-tk\Tkinter.py", line 1643, in __init__
> > >     self.tk = _tkinter.create(screenName, baseName, className, interactive, want
> > > objects, useTk, sync, use)
> > > _tkinter.TclError: Can't find a usable init.tcl in the following directories:
> > >     C:/Python26/lib/tcl8.5 H:/testvirtualenv/lib/tcl8.5 H:/lib/tcl8.5 H:/testvir
> > > tualenv/library H:/library H:/tcl8.5.2/library H:/tcl8.5.2/library

This probably means that Tcl wasn't installed properly.

Shawn Wheatley wrote on 2009-10-13:

Virtualenv does not copy the Tcl library file(s) (or any core C extensions, that I can see) to the new virtualenv, but instead adds a reference to the PYTHONPATH. If I copy the "tcl" folder from C:\Python26\ over to the root of the new Virtualenv, Tkinter.Tk() shows a new window without throwing an exception.
Jannis Leidel wrote on 2009-10-13:  #2

Just suspecting here but could you create a virtualenv on the same drive as the system Python (C:)? I have the feeling it could be related to https://bugs.launchpad.net/virtualenv/+bug/352844.

[...]

PR #888, opened 2016-04-05
Fix issue #93 - Tcl doesn't work inside a virtualenv on Windows

techtonik:
Copies:
```
C:\Python27\tcl\tcl8.5
C:\Python27\tcl\tk8.5
```

```
C:\Python35\tcl\tcl8.6
C:\Python35\tcl\tk8.6
```

Checked that Python 3.4 contains the same dirs as 3.5.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1062,6 +1062,16 @@ def copy_required_modules(dst_prefix, symlink):
                 if os.path.exists(pyfile):
                     copyfile(pyfile, dst_filename[:-1], symlink)
 
+def copy_tcltk(src, dest, symlink):
+    """ copy tcl/tk libraries on Windows (issue #93) """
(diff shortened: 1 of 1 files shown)


yan12125:
Python 3.3.5 still ships with Tcl 8.5, so the Python 3.3 build on Windows failed.  [1] From CPython's log, Tcl 8.6.1 was introduced in Python 3.4.0 [2] But anyway I think relying on CPython version is fragile.

[1] https://ci.appveyor.com/project/yan12125/virtualenv/build/job/9h8crh9s4k7ag6pb
[2] https://hg.python.org/cpython/rev/730d89d66b38

techtonik:
@yan12125 I don't have time to test it - edited on GitHub, so can you take a look if it at least doesn't break anything?

PR #892, opened 2016-05-03
Check if py_executable is a symlink.  If not, continue to symlink it.

ewongbb:
Current behaviour of install_python() overwrites py_executable without checking if py_executable is a symlink itself (especially to one of the mentioned symlinks in required_symlinks).  Right now,
while building Thunderbird (which symlinks python to python2.7 and python2), this situation
creates a circular symlink issue because it removes the python and symlinks it to python2.7
without removing the python2.7 symlink (which is symlinked to python).

This patch at least checks if py_executable is a symlink.  If so, it doesn't do anything to
it. If not, it does the usual for loop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ branches:
     - master
     - develop
     - rewrite
+    - chksymlinks
 
 notifications:
   irc:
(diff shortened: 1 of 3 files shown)


ewongbb:
wip (obviously)

ewongbb:
I believe I need to rethink this approach before making this pr.  sorry for the noise.

PR #875, opened 2016-02-20
Handling encodings correctly

yan12125:
Fixes #457 

This depends on the latest fix in pypa/distlib. Here are my steps:
```Sh
$ wget https://pypi.python.org/packages/source/p/pip/pip-8.0.2.tar.gz
$ tar xf pip-8.0.2.tar.gz
$ cd pip-8.0.2
$ pushd pip/_vendor/distlib
$ patch -i distlib-encoding.patch -p2
$ popd
$ python setup.py bdist_wheel
```
Where distlib-encoding.patch is from https://bitbucket.org/pypa/distlib/commits/1b082cb:
```Diff
diff --git a/distlib/util.py b/distlib/util.py
--- a/distlib/util.py
+++ b/distlib/util.py
@@ -33,7 +33,8 @@
                      cache_from_source, urlopen, urljoin, httplib, xmlrpclib,
                      splittype, HTTPHandler, HTTPSHandler as BaseHTTPSHandler,
                      BaseConfigurator, valid_ident, Container, configparser,
-                     URLError, match_hostname, CertificateError, ZipFile)
+                     URLError, match_hostname, CertificateError, ZipFile,
+                     fsdecode)
 
 logger = logging.getLogger(__name__)
 
@@ -165,7 +166,10 @@
 #    else:
 #        result = sys.executable
 #    return result
-    return os.path.normcase(sys.executable)
+    result = os.path.normcase(sys.executable)
+    if not isinstance(result, text_type):
+        result = fsdecode(result)
+    return result
 
 
 def proceed(prompt, allowed_chars, error_prompt=None, default=None):
```
When things are done, replace ```virtualenv_support/pip-8.0.2-py2.py3-none-any.whl``` with ```../pip-8.0.2/dist/pip-8.0.2-py2.py3-none-any.whl```

PR #900, opened 2016-05-19
Handling encodings correctly

### LINKED ISSUE #457: Problem with non ASCII car in path
I don't know if this is related to #186 but with my fresh install of python on my Windows 8 I've the following error when using virtualenv in the path C:\Users\oncleben\Dropbox\Developement\fraisDeSanté

```
C:\Users\oncleben\Dropbox\Developement\fraisDeSanté> virtualenv ENV
New python executable in ENV\Scripts\python.exe
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 50: ordinal not in range(128)
ERROR: The executable ENV\Scripts\python.exe is not functioning
ERROR: It thinks sys.prefix is u'c:\\users\\oncleben\\dropbox\\developement\\fraisdesant\xe9' (should be u'c:\\users\\on
cleben\\dropbox\\developement\\fraisdesant\xe9\\env')
ERROR: virtualenv is not compatible with this system or executable
Note: some Windows users have reported this error when they installed Python for "Only this user" or have multiple versi
ons of Python installed. Copying the appropriate PythonXX.dll to the virtualenv Scripts/ directory may fix this problem.
```

---

python v2.7.5
pip list
nose (1.3.0)
pip (1.4)
Scaffold (0.1.3)
setuptools (0.9.7)
virtualenv (1.10)

Edit:
- emoji - see https://github.com/pypa/virtualenv/issues/1014 - 🚒
- è - see https://github.com/pypa/virtualenv/issues/1392
- Cyrillic - see https://github.com/pypa/virtualenv/issues/1375
- $ character

yan12125:
Fixes #457

On Travis CI pypy 2.5.0 and pypy3 2.4.0 failed. On my machine pypy 5.1 works fine:

```
$ pypy virtualenv.py 中文
New pypy executable in /home/yen/tmp/virtualenv/中文/bin/pypy
```

And pypy3 gives a strange error:

```
$ pypy3 virtualenv.py 中文
New pypy executable in /home/yen/tmp/virtualenv/中文/bin/pypy3
Also creating executable in /home/yen/tmp/virtualenv/中文/bin/pypy
debug: OperationError:
debug:  operror-type: LookupError
debug:  operror-value: no codec search functions registered: can't find encoding
debug: OperationError:
debug:  operror-type: AttributeError
debug:  operror-value: stdout
ERROR: The executable /home/yen/tmp/virtualenv/中文/bin/pypy3 is not functioning
ERROR: It thinks sys.prefix is '/home/yen/tmp/virtualenv' (should be '/home/yen/tmp/virtualenv/中文')
ERROR: virtualenv is not compatible with this system or executable
```

I guess it's related to a PyPy bug. Reported to https://bitbucket.org/pypy/pypy/issues/2300.

---

_This was migrated from pypa/virtualenv#875 to reparent it to the `master` branch. Please see original pull request for any previous discussion._


yan12125:
Now only PyPy 3 fails. Let's wait for responses from PyPy.

An question: is it OK to just skip the non-ASCII test on PyPy3 and list it as a known issue?

yan12125:
Skipped PyPy3. PyPy3 is incomplete and full of hacky codes. I don't think virtualenv should completely support it.

Ivoz:
I'm worried this will only happen to work on linuxs with the default locale set as UTF-8. E.g, not on Windows.

yan12125:
You're right. [Appveyor tests failed](https://ci.appveyor.com/project/yan12125/virtualenv). I'll try to install Python on Windows and have a look.

yan12125:
Windows support fixed.

Ivoz:
@yan12125 could you split out the part  of this dealing specifically with pypy into a dependant / independent (whichever suits) PR?

yan12125:
Did you mean changes to ```.travis.yml```?

Ivoz:
and `external/pypy-subprocess.patch` and `tests/test_cmdline.py`

yan12125:
Done at #927

Ivoz:
So I'd suggest you'd want to cherry-pick 27c80568046c5c21aa5cd99a2bc55b0f12d748e5 and 23f7268f9dd392627df0c9b2eed9e75b11cef202 onto a new branch and force push it onto `encoding-fix` to clean up this PR

yan12125:
Done :)

Ivoz:
Uhh, maybe I wasn't clear enough. I mean remove all changes that are effectively in #927 from this PR / branch.

external/pypy-subprocess.patch
```diff
@@ -0,0 +1,14 @@
+--- lib-python/2.7/subprocess.py.orig	2016-05-20 02:12:48.780386785 +0800
++++ lib-python/2.7/subprocess.py	2016-05-20 02:18:24.743731346 +0800
+@@ -667,7 +667,10 @@
+              "elif hasattr(sys.stdout, 'buffer'):\n"
+              "    sys.stdout = sys.stdout.buffer\nsys.stdout.write(prefix)\n"],
+             ['-c', 'import sys;out=sys.stdout;getattr(out, "buffer"'
+-             ', out).write(sys.prefix.encode("utf-8"))']):  # 1.7.2 a9454bce
++             ', out).write(sys.prefix.encode("utf-8"))'],   # 1.7.2 a9454bce
++            ['-c', 'import sys;out=sys.stdout;prefix=sys.prefix;'
++             'prefix=prefix.encode("utf-8") if sys.version_info[0]==3 else prefix;'
```
dholth: Consider switching this to detect python 2 else do the python 3 behavior.

yan12125:
@Ivoz @dholth Done. Sorry for the late response.

yan12125:
Is there anything left that should be fixed?

virtualenv.py
```diff
@@ -1472,7 +1478,7 @@ def install_files(home_dir, bin_dir, prompt, files):
         content = content.replace('__VIRTUAL_ENV__', home_dir)
         content = content.replace('__VIRTUAL_NAME__', vname)
         content = content.replace('__BIN_NAME__', os.path.basename(bin_dir))
-        writefile(os.path.join(bin_dir, name), content)
+        writefile(os.path.join(bin_dir, name), content, use_fsenc=(name == 'activate.bat'))
```
pfmoore: Should also use fsenc for deactivate.bat surely?

virtualenv.py
```diff
@@ -354,11 +356,11 @@ def copyfile(src, dest, symlink=True):
         logger.info('Copying to %s', dest)
         copyfileordir(src, dest, symlink)
 
-def writefile(dest, content, overwrite=True):
+def writefile(dest, content, overwrite=True, use_fsenc=False):
     if not os.path.exists(dest):
         logger.info('Writing %s', dest)
         with open(dest, 'wb') as f:
-            f.write(content.encode('utf-8'))
+            f.write(content.encode(sys.getfilesystemencoding() if use_fsenc else 'utf-8'))
```
pfmoore: ```sys.filesystemencoding()``` is not the correct encoding for bat files - you need to use sys.stdout.encoding. See the following (hopefully the Unicode characters will display correctly...):

```
C:\Work\Scratch>py                                                                                            
Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:18:55) [MSC v.1900 64 bit (AMD64)] on win32                
>>> import sys                                                                                                
>>> sys.getfilesystemencoding()                                                                               
'mbcs'                                                                                                        
>>> sys.stdout.encoding                                                                                       
'cp437'                                                                                                       
>>> with open('x.bat', 'wb') as f:                                                                            
...   f.write("echo \N{POUND SIGN}\n".encode(sys.getfilesystemencoding()))                                    
...                                                                                                           
7                                                                                                             
>>>                                                                                                           
                                                                                                              
C:\Work\Scratch>x.bat                                                                                         
                                                                                                              
C:\Work\Scratch>echo ú                                                                                        
ú
```

virtualenv.py
```diff
@@ -227,7 +227,9 @@ def log(self, level, msg, *args, **kw):
                     else:
                         rendered = msg
                     rendered = ' '*self.indent + rendered
-                if hasattr(consumer, 'write'):
+                if hasattr(consumer, 'buffer'):
+                    consumer.buffer.write(rendered.encode('utf-8')+b'\n')
+                elif hasattr(consumer, 'write'):
```
pfmoore: This seems wrong - why do you assume the consumer is expecting UTF-8? Why is this needed at all?

virtualenv.py
```diff
@@ -729,6 +731,7 @@ def call_subprocess(cmd, show_stdout=True,
                 part = part.decode(sys.getfilesystemencoding())
         cmd_parts.append(part)
     cmd_desc = ' '.join(cmd_parts)
+    cmd_desc = cmd_desc.encode('utf-8')
```
pfmoore: Why UTF-8?

virtualenv.py
```diff
@@ -1376,8 +1379,9 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
         # the value:
         py_executable = '"%s"' % py_executable
     # NOTE: keep this check as one line, cmd.exe doesn't cope with line breaks
-    cmd = [py_executable, '-c', 'import sys;out=sys.stdout;'
-        'getattr(out, "buffer", out).write(sys.prefix.encode("utf-8"))']
+    cmd = [py_executable, '-c', 'import sys;out=sys.stdout;prefix=sys.prefix;'
+        'prefix=prefix if sys.version_info[0]==2 else prefix.encode(sys.getfilesystemencoding());'
+        'getattr(out, "buffer", out).write(prefix)']
```
pfmoore: Surely the point here is to explicitly use UTF-8, because it's guaranteed to encode all Unicode characters correctly, and it's a known encoding so we can send it from the subprocess and receive it in the main process without issue?

What would happen with your code if ```sys.prefix``` contained characters that weren't encodable in ```sys.getfilesystemencoding()```?

Also, why are you converting the protocol to use "string if Python 2 otherwise bytes" rather than a clean always-bytes protocol?

virtualenv.py
```diff
@@ -1391,7 +1395,7 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
         else:
             raise e
 
-    proc_stdout = proc_stdout.strip().decode("utf-8")
+    proc_stdout = proc_stdout.strip().decode(sys.getfilesystemencoding())
```
pfmoore: See above. Using UTF-8 here is completely OK, as the subprocess is guaranteed to send UTF-8 (before your change).

pfmoore:
Tests are failing on pypy. This should be fixed.

Also, I've added a number of comments. The most significant issue is that you're using the wrong encoding for bat files on Windows, but there also seem to be some unneeded changes which concern me.

yan12125:
Fixes for PyPy is in #927. Originally I put all changes in this PR, and split out supporting changes to #927, as requested by @Ivoz in https://github.com/pypa/virtualenv/pull/900#issuecomment-222367004

pfmoore:
I'm confused. Current master tests work on pypy, so maybe #927 is no longer needed? Maybe if you rebase this PR on master pypy will work? Regardless, though, the pypy tests need to be working on this PR before it can be merged.

virtualenv.py
yan12125: > why are you converting the protocol to use "string if Python 2 otherwise bytes" rather than a clean always-bytes protocol?

Actually my codes implement an "always-bytes protocol". On Python 2, ```sys.prefix``` is a bytes-like object, so I print it as-is.

> What would happen with your code if sys.prefix contained characters that weren't encodable in sys.getfilesystemencoding()?

Maybe sys.stdout.encoding is a better choice here.

yan12125:
The problem for PyPy is tricky: PyPy uses [a hack](https://bitbucket.org/pypy/pypy/src/4afa4a9d8caffdd737e4b8f24c157a939c9127d2/lib-python/2.7/subprocess.py?at=default&fileviewer=file-view-default#subprocess.py-658) that relies on virtualenv internals. If I change the statement in https://github.com/pypa/virtualenv/pull/900#discussion-diff-72204625L1379, PyPy should be patched, or libpypy-c.so can't be found.

virtualenv.py
yan12125: Maybe decoding ```sys.prefix``` on Python 2 to UTF-8 is a better choice.

virtualenv.py
pfmoore: > Actually my codes implement an "always-bytes protocol". On Python 2, sys.prefix is a bytes-like object, so I print it as-is.

But strings in Python 2 use ```sys.getdefaltencoding()``` as the encoding, not ```sys.getfilesystemencoding()```, so what you're writing is going to have a different encoding depending on the Python version (unless the default and filesystem encodings are the same, which isn't always the case - and on Windows there's also the OEM encoding to make things even more messy).

A "strings encoded in UTF-8" protocol is far more robust. If you've actually found a case where that protocol doesn't work, you don't make it clear what that case is, so I don't see the value in this change.

At the absolute minimum, this code needs a comment explaining what is wrong with the original approach (of passing string values between subprocess and main program using a known encoding) and how the modified version solves that issue. Ideally, there should be a test, but the virtualenv test suite doesn't really go down to this level :-(

Also, you're making the assumption that ```sys.getfilesystemencoding()``` is the same in both the subprocess Python and the current Python. While that's not likely to be violated, I don't see any benefit in taking that risk, when a specific UTF-8 encoding is just as good for converting arbitrary string data to a binary stream, and involves no such risk.

pfmoore:
So start with master, which passes all the tests on PyPy. Then make your changes, and (unless your code causes a problem for PyPy) all the tests should still pass. If your code breaks on PyPy, the fixes should be in this PR, not in a separate one.

The "fixes" you seem to be making to PyPy appear to be a set of hacks to PyPy to make our tests work. That's not what we should be doing, rather we should be ensuring our tests continue to work with the versions of PyPy that people are using, so that virtualenv works for them.

If you're arguing that virtualenv should drop support for PyPy, then that's a much bigger proposal, and you'd have to come up with a pretty strong argument for doing so.

virtualenv.py
yan12125: Thanks for that. Changed.

virtualenv.py
yan12125: Here's a problem - ```sys.stdout.encoding``` is ```None``` if the terminal is not interactive (python invoked via ```subprocess```, etc.). From testing results ```sys.getfilesystemencoding()``` is OK so I keep them.

virtualenv.py
yan12125: This patch is added at the early stage of this PR - now it's not necessary and I have removed it.

virtualenv.py
yan12125: Also, unnecessary patch anymore.

virtualenv.py
yan12125: Changed back to "always-UTF8" protocol. On Python 2 strings are just bytes - there's no "encoding". On Python 3 ```sys.prefix``` is decoded with ```Py_DecodeLocale``` during initialization. On Windows this function eventually calls ```mbstowcs```. On the other hand, ```sys.getfilesystemencoding()``` is hard-coded to "mbcs", and ```decode('mbcs')``` eventually calls ```MultiByteToWideChar```. If the locale is not changed on the fly, both functions return the same result. As a result, I use ```decode(sys.getfilesystemencoding())``` to get the same result for both Python 2 and Python 3.

virtualenv.py
yan12125: This patch is no longer necessary after switching back to "always-utf8" protocol.

.travis.yml
```diff
@@ -23,7 +23,30 @@ matrix:
...
+            pip install --upgrade --ignore-installed --user virtualenv
+            mkdir -p ~/pypy && pushd ~/pypy
+            wget "https://bitbucket.org/pypy/pypy/downloads/$PYPY_VERSION.tar.bz2"
+            tar jxf "$PYPY_VERSION.tar.bz2"
+            pushd "$PYPY_VERSION" && patch -i "$TRAVIS_BUILD_DIR/external/$TOXENV-subprocess.patch" -N -p0 -F0 && popd
+            virtualenv --python="$HOME/pypy/$PYPY_VERSION/bin/$PYPY_BIN" "$HOME/virtualenvs/$PYPY_VERSION"
+            source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+            popd
+        fi
+    - pip install tox
```
pfmoore: No, this is not acceptable. We need to run our tests with an *unpatched* version of PyPy. If the PR can't be made to work on PyPy, then the PR isn't acceptable, as we currently support PyPy.

appveyor.yml
```diff
@@ -5,15 +5,53 @@ environment:
...
+    - PYTHON: "C:\\Python35-x64"
+      LOCALE: zh-TW
+
+init:
+    - ps: >-
+        If ($env:LOCALE.compareTo("en-US") -ne 0) {
+            Set-WinSystemLocale $env:LOCALE
+            Start-Sleep -s 5
+            Restart-Computer
+        }
```
pfmoore: This adds a huge overhead to the appveyor tests, and doesn't seem reasonable. If you want to run tests that require a particular locale, you should probably mock out the relevant system calls, or something similar.

tests/test_cmdline.py
```diff
@@ -22,6 +30,19 @@ def test_commandline_explicit_interp(tmpdir):
         str(tmpdir.join('venv'))
     ])
 
+# On Windows, path with Chinese characters are not working without a Chinese code page
+@pytest.mark.skipif("get_code_page() not in (None, 936, 950)")
+def test_commandline_non_ascii_path(tmpdir):
+    target_path = str(tmpdir.join('venv中文'))
+    if sys.version_info[0] == 2:
```
pfmoore: Why not use a non-ASCII character that's present in the default code page on Appveyor, rather than messing about changing the locale of the system? Something like an accented character (é) should be available in the default codepage.

virtualenv.py
```diff
@@ -1472,7 +1475,7 @@ def install_files(home_dir, bin_dir, prompt, files):
         content = content.replace('__VIRTUAL_ENV__', home_dir)
         content = content.replace('__VIRTUAL_NAME__', vname)
         content = content.replace('__BIN_NAME__', os.path.basename(bin_dir))
-        writefile(os.path.join(bin_dir, name), content)
+        writefile(os.path.join(bin_dir, name), content, use_system_locale=(name.endswith('.bat')))
```
pfmoore: I'm sorry, but ```getfilesystemencoding``` simply won't work for Windows BAT files. I understand your point that ```sys.stdout.encoding``` may not be available, and I don't actually know a way of getting the "OEM" code page if that's the case. But that's what you need to do - using ```getfilesystemencoding``` is just as broken (on for example a UK English system) as using UTF-8.

tests/test_cmdline.py
yan12125: Some problems exist in multi-byte locales only. See the updated test case in 57a43e823a7e32c7d717f918f12abebf89ffeb2a.

appveyor.yml
yan12125: Mocking system calls sounds good - I'll have to find an alternative to ```sys.getfilesystemencoding()``` first.

yan12125:
Can't find a way to mock all system calls without making the code base very dirty, so changed to a test case that can be run under CP437. As a result, multi-byte related bugs (```normcase``` and ellipsis insertion) are not exposed by tests.

virtualenv.py
```diff
@@ -354,11 +354,21 @@ def copyfile(src, dest, symlink=True):
         logger.info('Copying to %s', dest)
         copyfileordir(src, dest, symlink)
 
-def writefile(dest, content, overwrite=True):
+def get_system_encoding():
+    if is_win:
+        # .encode('mbcs') is broken on Windows. For example,
+        # u'\u00a3'.encode('mbcs') gives a wrong result in CP437 (en-US), and
```
pfmoore: What do you mean a wrong result? u"\u00a3" (£ sign) is encoded to \x9c in CP437. Is that not what you get? Note that "mbcs" is the ANSI codepage, which is *not* the same code page as GetOEMCP.

virtualenv.py
```diff
@@ -718,15 +728,15 @@ def call_subprocess(cmd, show_stdout=True,
...
-            part = '"%s"' % part.replace('"', '\\"')
         if hasattr(part, 'decode'):
             try:
                 part = part.decode(sys.getdefaultencoding())
             except UnicodeDecodeError:
                 part = part.decode(sys.getfilesystemencoding())
+        if len(part) > 45:
+            part = part[:20]+"..."+part[-20:]
+        if ' ' in part or '\n' in part or '"' in part or "'" in part:
+            part = '"%s"' % part.replace('"', '\\"')
```
pfmoore: Why did you move this to after the decode step?

virtualenv.py
```diff
@@ -1284,6 +1294,15 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
...
                 copyfile(d, os.path.join(home_dir, 'lib_pypy'), symlink)
+            else:
+                # On Unix-like systems, if PyPy is built with --shared, it
+                # requires libpypy-c.(so|dylib) to run. See
+                # _pypy_install_libs_after_virtualenv() in
+                # lib-python/(2.7|3)/subprocess.py of PyPy
+                for name in ['libpypy-c.so', 'libpypy-c.dylib']:
+                    src = join(prefix, 'bin', name)
+                    if os.path.exists(src):
+                        copyfile(src, join(bin_dir, name), symlink)
```
pfmoore: This seems unrelated to encodings...

virtualenv.py
yan12125: On other systems ```sys.getfilesystemencoding()``` works fine, while on Windows ```sys.getfilesystemencoding()``` returns ```'mbcs'```, which is not what virtualenv needs. If this comment is confusing I'll change or remove it.

Ivoz:
@yan12125 If there are "broken" stuff in master which currently "happens" to work, or stuff that pypy itself has hacked around specifically, that *otherwise* correct changes in this PR would then break, then what you want to do is make a PR dealing with/fixing specifically those things, and A) ask us to merge that first, B) remake this PR, basing its branch it off the above created PR/branch, and stating so that it is designed to be merged after the pypy-hack-fixing PR.  
Also exactly the same deal with "windows hacks".

If that's not the situation, please disregard the above advice.

The reason to do all this (if that seems like a lot of effort) is so it's easy for us as maintainers to understand your logic as you changed the code and be confident after reading in our own comprehension that the changes are reasonable and correct. I have merged one or two things already in this project (and many in other projects) where I have simply relied on the PR author's earnest statement that their changes are correct and fixed things, and many timed have regretted doing so.

Unfortunately these topics (fixing pypy and windows) also suffer from being typically messy and ugly problems in the first place, naturally making them more effort to fix correctly.

virtualenv.py
yan12125: As mentioned at https://github.com/pypa/virtualenv/pull/900#issuecomment-235240280:

> If your code breaks on PyPy, the fixes should be in this PR, not in a separate one.

This is the (hopefully) correct fix that supersedes my previous dirty PyPy patches.

virtualenv.py
yan12125: For example, before my changes, if part is: ```b'/some/very/long/path/to/break/existing/codes/\xe5\x83\x85\xe6\x98\xaf\xe4\xb8\xad\xe6\x96\x87/bin/python'```, ```part[:20]+"..."+part[-20:]``` gives invalid UTF-8 sequences. As a result ellipsis insertion should occur after decoding.

pfmoore:
Agreed. There seem to be 3 things going on in this PR:

1. Somehow hacking PyPy to stop it somehow behaving wrongly in a way that makes the encoding changes you want to make "not work". That seems more likely to be something that should be submitted upstream as a PR to PyPy, but maybe there's a need for a temporary patch to our tests to deal with the issue until PyPy is fixed. Personally, I'm not completely sure that we should be running our tests with a patched version of PyPy - that's a sure way to have virtualenv fail when our users use it with an unpatched PyPy - but that can be debated in a separate PR.
2. Changes to make virtualenv more cleanly handle encodings in general (i.e., on all platforms). I don't know which parts of this PR those are, but again they should be separated. I may have comments on this part.
3. Windows-specific patches to handle the fact that Windows specific files (bat and ps1) need specific encodings which are *not* the default encoding used by Python. These are the changes I'm most interested in commenting on (and the ones I'd be willing to merge once I'm sure they are implemented correctly).

As @Ivoz suggests, we need multiple PRs here. Otherwise, this PR will simply sit here and never get applied, as no-one is in a position to understand and approve the whole thing. I realise this is more work for you - hopefully you can see why it would be helpful.

virtualenv.py
pfmoore: Ah, OK. Good catch. Thanks for the explanation.

yan12125:
Agreed with you two - this PR is always messy as there are too many bugs in ```virtualenv.py```. Seems PyPy is the easiet part and I'll submit it as an individual one.

Ivoz:
Also note that I am fine, currently, with anything that would leave the code otherwise "correct", but currently fails to work with pypy3 (using 3.2).

A) Not even pip supports 3.2 itself, and B) pypy3, I would pretty much consider beta software until it manages to get a stable release as 3.3 compatible. And otherwise I shudder to think of anyone brave enough to use pypy3 in production somewhere.

So leaving pypy3 able to fail, and fixing it later, while correctly solving issues for pypy (2), I would consider acceptable.

yan12125:
So I can just drop the part that downloads PyPy 3.3 alpha from ```.travis.yml``` and skip my new test? PyPy 3.3 alpha seems the only version that handles Chinese characters correctly. I know it's crazy to use a dependency of alpha version in unit/integraion tests...

virtualenv.py
yan12125: Submitted as a dependency PR at #946

pfmoore:
@Ivoz if we're not going to require tests to succeed on PyPy3, should we not remove it from ```.travis.yml```? Or at least find a way to tell travis that if the PyPy3 tests fail, don't mark the build as failed?

I don't really feel comfortable with having to start merging PRs where our required CI checks are failing.

Note that I'm perfectly OK with us not supporting PyPy3.

yan12125:
As a side note: In the last few months, most new PyPy commits are on the default branch (2.7 compatible) and py3.5/py3.5-async branch, and the progress on py3.5/py3.5-async branches is parsing. Though there's an alpha version, I don't expect a Python 3.3+ compatible PyPy3 release in near future.

Ivoz:
@pfmoore https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

techtonik:
OMG. That's a huge thread. I don't know when I can realistically review this. It would be nice to see some separate repository with explanation of the problem and necessary code to validate concepts on my system.

yan12125:
> necessary code to validate concepts on my system.

Did you mean something like this?
```
git clone -b encoding-fix https://github.com/yan12125/virtualenv
```

yan12125:
A minor update after 49721c1. Can anyone have a look into #946, the first step to the whole solution?

yan12125:
@pfmoore in #1014 you said:

> ... which as noted in that PR, is stalled on being refactored out into separate PRs.

Is there something else I have to do other than splitting out pypy codes into #946?

pfmoore:
Sorry, I can't see how to create a lint to an individual comment, but up above I said:

> Agreed. There seem to be 3 things going on in this PR:
> ...
> As @Ivoz suggests, we need multiple PRs here. Otherwise, this PR will simply sit here and never get applied, as no-one is in a position to understand and approve the whole thing. I realise this is more work for you - hopefully you can see why it would be helpful.

You split out the PyPy changes (which is waiting for someone to review and comment on, but I can't help there as I know nothing about PyPy). You still need (IMO) to split out the general encoding stuff and the Windows script file bits.

I'd happily review the Windows script file bits, I don't have an issue with that. For encodings, it's going to be more complex, as we lack a good test suite, but I'm willing to do what I can (with the proviso that I'd like to see tests demonstrating problem cases, not just reports that can't be reproduced elsewhere...)

gaborbernat:
@yan12125 can you rebase this and fix the CI? thanks!

yan12125:
It's my honor that someone is still looking at my PR. However, as this PR fixes a Python2-only issue, and Python 2 is going to be dead soon, I won't put more efforts into it. If another developer still needs it, feel free to take my branch.

PR #894, opened 2016-05-04
fix error when invoking virtualenv with multibyte directory on windows

aodag:
- on windows py27, sys.prefix is in mbcs, decode to unicode.
- normalizing encoded home_dir causes wrong translation. (e.g テスト -> テベト)

PR #902, opened 2016-05-21
fix error when invoking virtualenv with multibyte directory on windows 

aodag:
on windows py27, sys.prefix is in mbcs, decode to unicode.
normalizing encoded home_dir causes wrong translation. (e.g テスト -> テベト)

---

*This was migrated from pypa/virtualenv#894 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*


yan12125:
Due to silly hacking in PyPy, changed command line in virtualenv requires changing PyPy. See [my workaround](https://github.com/pypa/virtualenv/pull/900/commits/01866b9fa8e580d35c9c4ebbb2778eda6df0cf8b).

Seems you are also affected by #457? My implementation at #900 is quite similar to yours. Maybe we should have a merge. Could you try my patch at #900? It was implemented on Linux and I have never tested it on Windows.

aodag:
@yan12125 
I'm sure that #900 has resolved this issue.

PR #903, opened 2016-05-21
Detect if symlinks are supported on target file system

techtonik:
This detects symlink suppot on destination filesystem and not only in Python interpreter or OS. If symlinks are not supported, falls back to copy behavior.

This fixes some case reported in #92 and #409, without specifying --always-copy explicitly.

Ideal case would be to invert virtualenv symlinking behavior - symlink only when explicitly requested.

---

*This was migrated from pypa/virtualenv#663 to reparent it to the ``master`` branch. Please see original pull request for any previous discussion.*

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1100,6 +1100,27 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
         ## Maybe it should delete everything with #!/path/to/venv/python in it
         logger.notify('Not deleting %s', bin_dir)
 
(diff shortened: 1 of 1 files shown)


techtonik:
@Ivoz @dstufft ready for review. People are patching utilities like tox with workarounds against this bug, and this PR should make it unnecessary - https://bitbucket.org/hpk42/tox/pull-requests/205/fix-issue-218/diff

techtonik:
Ping?

techtonik:
I could wait for 2 years. Sorry. :laughing:

PR #904, opened 2016-05-22
Always store _OLD_VIRTUAL_PATH in a sh compat format

ViViDboarder:
If in a subshell from within `fish`, the PATH variable will be converted to the proper format for sh, but none of the custom path-like variables will. This causes errors because it sets the PATH to an invalid value when activating.

This patch makes activate.fish always store the old path using `:` deliminators and then converts it back to `fish` format later.
### Issue reproduction

Simplest example:

`PATH` converts correctly

```
fish $ echo $PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $PATH'
  /bin:/usr/local/bin
```

`OLD_PATH` does not

```
fish $ set -x OLD_PATH $PATH
fish $ echo $OLD_PATH
  /bin /usr/local/bin
fish $ bash -c 'echo $OLD_PATH'
  /bin/usr/local/bin
```

There you will see the `OLD_PATH` is not a valid path for sh.

---

_This was migrated from pypa/virtualenv#827 to reparent it to the `master` branch. Please see original pull request for any previous discussion._


ViViDboarder:
@dstufft any chance someone can review this?

ViViDboarder:
@gaborbernat are there existing tests I can add to? I can't find any fish tests here, maybe I'm overlooking something.

ViViDboarder:
Updated and added a test.

gaborbernat:
@ViViDboarder please add a changelog for this, not sure why the CI did not trigger here.

gaborbernat:
https://dev.azure.com/pypa/virtualenv/_build/results?buildId=3174 note you'll need it to pass

gaborbernat:
run tox -e fix_lint to fix the final failure 😁

ViViDboarder:
*phew* Done.

Installed the hooks too so that won't happen again.

gaborbernat:
There's a bug at the moment which causes shell tests not available to be silently skipped in CI. I'll need to fix this tomorrow, but this could bring up some platform incompatibilities, and we probably need to provision fish for the CI. Only after if tests pass we're done. Leave it with me for now, unless you want to add fish provisioning into the CI.

ViViDboarder:
Ah. I thought that might need to be added somewhere.

FWIW, it passes for me locally because I have fish installed. I'm not sure where to add it to your build configuration since I've never worked with Azure. If you let me know where to make the change, I'm happy to do that. Otherwise I can leave it to you.

Thanks for the feedback!

ViViDboarder:
Got it!

gaborbernat:
Hmm, seems you did not mark the PR as edit-able. Will close and reopen from another branch.

gaborbernat:
See https://github.com/pypa/virtualenv/pull/1258.

PR #891, opened 2016-04-28
Resubmit PR: drop $_OLD_VIRTUAL_PATH, use remove_from_path instead

maxiimou:
The PR #722 has never been merged since the tests didn't pass OK. I was really interested in the feature it proposed so I fixed it and I added a test to check it's effectivness.

Tests `tests/test_activate.sh` were never run in the TravisCI and thus some of the tests has to be updated.
All tests were re-written and now uses shunit2 and run during CI.

Please let me know what you think, I'll be glad to explain and modify!

PR #905, opened 2016-05-23
drop $_OLD_VIRTUAL_PATH, use remove_from_path instead

MisterFruits:
The PR #722 has never been merged since the tests didn't pass OK. I was really interested in the feature it proposed so I fixed it and I added a test to check it's effectiveness.

Tests tests/test_activate.sh were never run in the TravisCI and thus some of the tests has to be updated.
All tests were re-written and now uses shunit2 and run during CI.

Please let me know what you think, I'll be glad to explain and modify!

---

_This was migrated from pypa/virtualenv#891 to reparent it to the `master` branch. Please see original pull request for any previous discussion._
_Which was migrated from pypa/virtualenv#722, original feature request_


gaborbernat:
Closing due to inactivity.

PR #453, opened 2013-07-12
Bug fix: error at non ASCII symbols in the path

magres:
I had problems when I created an environment in a non ASCII folder, for example:

c:\Привет>virtualenv myenv

Now it's working.

PR #907, opened 2016-05-26
Bug fix: error at non ASCII symbols in the path

BrownTruck:
I had problems when I created an environment in a non ASCII folder, for example:

c:\Привет>virtualenv myenv

Now it's working.

---

_This was automatically migrated from pypa/virtualenv#453 to reparent it to the `master` branch. Please see original pull request for any previous discussion._

_Original Submitter: @magres_


gaborbernat:
Closing due to inactivity.

PR #910, opened 2016-05-26
Fixed file paths with spaces causing pip to choke on *nix systems

BrownTruck:
Spaces in the file path causes virtualenv pip to choke. I added a bit to re-write the shebang if the system is not windows, and the path contains a space. I basically just modified the "fixup_scripts" function.

Tested locally, seems to work pretty well :)

---

_This was automatically migrated from pypa/virtualenv#611 to reparent it to the `master` branch. Please see original pull request for any previous discussion._

_Original Submitter: @blakfeld_


gaborbernat:
Closing due to inactivity.

PR #912, opened 2016-05-26
support install into non-ASCII directories

BrownTruck:
On Python 2 sys.prefix is already not unicode and should not be encoded.

So far only tested on Linux with sys.getfilesystemencoding() == 'utf-8' (certainly it is ascii sometimes, even on the same system, depending on how I've logged in).

---

_This was automatically migrated from pypa/virtualenv#630 to reparent it to the `master` branch. Please see original pull request for any previous discussion._

_Original Submitter: @dholth_


gaborbernat:
Closing due to inactivity.

PR #919, opened 2016-05-26
use relative paths in symlinks

BrownTruck:
Hi, I don't know how much this is related with #473, but this makes all symlinks relative.

I meet this problem because the symlinks break when I want to use printed paths outside the LXC container (much like a chroot if you don't know it) in which I use virtualenv. If all symlinks are relative, I can simply prepend the path to the LXC root to refer to the files.

---

_This was automatically migrated from pypa/virtualenv#758 to reparent it to the `master` branch. Please see original pull request for any previous discussion._

_Original Submitter: @lilydjwg_

### DIFF
--- a/src/virtualenv.py
+++ b/src/virtualenv.py
@@ -401,7 +401,9 @@ def copyfile(src, dest, symlink=True):
     if symlink and hasattr(os, "symlink") and not is_win:
         logger.info("Symlinking %s", dest)
         try:
-            os.symlink(os.path.realpath(src), dest)
+            srcpath = os.path.realpath(src)
+            srcpath = os.path.relpath(srcpath, os.path.dirname(os.path.abspath(dest)))
+            os.symlink(srcpath, dest)
         except (OSError, NotImplementedError):
             logger.info("Symlinking failed, copying to %s", dest)
             copyfileordir(src, dest, symlink)


lilydjwg:
Too bad that all the two failed tests run on MacOS :-(.

gaborbernat:
The biggest challenge of this project is to come up with solutions that are truly cross-platform.

gaborbernat:
In this case, I believe you're conflicting with the changes of https://github.com/pypa/virtualenv/pull/956. Maybe we can't satisfy both requirements at the same time. 🤔

src/virtualenv.py
```diff
@@ -1221,7 +1221,7 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
                 continue
             for fn in os.listdir(stdlib_dir):
                 bn = os.path.splitext(fn)[0]
-                if fn != "site-packages" and bn in REQUIRED_FILES:
+                if fn not in ("site-packages", "lib-dynload") and bn in REQUIRED_FILES:
```
gaborbernat: we should have some explanation attached here as a comment as why not link these 👍

lilydjwg: Yes, I'm still trying to figure out why the tests fail. I'll add a comment when I'm sure this is the solution.

lilydjwg: And no, that's not the right way to fix. I think I need to read the code thoroughly....

gaborbernat: I've merged in realpath, so try based on top of that?

gaborbernat:
@lilydjwg do you still plan to solve this?

lilydjwg:
No, I give up.

PR #784, opened 2015-08-17
Executing activate_this.py twice corrupts sys.prefix

werat:
After executing activate_this.py, sys.prefix is modified and no longer is the "real" prefix. If activate_this.py  is executed second time (e.g. run some script from another virtualenv), sys.real_prefix becomes sys.prefix and "real" prefix is lost.

PR #920, opened 2016-05-26
Executing activate_this.py twice corrupts sys.prefix

BrownTruck:
After executing activate_this.py, sys.prefix is modified and no longer is the "real" prefix. If activate_this.py  is executed second time (e.g. run some script from another virtualenv), sys.real_prefix becomes sys.prefix and "real" prefix is lost.

---

_This was automatically migrated from pypa/virtualenv#784 to reparent it to the `master` branch. Please see original pull request for any previous discussion._

_Original Submitter: @werat_


gaborbernat:
Closing due to inactivity.

PR #923, opened 2016-05-27
Add support for setting LD_LIBRARY_PATH in bash

tomspur:
When activating the virtualenv, prepend $VIRTUAL_ENV/lib to LD_LIBRARY_PATH.

This is an continuation of #326. Further issues/todo items that showed up in that issue, that require more discussion from virtualenv core developers:

Add support for
- [ ] CPATH
- [ ] PYTHONPATH
- [ ] implementation into other shells than bash


stuaxo:
This helps with aubio + probably various other libraries with native components.

gaborbernat:
Closing due to inactivity.

PR #927, opened 2016-05-29
Encoding fix dependency

yan12125:
Dependencies for PR #900

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,29 @@ matrix:
     - python: 2.7
       env: TOXENV=docs
 
-install: pip install tox
+install:
+    - |
+        case "$TOXENV" in
+        "pypy")
+            PYPY_VERSION="pypy-2.5.0-linux64"
+            PYPY_BIN="pypy"
+            ;;
+        "pypy3")
+            PYPY_VERSION="pypy3.3-v5.2.0-alpha1-linux64"
+            PYPY_BIN="pypy3"
+            ;;
+        esac
+        if [ -n "$PYPY_VERSION" ] ; then
+            deactivate
+            pip install --upgrade --ignore-installed --user virtualenv
+            mkdir -p ~/pypy && pushd ~/pypy
+            wget "https://bitbucket.org/pypy/pypy/downloads/$PYPY_VERSION.tar.bz2"
+            tar jxf "$PYPY_VERSION.tar.bz2"
+            virtualenv --python="$HOME/pypy/$PYPY_VERSION/bin/$PYPY_BIN" "$HOME/virtualenvs/$PYPY_VERSION"
+            source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+            popd
+        fi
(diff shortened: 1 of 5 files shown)


yan12125:
```appveyor.yml``` is updated and now ```test_commandline_non_ascii_path``` is also run on Appveyor. See https://ci.appveyor.com/project/yan12125/virtualenv/build/1.0.25 (3.3 test failed due to #888)

File: appveyor.yml [hunk 1]
```diff
@@ -15,6 +15,11 @@ environment:
     - PYTHON: "C:\\Python34-x64"
     - PYTHON: "C:\\Python35-x64"
 
+init:
+    - ps: Set-WinSystemLocale zh-TW
+    - ps: Start-Sleep -s 5
+    - ps: Restart-Computer
+
```
pfmoore: What is going on here? Why would we run our test suite with a specific locale? If we were running the test suite with both English and Chinese locales, that would be one thing, but this change means that we never run the tests in an English locale.

.travis.yml
```diff
@@ -23,7 +23,30 @@ matrix:
...
+            ;;
+        esac
+        if [ -n "$PYPY_VERSION" ] ; then
+            deactivate
+            pip install --upgrade --ignore-installed --user virtualenv
+            mkdir -p ~/pypy && pushd ~/pypy
+            wget "https://bitbucket.org/pypy/pypy/downloads/$PYPY_VERSION.tar.bz2"
+            tar jxf "$PYPY_VERSION.tar.bz2"
+            pushd "$PYPY_VERSION" && patch -i "$TRAVIS_BUILD_DIR/external/$TOXENV-subprocess.patch" -N -p0 -F0 && popd
+            virtualenv --python="$HOME/pypy/$PYPY_VERSION/bin/$PYPY_BIN" "$HOME/virtualenvs/$PYPY_VERSION"
```
pfmoore: We shouldn't be patching PyPy to make our tests pass - we should be fixing our tests to make them pass with PyPy!

yan12125: PyPy has patched ```subprocess.py```, which copies ```libpypy-c.so``` if pypy thinks the caller is virtualenv. To make the decision, pypy [checks ```sys.argv```](https://bitbucket.org/pypy/pypy/src/4afa4a9d8caffdd737e4b8f24c157a939c9127d2/lib-python/2.7/subprocess.py?at=default&fileviewer=file-view-default#subprocess.py-660). You can see virtualenv commit IDs (a9454bce, 0e9342ce, 10ba3f3c) in PyPy's source codes mentioned above. That's the truth - in some cases PyPy should be updated if virtualenv is changed. PyPy's architecture is silly, but I don't have many choices other than patching PyPy.

appveyor.yml
yan12125: This PR is merged into #900 and I've changed ```appveyor.yml``` there.

yan12125:
Merged into #900.

PR #926, opened 2016-05-29
Fix Tcl for Python 3.3-

techtonik:
https://github.com/pypa/virtualenv/pull/888#issuecomment-222345374

PR #931, opened 2016-06-08
Check if tcl/tk files/dirs exist before copying

nodirt:
Copying fails if tcl/tk files are not there. This happens in our setup,
where Python 2.7.6 is used.
Check if they exist before copying.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1074,7 +1074,8 @@ def copy_tcltk(src, dest, symlink):
     for name in ['tcl', 'tk']:
         srcdir = src + '/tcl/' + name + libver
         dstdir = dest + '/tcl/' + name + libver
-        copyfileordir(srcdir, dstdir, symlink)
+        if os.path.exists(srcdir):
+            copyfileordir(srcdir, dstdir, symlink)
 
 def subst_path(prefix_path, prefix, home_dir):
     prefix_path = os.path.normpath(prefix_path)


nodirt:
cc @techtonik

techtonik:
How come that they are not there? I'd be against workarounds for hacked installations unless there is a really good reason.

Also, https://www.cvedetails.com/vulnerability-list/vendor_id-10210/product_id-18230/version_id-160977/Python-Python-2.7.6.html

nodirt:
We don't use tcl/tk in our installation and prefer to deploy it to thousands of our machines

techtonik:
@nodirt I am interested to use the socket vulnerabilities on thousands of your machines. I don't mind against the change, especially if PyPA group gets some funding as a result of incorporating this change (because I don't have write access). Just curious what do you do with Python 2.7.6 there?

rbcarson:
I support this change. It's necessary to check that the tcl/tk directories being copied actually exist, because they're an optional install item. In my case, I haven't installed it on my Python 2.7.11 and 3.5.1, and my Tox stopped working due to the change in #888.

techtonik:
>  It's necessary to check that the tcl/tk directories being copied actually exist, because they're an optional install item

@rbcarson good point.

nodirt:
@techtonik you are right, we should update Python. Turned out it has nothing to do with this PR. Even with the latest version we still prefer not to deploy TCL/TK.

techtonik:
I see that my fix for #926 was not merged, so unless somebody finds a way to sponsor virtualenv development, it may not be fixed any time soon.

nodirt:
@dstufft it's been a week. Please take a look. Alternatively, please take a look at https://github.com/pypa/virtualenv/pull/933
We can't update to the new virtualenv because of this

Ivoz:
This change should be obsoleted by #937, cheers

PR #933, opened 2016-06-10
Generalize Windows Tcl/Tk search; handle case where it's missing.

rbcarson:
This change copies over Tcl/Tk libraries on Windows into the virtualenv, but adds a check so handle the case where the libraries aren't present. Tcl/Tk is an optional component of the installation of Python 2 and 3. This also generalizes the search so either 8.5 or 8.6 are picked up, without needing to be aware of which Python versions ship with which Tcl/Tk version.

The goal here is to combine the solutions of #926 and #931 as a follow-up to #888 and #93.

I looked at the contributing guidelines and see that I should be making pull requests against `develop`, but that branch seems to have disappeared.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1065,16 +1065,18 @@ def copy_required_modules(dst_prefix, symlink):
                 if os.path.exists(pyfile):
                     copyfile(pyfile, dst_filename[:-1], symlink)
 
+
 def copy_tcltk(src, dest, symlink):
     """ copy tcl/tk libraries on Windows (issue #93) """
-    if majver == 2:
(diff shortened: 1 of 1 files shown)


nodirt:
+1
Currently we can't switch to the new version of virtualenv because we don't deploy tcl/tk. This PR will solve our problem.

lichray:
Just encountered this issue.  Terrible.

Ivoz:
I've basically copied the approach in fead5d4542457be0d7493b1301c4f0b1cd4fd3c2 since the merge wasn't going to be clean. Thanks for approach.

PR #936, opened 2016-07-01
Remove invalid quotes and comment

zooba:
Since CMD is being passed to `Popen(..., shell=False)`, the quotes are not necessary and actually _break_ paths with spaces, since the underlying `CreateProcess` call cannot handle quotes around the process path.

Tested with Python 2.6.6, 2.7.12, 3.2.3 and 3.5.2.


zooba:
IIUC, the failed AppVeyor build is fixed by #937. Only Python 3.3 failed, and none of it was due to spaces in the path.

PR #946, opened 2016-08-01
Copy libpypy-c.(so|dylib)

yan12125:
Dependency of #900. To test it, add a blank in [line 1383](https://github.com/yan12125/virtualenv/blob/79c27eff71d11ea8977361ea11084751542284d1/virtualenv.py#L1383) (of course the codes should be equivalent) and see how PyPy tests on Travis CI go red.


yan12125:
Closing for the same reason as https://github.com/pypa/virtualenv/pull/900.

PR #947, opened 2016-08-03
Replace imp.load_module() with importlib.machinery

otarabai:
`imp.load_module()` is [deprecated](https://docs.python.org/3/library/imp.html#imp.load_module). Replacing it with `importlib.machinery.SourceFileLoader(name, path).load_module()`

### DIFF
--- a/virtualenv_embedded/distutils-init.py
+++ b/virtualenv_embedded/distutils-init.py
@@ -1,7 +1,7 @@
 import os
 import sys
 import warnings 
-import imp
+import importlib
 import opcode # opcode is not a virtualenv module, so we can use it to find the stdlib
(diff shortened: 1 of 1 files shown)


pfmoore:
I think we still support Python 2.6, and IIRC, importlib isn't available on 2.6. So unless we want to drop support for 2.6, we can't unconditionally use importlib (and I'm not sure there's much benefit in doing the conditional dance just to avoid a deprecation).

commodo:
also, if doing this change, I think you may need to update this var too
https://github.com/pypa/virtualenv/blob/master/virtualenv.py#L2117

that one contains a gzipped + base64 encoded file that also has a `imp.load_module()` call
that file gets generated when installing a `virtualenv`

PR #948, opened 2016-08-11
Only remove the virtualenv from the PATH

robsonpeixoto:
Today we have to use a lot of tools to develop a system, and because of it we need version managers from python, node, ruby ...
The current activate implementation will keep a old path saved and change using the new one.
But after the deactivate, it will restore the old path and the developer can loose all configuration from other version manager.

With this PR it will only remove the PATH the VIRTUAL_ENV path.

### DIFF
--- a/virtualenv_embedded/activate.fish
+++ b/virtualenv_embedded/activate.fish
@@ -2,10 +2,13 @@
 # Do not run it directly.
 
 function deactivate -d 'Exit virtualenv mode and return to the normal environment.'
-    # reset old environment variables
-    if test -n "$_OLD_VIRTUAL_PATH"
-        set -gx PATH $_OLD_VIRTUAL_PATH
-        set -e _OLD_VIRTUAL_PATH
+    # remove virtualenv from path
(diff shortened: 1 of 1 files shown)


techtonik:
I guess the decision was that `activate/deactivate` should be atomic. For example, if tests that are run in `virtualenv` modify PATH, you can get plenty of pain after you exit.

robsonpeixoto:
Are you caring about the code change the PATH? Is the virtualenv responsibility take care about the user code?

techtonik:
`virtualenv` responsibility is to provide an isolated environment for running Python programs. Why should it care about user code?

robsonpeixoto:
Sorry, I do not understand it: `For example, if tests that are run in virtualenv modify PATH, you can get plenty of pain after you exit.`

But I get your point. Are there a way to change the `activate.fish` template?

techtonik:
The point is that scripts inside `virtualenv` can modify environment variables too, and it is better to clean this before exit. But you may try to get more opinions by asking in mailing list.

> Are there a way to change the activate.fish template?

You've already changed it. Did you forget to run `bin/rebuild-script.py`?

Issue #268, opened 2012-05-11
os.readlink doesn't dereference relative symlinks in mkvirtualenv


thom-nic:
So I can't say how or why this problem hasn't occurred for other people, but here's my situation....

I've got python installed via homebrew on Mac.  This means (among other things) that `/usr/local/bin/python` is symlinked to `../Cellar/python/2.7.3/include/python2.7`.  

Now, when I run `virtualenv test`, I get a bunch of not-helpful information...

```
PYTHONHOME is set.  You *must* activate the virtualenv before using it
New python executable in test/bin/python
Traceback (most recent call last):
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/site.py", line 563, in <module>
        main()
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/site.py", line 545, in main
        known_paths = addusersitepackages(known_paths)
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/site.py", line 278, in addusersitepackages
        user_site = getusersitepackages()
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/site.py", line 253, in getusersitepackages
        user_base = getuserbase() # this will also set USER_BASE
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/site.py", line 243, in getuserbase
        USER_BASE = get_config_var('userbase')
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/sysconfig.py", line 521, in get_config_var
        return get_config_vars().get(name)
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/sysconfig.py", line 420, in get_config_vars
        _init_posix(_CONFIG_VARS)
    File "/usr/local/Cellar/python/2.7.3/lib/python2.7/sysconfig.py", line 299, in _init_posix
        raise IOError(msg)
IOError: invalid Python installation: unable to open /Users/tnichols/.virtualenvs/test/include/python2.7/pyconfig.h (No such file or directory)
ERROR: The executable test/bin/python is not functioning
ERROR: It thinks sys.prefix is u'/Users/tnichols/.virtualenvs' (should be u'/Users/tnichols/.virtualenvs/test')
ERROR: virtualenv is not compatible with this system or executable
```

But believe it or not, the problem is that the copied files are broken symlinks:

```
$ ll ~/.virtualenvs/test/include/python2.7
[...]

PR #956, opened 2016-09-20
Dereference symlinks more thoroughly in copyfile

atsampson:
copyfile was broken for symlinks that pointed to relative paths, and for
symlinks that pointed to further symlinks (which meant that virtualenv
failed to work in some cases when the Python executable it found was
a symlink, e.g. when Python had been installed using stow or homebrew;
issue #268).

Make it follow symlinks until it finds an actual file to copy.

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -337,6 +337,34 @@ def test_relative_symlink(tmpdir):
     assert os.path.exists(lib64)
 
 
+@pytest.mark.skipif(not hasattr(os, "symlink"), reason="requires working symlink implementation")
+def test_copyfile_from_symlink(tmp_path):
+    """Test that copyfile works correctly when the source is a symlink with a
+    relative target, and a symlink to a symlink. (This can occur when creating
+    an environment if Python was installed using stow or homebrew.)"""
+
+    # Set up src/link2 -> ../src/link1 -> file.
+    # We will copy to a different directory, so misinterpreting either symlink
+    # will be detected.
+    src_dir = tmp_path / "src"
+    src_dir.mkdir()
+    with open(str(src_dir / "file"), "w") as f:
+        f.write("contents")
+    os.symlink("file", str(src_dir / "link1"))
+    os.symlink(str(Path("..") / "src" / "link1"), str(src_dir / "link2"))
+
+    # Check that copyfile works on link2.
+    # This may produce a symlink or a regular file depending on the platform --
+    # which doesn't matter as long as it has the right contents.
+    copy_path = tmp_path / "copy"
+    virtualenv.copyfile(str(src_dir / "link2"), str(copy_path))
+    with open(str(copy_path), "r") as f:
+        assert f.read() == "contents"
+
+    shutil.rmtree(str(src_dir))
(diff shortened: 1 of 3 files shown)


atsampson:
I've rebased this on current master, and added a test case where `python` is a relative symlink to a relative symlink to the real Python.

tests/test_virtualenv.py
```diff
@@ -337,6 +338,39 @@ def test_relative_symlink(tmpdir):
...
+    bin_path = os.path.join(tmpdir, "bin")
+    os.mkdir(bin_path)
+    real_python_path = distutils.spawn.find_executable("python")
+    symlink_1_path = os.path.join(bin_path, "monty")
+    os.symlink(os.path.relpath(real_python_path, bin_path), symlink_1_path)
+    symlink_2_path = os.path.join(bin_path, "python")
+    os.symlink("monty", symlink_2_path)
+
+    # Adjust PATH and check we find our new symlink.
+    os.environ["PATH"] = bin_path + ":" + orig_path
```
gaborbernat: let' use pytest ``monkeypatch`` here, modifying a global variable without resetting it is not nice

tests/test_virtualenv.py
```diff
@@ -337,6 +338,39 @@ def test_relative_symlink(tmpdir):
     assert os.path.exists(lib64)
 
 
+@pytest.mark.skipif(not hasattr(os, "symlink"), reason="requires working symlink implementation")
+def test_python_is_symlink(tmpdir):
```
gaborbernat: can you use the new ``tmp_path`` instead? thanks!

tests/test_virtualenv.py
gaborbernat: you'll want to use here the ``os.pathsep``, as yes Windows does has symlink features

tests/test_virtualenv.py
```diff
@@ -337,6 +338,39 @@ def test_relative_symlink(tmpdir):
...
+
+    # Create the venv, and check that python works.
+    ve_path = os.path.join(tmpdir, "venv")
+    os.mkdir(ve_path)
+    virtualenv.create_environment(ve_path)
+    assert os.path.exists(os.path.join(ve_path, "bin", "python"))
+
+    shutil.rmtree(bin_path)
+    shutil.rmtree(ve_path)
+    os.environ["PATH"] = orig_path
```
gaborbernat: prefer monkeypatch, e.g. this will not reset if the test fails

atsampson:
Thanks for the review! I've made it use `tmp_path`, and I've replaced the `PATH` manipulation with a more direct test of `copyfile` (since even with some fiddling, the `PATH` stuff didn't work on all the MacOS and Windows test systems).

atsampson:
Done!

lilydjwg:
Hi there, can someone tell me why `os.path.realpath` isn't used?

gaborbernat:
I'll defer to @atsampson to explain his choices 👍

atsampson:
It hadn't occurred to me to use `realpath` - that's much simpler, thanks! I'll submit another pull request.

PR #959, opened 2016-09-27
Allow sourcing activate.sh in shells "exit on error" mode

### LINKED ISSUE #958: `activate.sh` cannot be sourced in a `setopt -e` script in zsh
if the zsh shell is using `-e`, aka `exit on any error`, then sourcing activate.sh doesn't work **any longer**. 
The problem is that the command `alias pydoc 2>/dev/null >/dev/null && unalias pydoc` fails, and causes the shell to exit (without any error message). Since not having `pydoc` alias when running the script is hardly  an erroneous condition, it shouldn't fail there.

Thus the `alias pydoc 2>/dev/null >/dev/null && unalias pydoc` should be changed to `alias pydoc 2>/dev/null >/dev/null && unalias pydoc || true`

ztane:
The newly added `pydoc` unaliasing command ends with exit code 1 if `pydoc` alias is not set; this in turn means that sourcing `activate.sh` in a shell that had been set to exit on first error, upon sourcing that file, will exit on that line. And thanks to suppressing the output of `pydoc` there is no error message being displayed. Simply adding `|| true` should ensure that the command ends successfully.

Fixes  #958

### DIFF
--- a/virtualenv_embedded/activate.sh
+++ b/virtualenv_embedded/activate.sh
@@ -64,7 +64,7 @@ if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
 fi
 
 # Make sure to unalias pydoc if it's already there
-alias pydoc 2>/dev/null >/dev/null && unalias pydoc
+alias pydoc 2>/dev/null >/dev/null && unalias pydoc || true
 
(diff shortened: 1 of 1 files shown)


gaborbernat:
Thanks!

PR #967, opened 2016-10-20
Enable tab completion on python3.

ar45:
Coppied enablerlcompleter from
https://hg.python.org/cpython/file/2b089035a453/Lib/site.py#l376


davidism:
Flask would like to see this as well. `flask shell` calls `code.interact`. We'd like to be able to call `sys.__interacthook__` to enable completions like the standard shell does, but the `site.py` created by virtualenv doesn't define it. `venv`, however, does define it.

gaborbernat:
@ar45, @davidism thanks for this superseded by https://github.com/pypa/virtualenv/pull/1239 this will be part of ``16.1`` to be released soon

PR #975, opened 2016-11-07
install_wheel.SCRIPT: Project names passed through format() instead o…

AlexeySanko:
…f sys.args

When virtualenv was called from other tool, this tool can use input arguments and these arguments will be passed to SCRIPT.
For example, if I call pip repo test from PyCharm I get next command line: 
['install', '--ignore-installed', '--cert', 'c:\\users\\asanko\\appdata\\local\\temp\\tmp0elwlm', 'C:\\Program Files (x86)\\JetBrains\\PyCharm Community Edition 2016.2.3\\helpers\\pydev\\pydevd.py', '--multiproc', '--qt-support', '--client', '127.0.0.1', '--port', '50821', '--file', 'setuptools']   
Instead of expected:
['install', '--ignore-installed', '--cert', 'c:\\users\\asanko\\appdata\\local\\temp\\tmp0elwlm' 'setuptools']   
We have project names and can pass it directly to script.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ envlist =
 deps =
     mock
     pytest
+    pip
 commands =
     py.test
 
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,6 @@ envlist =
 deps =
     mock
     pytest
-    pip
 commands =
     py.test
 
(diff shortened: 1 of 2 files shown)


AlexeySanko:
Checks didn't pass because [issues-978](https://github.com/pypa/virtualenv/issues/978)

AlexeySanko:
I need to re-format pull request from notmaster branch.

PR #977, opened 2016-11-08
Use lexists when overwriting symlinks

### LINKED ISSUE #976: Overwriting virtualenv with different "-p" argument fails (Fedora 24)
This is an odd problem, I admit.

If an existing virtualenv has broken symlinks, you can't overwrite it with a new virtualenv 

```
$ ls -l ./test/bin/python2.7
lrwxrwxrwx. 1 iwienand iwienand 12 Nov  8 15:41 ./test/bin/python2.7 -> /tmp/nothere
```

If you try to recreate this virtualenv, it fails

```
$ virtualenv ./test/
New python executable in /tmp/test/bin/python2
Not overwriting existing python script /tmp/test/bin/python (you must use /tmp/test/bin/python2)
  File "/usr/bin/virtualenv", line 9, in <module>
    load_entry_point('virtualenv==15.0.3', 'console_scripts', 'virtualenv')()
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 711, in main
    symlink=options.symlink)
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 924, in create_environment
    site_packages=site_packages, clear=clear, symlink=symlink))
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 1369, in install_python
    os.symlink(py_executable_base, full_pth)
OSError: [Errno 17] File exists
````

This is all part of an automated CI image build for openstack, where we were creating the virtualenv with puppet then overwriting it later on in a later part of the build.  Nothing about that is right and I still haven't fully ascertained why it gets broken.  However it does seem we could lexists() here and overwrite the broken symlink; testing shows that does work to fix up the virtualenv

ianw:
Use lexists so that any existing broken symlinks are detected and
removed when setting up the required symlinks.

Fixes #976

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1365,7 +1365,7 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
 
         for pth in required_symlinks:
             full_pth = join(bin_dir, pth)
-            if os.path.exists(full_pth):
+            if os.path.lexists(full_pth):
                 os.unlink(full_pth)
             if symlink:
                 os.symlink(py_executable_base, full_pth)


ianw:
This is not correct; more investigation has revealed it's not quite a broken symlink

PR #992, opened 2016-12-02
Copy libs/pythonXY.lib to support building of C extensions

jcfr:
This issue was discovered while using pytest-virtualenv to create an
isolated environment and test that a source distribution would compile.

References:
 * https://docs.python.org/2/extending/windows.html#using-dlls-in-practice
 * https://docs.python.org/3.6/extending/windows.html#using-dlls-in-practice

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1261,6 +1261,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
             elif os.path.exists(pythondll_d_dest):
                 logger.info('Removed %s as the source does not exist' % pythondll_d_dest)
                 os.unlink(pythondll_d_dest)
+            # we also need to copy the static library to support the case where
+            # the interpreter is statically linked.
+            # considering that on windows, python static libraries are located in
+            # the  "<root>/libs" directory, the following code will look for and
+            # copy "pythonXY.lib".
+            pythonlib_name = "python%s%s.lib" % (
+                sys.version_info[0], sys.version_info[1])
+            pythonlib = os.path.join(os.path.dirname(sys.executable), "libs", pythonlib_name)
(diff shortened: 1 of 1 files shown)


jcfr:
Cc: @thewtex

jcfr:
Cc: @dstufft

jcfr:
Hi @dstufft  , any chance to get this merged :smile:  

Let us know if you need anything else to move this forward.

jcfr:
After further investigation, this is **NOT** required to build python extension. Indeed, on windows, building an extension using the distutils based build system do not attempt to link against python library.

Improving [scikit-build](https://github.com/scikit-build/scikit-build) to better handle the case where the python installation does not provide libraries allow to build extension as expected.

jcfr:
> building an extension using the distutils based build system do not attempt to link against python library.

In fact, the library is implicitly specified in [PC/pyconfig.h](https://github.com/python/cpython/blob/26436b94042ff238bf6e96a9370495586cdc0761/PC/pyconfig.h#L285-L301), copying the library into the virtual env is consistent.

```
[...]

/* For an MSVC DLL, we can nominate the .lib files used by extensions */
#ifdef MS_COREDLL
#       ifndef Py_BUILD_CORE /* not building the core - must be an ext */
#               if defined(_MSC_VER)
                        /* So MSVC users need not specify the .lib file in
                        their Makefile (other compilers are generally
                        taken care of by distutils.) */
#                       if defined(_DEBUG)
#                               pragma comment(lib,"python35_d.lib")
#                       elif defined(Py_LIMITED_API)
#                               pragma comment(lib,"python3.lib")
#                       else
#                               pragma comment(lib,"python35.lib")
#                       endif /* _DEBUG */
#               endif /* _MSC_VER */
#       endif /* Py_BUILD_CORE */
#endif /* MS_COREDLL */

[...]
```

jcfr:
Closing. Superseded by #1023

PR #999, opened 2017-01-08
Handle PS1 prompts starting with newlines

HaraldNordgren:
I use zsh as my primary shell and I have a custom `PS1` starting with newlines:

```
% echo $PS1

${user}${host} %3~
 ${psign} 
```

And this is what a few empty lines on my regular terminal windows looks like:

```
harald@harald-K56CA sandbox/python/virtualenv
 % 

harald@harald-K56CA sandbox/python/virtualenv
 % 

harald@harald-K56CA sandbox/python/virtualenv
 % 
```

Using virtualenv, I get a rather ugly terminal looking like this:

```
(env) 
harald@harald-K56CA sandbox/python/virtualenv
 % 
(env) 
harald@harald-K56CA sandbox/python/virtualenv
 % 
(env) 
harald@harald-K56CA sandbox/python/virtualenv
 % 
```

This is because the virtualenv directory is simply tacked on at the beginning of the prompt.

With this code, I'm attempting to do a bit more clever matching, to move the virtualenv text after the newlines and just before the "real" prompt begins. This makes my prompt look like this instead:

```
(env) harald@harald-K56CA sandbox/python/virtualenv
 % 

(env) harald@harald-K56CA sandbox/python/virtualenv
 % 

(env) harald@harald-K56CA sandbox/python/virtualenv
 % 
```


HaraldNordgren:
I was confused about the `ACTIVATE_SH = convert("""` line. I don't quite understand why not simply put the activate script there instead.

But I attempted to conform the the current model by running my updated scripts into a `deconvert.py` script I put together:

```
#!/usr/bin/env python

import base64
import sys
import textwrap
import zlib


file_name = sys.argv[1]

utf = open(file_name, 'r').read().encode('utf-8')
zlib_compressed = zlib.compress(utf)
b64_encoded = base64.b64encode(zlib_compressed)
deconverted = b64_encoded.decode('ascii')

for line in textwrap.wrap(deconverted, 76):
    print(line)
```

I'd love to hear about the proper way to handle this. My current method has a fairly long debug cycle.

HaraldNordgren:
Adding a test in `tests/test_activate.sh`.

pradyunsg:
Maybe it'll be nicer if the actual substitution is done in the shell, instead of in a Python script
...

HaraldNordgren:
@pradyunsg I considered that, but shouldn't it be perfectly acceptable to use Python when you are in a Python extension? 😀

HaraldNordgren:
@pjenvey @mattip @dstufft @pradyunsg Some attention on this one?

pradyunsg:
1. I don't think it's a good idea to create a Python subprocess for something that can be trivially done in the shell itself.

2. I'm not a committer for this project so, I don't really have the ability to merge these changes.

3. About the lack of a response on this PR -- that's mostly because @pypa is basically a group of volunteers and I guess it's just been the matter of no one having the time to follow up on this. :)

virtualenv_embedded/activate.sh
```diff
@@ -53,12 +53,23 @@ if ! [ -z "${PYTHONHOME+_}" ] ; then
...
++)(.*)'
+
 if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
     _OLD_VIRTUAL_PS1="$PS1"
     if [ "x__VIRTUAL_PROMPT__" != x ] ; then
         PS1="__VIRTUAL_PROMPT__$PS1"
     else
-        PS1="(`basename \"$VIRTUAL_ENV\"`) $PS1"
+        env_basename=`basename "$VIRTUAL_ENV"`
+        set +u
```
HaraldNordgren: Setting this so that `tests/test_activate.sh` won't fail on an unbound variable `ZSH_VERSION` when running on Bash where it is supposed to be unset.

HaraldNordgren:
I think maybe you have a point about moving the logic to the shell, I gave it a try. I don't think it was trivial though. Since I'm using `zsh` and I would want my changes to work in `bash` and `sh` I have to do a bunch of extra logic now.

The nice thing about Python is that it's shell independent.

HaraldNordgren:
I pinged some of the recent committers. Probably should have pinged the maintainers instead.

So: @pfmoore @xavfernandez

Ivoz:
@HaraldNordgren I believe the intention was, if virtualenv's own mangling of PS1 was not to your liking in some way or another, you can set the env var [`VIRTUAL_ENV_DISABLE_PROMPT`](https://virtualenv.pypa.io/en/stable/reference/#envvar-VIRTUAL_ENV_DISABLE_PROMPT) to stop it doing so; and then in your own PS1-setting code, merely make use of the env var `VIRTUAL_ENV` to display it.

If you have, separately, additions that could improve the compatibility of the tests with zsh, then that could be a good separate PR.

HaraldNordgren:
@gaborbernat Test are already there -- could you be specific about what needs extra testing if you want me to add that? 😊

I can put in the work to rebase the change, but only if I have a sense that this will actually get merged. The PR is almost 2 years old, and some of the comments were pretty negative.

gaborbernat:
@HaraldNordgren I do plan to merge it, hence me bugging on this (I've took over maintenance of this a few months back, going through the backlog now). While there are some tests I would prefer to move the shell one into Python code, similar to how we have with powershell/xonosh. Furthermore, I don't there are tests for the use case of having newline in the PS1, so that would be the extra.

gaborbernat:
@HaraldNordgren are you still open to tackle this?

HaraldNordgren:
@gaborbernat Yes, I think so. Possibly within the next few weeks 😊

bskinn:
Along the same lines as the discussion on #1168, what about a possible case where a user wants the env name to be inserted before the leading newlines, at the start of the prompt per normal behavior?

It seems to me this `env-prompt-after-newlines` behavior should have to be specifically switched on, either at env creation time (arg to `virtualenv`) or at activation time, say via an environment variable.

gaborbernat:

Issue #997, opened 2017-01-05
Long env path names cause weird errors after activiting virtual environment


HaraldNordgren:
After creating a virtualenv in the path

`/home/harald/git-repos/sandbox/stackoverflow/41454203-how-do-you-import-the-library-being-tested-when-running-a-nosetest/env`

I could not run the env's pip,

```
nosetest/env$ ./bin/pip
bash: ./bin/pip: /home/harald/git-repos/sandbox/stackoverflow/41454203-how-do-you-import-the-l: bad interpreter: Permission denied
```

and running just pip gives me the system's pip, with all of its packages.

yan12125:
Not virtualenv's fault. Shebangs have length limits, see http://www.in-ulm.de/~mascheck/various/shebang/.

This is basically a duplicate of #994 - virtualenv are broken with fragile kernel shebang parsing.

HaraldNordgren:
Where is the shebang written to file?

And is there maybe some workaround we could employ to make virtual environments more sturdy?

yan12125:
> Where is the shebang written to file?

Do you mean the location of the shebang or the codes that generate a shebang? 1. In the first line of the file ```bin/pip``` 2. In [distlib](https://bitbucket.org/pypa/distlib/src/f702f6afc1d03717fa88a742b9fb99bedd599ff2/distlib/scripts.py?at=default&fileviewer=file-view-default#scripts.py-139)

> And is there maybe some workaround we could employ to make virtual environments more sturdy?

I don't think there are simple workarounds. A possibility is writing a user space shebang parser just like py.exe on Windows.

HaraldNordgren:
Using pip as an example, one thing we could do is move the current `pip` to `.pip`, remove the shebang line and create a new `pip` file with the following content:

```
#!/bin/bash
/path/to/env/bin/python3 /path/to/env/bin/.pip $@
```

Basically, wrap the programs in shell scripts calling the original file to get around the problem of long shebangs.

By quoting the paths we could probably solve the problem with spaces as well.

yan12125:
OK maybe my definition for "workarounds" is too strict. I thought it shouldn't break existing things. In this case, using /bin/bash usually works but not always. Some systems do not install bash in /bin/ (FreeBSD install it at /usr/local/bin/bash), and some systems do not provide bash at all. (Most Android uses /system/bin/sh)

[...]

Issue #994, opened 2016-12-08
Pip fails with space in virtualenv path


mnm678:
Failed to execute process '/Users/.../venv/bin/pip'. Reason:
The file '/Users/.../venv/bin/pip' does not exist or could not be executed.

python 2.7 installed via homebrew
virtualenv installed via pip

yan12125:
Both Apple's XNU kernel [1] and Linux [2] cuts off interpreter name at the first whitespace. In other words, paths with whitespaces (' ' 0x20 and '\t' 0x09) can't work correctly. I suggest virtualenv forbid such paths.

[1] https://github.com/opensource-apple/xnu/blob/10.11/bsd/kern/kern_exec.c#L532-L535
[2] https://github.com/torvalds/linux/blob/v4.9-rc8/fs/binfmt_script.c#L60-L68

yan12125:
Seems this issue is identical to https://github.com/pypa/pip/issues/923

batmanav:
+1. I'm also facing this issue.

JDLH:
This looks to me like a duplicate of #53 "Whitespace in root path of virtualenv breaks scripts". 
I want to repeat the comment from https://github.com/pypa/virtualenv/issues/997#issuecomment-270681253, "virtualenv are broken with fragile kernel shebang parsing."  And in that spirit, #1014 "Not compatible with a directory having emojis in its path" is another example of virtualenv being broken by fragile kernel shebang parsing.  I'll bet the problem occurs with any non-ASCII characters in the path, I'll bet.

Maybe we should collect all three aspects of fragile kernel shebang parsing into one issue, so that we can be sure that one fix can address spaces, length, and non-ASCII characters?  I nominate #53, because it's the oldest.

gaborbernat:
You are right, thanks!

PR #1004, opened 2017-01-09
Create proxies for pip, wheel, etc. to get around problems with long paths

HaraldNordgren:
This will move `pip` to `.pip` after installation, then creates a new `pip` file containing

```
#!/usr/bin/env sh
"/path/to/env/bin/python" "/path/to/env/bin/.pip" $@
```

This eliminates the problems with long paths that stem from limitations of the shebang on Unix. This should also solve the problems with spaces in path names.

(See https://github.com/pypa/virtualenv/issues/997 for more discussion)

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -28,6 +28,7 @@ import subprocess
 import pkgutil
 import tempfile
 import textwrap
+import stat
 from distutils.util import strtobool
 from os.path import join
 
@@ -904,6 +905,34 @@ def install_wheel(project_names, py_executable, search_dirs=None,
         logger.end_progress()
 
 
+def create_proxies(py_executable, bin_dir):
+    scripts = glob.glob(os.path.join(bin_dir, "pip*"))
+    scripts += glob.glob(os.path.join(bin_dir, "easy_install*"))
+    scripts += glob.glob(os.path.join(bin_dir, "wheel*"))
+
+    execute_bits = stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH
+
+    for script in scripts:
+        head, tail = os.path.split(script)
+        hidden_destination = os.path.join(head, '.' + tail)
+
+        os.rename(script, hidden_destination)
+
+        with open(script, "w") as script_proxy:
+            text = "#!/usr/bin/env sh"
+            text += "\n'%s' '%s' $@" % (
+                py_executable,
+                hidden_destination,
+            )
+            script_proxy.write(text)
+
+        os.chmod(script, os.stat(script).st_mode | execute_bits)
+        os.chmod(
+            hidden_destination,
+            os.stat(hidden_destination).st_mode & ~execute_bits,
+        )
+
+
 def create_environment(home_dir, site_packages=False, clear=False,
                        unzip_setuptools=False,
                        prompt=None, search_dirs=None, download=False,
@@ -945,8 +974,13 @@ def create_environment(home_dir, site_packages=False, clear=False,
             download=download,
         )
 
-    install_activate(home_dir, bin_dir, prompt)
+        if not is_win:
+            create_proxies(
+                py_executable=py_executable,
+                bin_dir=bin_dir
+            )
 
+    install_activate(home_dir, bin_dir, prompt)
     install_python_config(home_dir, bin_dir, prompt)
 
 def is_executable_file(fpath):


HaraldNordgren:
Related to https://github.com/pypa/virtualenv/issues/994

virtualenv.py
```diff
@@ -904,6 +905,34 @@ def space_path2url(p):
...
+    execute_bits = stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH
+
+    for script in scripts:
+        head, tail = os.path.split(script)
+        hidden_destination = os.path.join(head, '.' + tail)
+
+        os.rename(script, hidden_destination)
+
+        with open(script, "w") as script_proxy:
+            text = "#!/usr/bin/env sh"
```
pfmoore: Should this not force an appropriate encoding? What if the venv name contains non-ASCII characters?

yan12125:
The change should be in distlib instead of virtualenv.
```
$ patch -p1 -i 1004.patch 
patching file virtualenv.py
$ python virtualenv.py "foo bar"
Using base prefix '/usr'
New python executable in /home/yen/Projects/virtualenv/foo bar/bin/python
$ source ./foo\ bar/bin/activate
(foo bar) $ pip --version   
pip 9.0.1 from /home/yen/Projects/virtualenv/foo bar/lib/python3.6/site-packages (python 3.6)
(foo bar) $ pip install nose     
Collecting nose
  Using cached nose-1.3.7-py3-none-any.whl
Installing collected packages: nose
Successfully installed nose-1.3.7
(foo bar) $ nosetests
zsh: /home/yen/Projects/virtualenv/foo bar/bin/nosetests: bad interpreter: "/home/yen/Projects/virtualenv/foo: no such file or directory
```

HaraldNordgren:
@yan12125 Oh, this is not applied to packages installed by the new virtualenv. That's bad.

Either I extend the code to do the replacement on every installed executable starting with a shebang, or we find some way to put the code deeper into the system, like distlib.

pfmoore:
@HaraldNordgren I assumed that it was a deliberate to only target the wrappers virtualenv itself installs. If it wasn't, then yes it would need to go into distlib at a minimum so that anything installed by pip would be handled. And technically into any other tool that people might use to install wrappers stuff into the virtualenv (like easy_install/setuptools). Ultimately, as it's a workaround for an OS limitation, it's up to you how far you want to go patching the various tools.

HaraldNordgren:
@yan12125 @pfmoore 

I created a pull request for the corresponding changes in `distlib`: https://bitbucket.org/pypa/distlib/pull-requests/31/create-proxy-files-pip-pip-to-get-around/diff And it seems to work really well from Virtualenv.

For development, I modified `virtualenv/virtualenv.py` with

```
    SCRIPT = textwrap.dedent("""
        import sys
        import pkgutil
        import tempfile
        import os

        import sys
        sys.path.insert(0, '/home/harald/git-repos/pip')

        import pip
        ...
```

to get virtualenv to run my custom `pip` where the `pip/_vendor/distlib` folder is replaced with a symlink to the changed distlib files.

After creating and activating the env, I also do

```
export PYTHONPATH=/home/harald/git-repos/pip
```

before trying for example `pip install flask`. Hopefully this won't have to be done if we can get these changes into an actual release on distlib.

pfmoore:
@HaraldNordgren Did you test your PR on Windows? Because I saw it and had a quick look, and it doesn't look like it limits the change to Windows. I'd comment on the PR itself, but I don't have access to bitbucket right at the moment,,,

HaraldNordgren:
I has not been tested on Windows, only Ubuntu 16.04.

I was hoping that the `use_launcher` logic (https://bitbucket.org/pypa/distlib/src/f702f6afc1d03717fa88a742b9fb99bedd599ff2/distlib/scripts.py#scripts.py-206) would isolate this change to Unix.

pfmoore:
Hmm, OK. Someone will have to test on Windows, I guess.

HaraldNordgren:
Maybe here is an even better solution, based on an idea from a Mach bootstrap script: https://bitbucket.org/pypa/distlib/pull-requests/32

pfmoore:
IMO, this is an immensely complex (and hence difficult to maintain) workaround to an OS limitation of certain Unix systems, that comes up relatively rarely. While I agree that it fails in a pretty unpleasant way, and it would be good to do *something*, I'd rather any workaround be limited to cases where it's needed - which is of course hard to determine precisely because it's an OS issue rather than a pip/distlib/virtualenv issue.

On the other hand, I don't have any direct interest in the solution, as I am a Windows user and this issue doesn't occur on Windows, so if the Unix developers are happy with the proposed solution, I won't object (as long as it is only a Unix solution, and Windows behaviour remains as it currently is).

HaraldNordgren:
I think you have a point that we could limit the fix to the situations where it breaks. I tried to change the Bitbucket PR to achieve that. So only pad the shebang if it would break otherwise. I think that also highlights the need to the PR -- without it distlib actually breaks.

Since we are dealing with shebangs, I am fairly certain that it won't affect Windows.

yan12125:
> Since we are dealing with shebangs, I am fairly certain that it won't affect Windows.

Just FYI: py.exe on Windows (PC\launcher.c) uses shebangs.

HaraldNordgren:
Ok, good to know.

I'll keep working on this. I don't think Vinay Sajip at Distlib will accept the change there, but he pointed me to `ScriptMaker.exectuable` which should be able to achieve the same thing. So I submitted another pull reuqest for `pip` instead:

https://github.com/pypa/pip/pull/4237

Issue #565, opened 2014-02-12
Breakage in --always-copy - fails with: ImportError: No module named time


ionelmc:
On CentOS 6.5 (python2.7 from SCL):

```
[dev@centbox core]$ virtualenv-2.7 --always-copy .break
sys.path (patched logging) = ['/opt/rh/python27/root/usr/bin', '/opt/rh/python27/root/usr/lib/python2.7/site-packages/setuptools-2.2-py2.7.egg', '/opt/rh/python27/root/usr/lib/python2.7/site-packages/wheel-0.22.0-py2.7.egg', '/opt/rh/python27/root/usr/lib/python2.7/site-packages/virtualenv-1.11.2-py2.7.egg', '/opt/rh/python27/root/usr/lib/python2.7/site-packages/pip-1.5.2-py2.7.egg', '/opt/rh/python27/root/usr/lib64/python27.zip', '/opt/rh/python27/root/usr/lib64/python2.7', '/opt/rh/python27/root/usr/lib64/python2.7/plat-linux2', '/opt/rh/python27/root/usr/lib64/python2.7/lib-tk', '/opt/rh/python27/root/usr/lib64/python2.7/lib-old', '/opt/rh/python27/root/usr/lib64/python2.7/lib-dynload', '/opt/rh/python27/root/usr/lib64/python2.7/site-packages', '/opt/rh/python27/root/usr/lib/python2.7/site-packages']
Cannot find file lib (bad symlink)
New python executable in .break/bin/python2.7
Also creating executable in .break/bin/python
Installing setuptools, pip...
  Complete output from command /home/dev/core/.break/bin/python2.7 -c "import sys, pip; sys...d\"] + sys.argv[1:]))" setuptools pip:
  Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/virtualenv-1.11.2-py2.7.egg/virtualenv_support/pip-1.5.2-py2.py3-none-any.whl/pip/__init__.py", line 9, in <module>
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/virtualenv-1.11.2-py2.7.egg/virtualenv_support/pip-1.5.2-py2.py3-none-any.whl/pip/log.py", line 6, in <module>
  File "/opt/rh/python27/root/usr/lib64/python2.7/logging/__init__.py", line 26, in <module>
    import sys, os, time, cStringIO, traceback, warnings, weakref
[...]

Issue #985, opened 2016-11-17
Failure to install wheel when using --always-copy


terminalmage:
This is on Python 2.7.12 and virtualenv 15.0.3, both installed on CentOS 5 and packaged by the [IUS Community](https://ius.io/) project.

When I try to create a virtualenv with ``--always-copy``, it fails with the following error:

```
# virtualenv-2.7 --no-site-packages --always-copy foo
Cannot find file lib (bad symlink)
New python executable in /root/foo/bin/python2.7
Also creating executable in /root/foo/bin/python
Installing setuptools, pip, wheel...
  Complete output from command /root/foo/bin/python2.7 - setuptools pip wheel:
  Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/usr/lib64/python2.7/tempfile.py", line 32, in <module>
    import io as _io
  File "/usr/lib64/python2.7/io.py", line 51, in <module>
    import _io
ImportError: No module named _io
----------------------------------------
...Installing setuptools, pip, wheel...done.
Traceback (most recent call last):
  File "/usr/bin/virtualenv-2.7", line 9, in <module>
    load_entry_point('virtualenv==15.0.3', 'console_scripts', 'virtualenv')()
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 710, in main
    symlink=options.symlink)
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 946, in create_environment
    download=download,
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 902, in install_wheel
    call_subprocess(cmd, show_stdout=False, extra_env=env, stdin=SCRIPT)
  File "/usr/lib/python2.7/site-packages/virtualenv.py", line 794, in call_subprocess
    % (cmd_desc, proc.returncode))
OSError: Command /root/foo/bin/python2.7 - setuptools pip wheel failed with error code 1
```

Tracking this down in virtualenv.py, this is caused by ``install_wheel()`` running an embedded script using a munged pythonpath (assembled [here](https://github.com/pypa/virtualenv/blob/15.0.3/virtualenv.py#L837-L838) and containing only the paths of the wheels found by ``find_wheels()``). The problem with this is that the embedded python script which ``install_wheel()`` wants to run imports ``tempfile``, which eventually needs to import ``_io.so`` from ``/usr/lib64/python2.7/lib-dynload``.

[...]

PR #1010, opened 2017-02-01
Supposed typo: https://github.com/pypa/virtualenv/issues/565#issuecom…

Neurocinetics:
Supposed typo: [https://github.com/pypa/virtualenv/issues/565#issuecomment-115036122](https://github.com/pypa/virtualenv/issues/565#issuecomment-115036122) On CentOS 7 building certain projects fails without this "fix". Also relevant, [https://github.com/spotify/dh-virtualenv/issues/150](https://github.com/spotify/dh-virtualenv/issues/150) and [https://github.com/numpy/numpy/issues/7570](https://github.com/numpy/numpy/issues/7570).

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1538,9 +1538,9 @@ def fix_lib64(lib_dir, symlink=True):
     if os.path.lexists(lib64_link):
         return
     if symlink:
-        os.symlink('lib', lib64_link)
+        os.symlink(lib_dir, lib64_link)
     else:
-        copyfile('lib', lib64_link)
+        copyfile(lib_dir, lib64_link)
 
 def resolve_interpreter(exe):
     """


brycepg:
This commit fixed an error for me using the `--always-copy` flag on Centos 7.3 with virtualenv version 15.1.0

Here's the traceback which no longer occurs:
    [user]~/% virtualenv -p $(which python3) dist --always-copy
    Running virtualenv with interpreter /usr/bin/python3
    Using base prefix '/usr'
    Cannot find file lib (bad symlink)
    New python executable in /home/user/dist/bin/python3
    Also creating executable in /home/user/dist/bin/python
      Complete output from command /home/user/dist/bin/python3 - setuptools pip wheel:
      File "<stdin>", line 4, in <module>
      File "/home/user/dist/lib64/python3.4/tempfile.py", line 34, in <module>
        import shutil as _shutil
      File "/home/user/dist/lib64/python3.4/shutil.py", line 14, in <module>
        import tarfile
      File "/home/user/dist/lib64/python3.4/tarfile.py", line 47, in <module>
        import time
    ImportError: No module named 'time'

Hopefully it gets merged. Thank you.

spacefito:
This patch also fixes this issue: https://github.com/pypa/virtualenv/issues/985
please merge is right.

leondutoit:
+1 for merging this - I need this fix on Centos 7. Good job on finding the root cause!

pfmoore:
The comment in https://github.com/pypa/virtualenv/issues/565#issuecomment-149987121 seems to be saying that this fix will cause issues on some other systems (CentOS 6, AIUI). Any fix that gets merged should work everywhere, not just trade a failure on one system for a failure on another.

aelnaggar:
PLEASE MERGE IT

pfmoore:
@aelnaggar Shouting at the developers is neither polite nor likely to be productive. We're aware of the potential benefit, but haven't a clear picture of the potential issues on other systems yet. It's extremely hard to maintain virtualenv precisely because of the various distribution idiosyncracies we have to cater for, and the virtualenv developers (who are all volunteers) have little enough time as it is.

If this issue is important enough to you, you're perfectly welcome to maintain this change as a local patch to your copy of virtualenv. Otherwise please be patient and respect the fact that you are getting virtualenv for free, as a result of others' volunteer efforts.

Neurocinetics:
@pfmoore Should I/we change the pull request, so that the updated pull request can distinguish a Debian system from a RHEL system? That would be a more appropriate fix right?

joesanford:
If any updates are required to get this PR merged in, please let me know and I will gladly take over dev responsibility for this PR. If what is mentioned in comment #565 is what is needed, please let me know and I can ensure this only is run on affected distros.

pfmoore:
I don't honestly know, myself. I'm not familiar with the differences between the various Linux distros, and how this change would need to cover that. Someone needs to do the due dilligence on how things vary between the various platforms (RHEL/CentOS, Fedora, Debian/Ubuntu, SuSE, Gentoo, Arch, ...?), and confirm that. Then I guess one of the virtualenv maintainers who does work with Linux should review it and take it from there.

(The PR is described as a "typo fix" - it may well be that to someone who knows the platforms, it's "obviously right", but I can't say on that).

joesanford:
From what I can see in the suggested change, it's not really a "typo fix" and moreso branching behavior based on which distro is currently being used. If that's the case, I could add in the logic around `platform.dist()` (https://docs.python.org/3.7/library/platform.html#platform.dist) to only have this run on RHEL/CentOS/SuSe.

Thoughts? I can have an update to this PR this evening.

gaborbernat:
Seems already done in https://github.com/pypa/virtualenv/blob/master/src/virtualenv.py#L1622-L1625

Issue #1015, opened 2017-02-08
virtualenv should also copy shared libraries? 


woosley:
see this patch for pyenv

https://github.com/yyuu/pyenv/commit/2657f1049cd45656918f601096509957d5b74e7c

When compile python with '--enable-shared -Wl,-rpath='\$\$ORIGIN/../lib', the created python can be dropped to any folder. 

But after  virtualenv copy the python binary to a different folder, the relative path "bin/../lib" to libpython.so will be broken, the python binary will die with 

```
/root/test/bin/python3.5: error while loading shared libraries: libpython3.5m.so.1.0: cannot open shared object file: No such file or directory
```

In this case, virtualenv should also consider copy the shared object file to the new folder.

yan12125:
I've just installed pyenv and reverted yyuu/pyenv@afbd54dd13ff2ff54280e1376dced74c5e033cd7, but there's still no rpath set. How to install python with rpath?

yyuu:
It'll set rpath if PYTHON_CONFIGURE_OPTS includes --enable-shared.

https://github.com/yyuu/pyenv/blob/master/plugins/python-build/bin/python-build#L1914

yan12125:
OK reproduced.

By the way, if you have no need on Python < 3.3, I suggest the standard [venv](https://docs.python.org/3/library/venv.html) library, which is much better in handling such cases.

woosley:
I do need Python < 3.3. 
What is the best approach for this issue? maybe an option --with-shared? if so, I can make a patch.

yan12125:
It's better to have things auto-detected.

nnnn20430:
what is the point of copying the binary anyway? why not copy a script like this
`
#!/bin/bash
exec "-a" "$0" "python2.7" "$@"
`
replacing "python2.7" with what virtualenv -p got
and would have to export _OLD_VIRTUAL_PATH so that script can also use old env after using activate script, otherwise it goes into a infinite loop

woosley:
this is a bit tricky 
the only solid way to detect $ORIGIN is readelf
```
[root@maddog 2.7.9]# readelf -d  bin/python|grep  RPATH
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../lib]
```
I am wondering in what case this readelf will be not available

yan12125:
[...]

PR #1019, opened 2017-02-20
install libpython under $ROOT/lib

woosley:
This patch fixes https://github.com/pypa/virtualenv/issues/1015

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -924,6 +924,7 @@ def create_environment(home_dir, site_packages=False, clear=False,
         home_dir, lib_dir, inc_dir, bin_dir,
         site_packages=site_packages, clear=clear, symlink=symlink))
 
+    install_shared(bin_dir, symlink=symlink)
     install_distutils(home_dir)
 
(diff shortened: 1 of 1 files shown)


gst:
any idea when this can be applied ?

virtualenv.py
```diff
@@ -1478,6 +1479,13 @@ def install_python_config(home_dir, bin_dir, prompt=None):
     for name, content in files.items():
         make_exe(os.path.join(bin_dir, name))
 
+def install_shared(bin_dir, symlink=True):
+    current_shared = os.path.realpath(join(os.path.dirname(sys.executable), "..", "lib"))
```
Ivoz: Is this going to exist on all platforms (MacOS, Windows, Linus) / most distros? This function only seems to care about nixxy platforms AFAICT.

woosley: I think my original idea is that its existence does not matter, because if it does not exist, the next `glob.glob` would return empty list then nothing will be copied.

PR #1023, opened 2017-02-23
Copy libs/pythonXY.lib to support building of C extensions without distutils

jcfr:
This issue was discovered while using virtualenv to create an
isolated environment and test that a source distribution would compile
without using the knowledge hardcoded in `distutils.sysconfig` and
`virtualenv/virtualenv_embedded/distutils-init.py`

In other word, without relying on the fact virtualenv monkey patch distutils
to ensure the value associated with `LIBDIR` variable is correct.

This is relevant because the library location is implicitly specified
in PC/pyconfig.h

```
[...]

/* For an MSVC DLL, we can nominate the .lib files used by extensions */
                        /* So MSVC users need not specify the .lib file in
                        their Makefile (other compilers are generally
                        taken care of by distutils.) */

[...]
```

References:
 * https://docs.python.org/2/extending/windows.html#using-dlls-in-practice
 * https://docs.python.org/3.6/extending/windows.html#using-dlls-in-practice

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1395,6 +1395,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
                 elif os.path.exists(python_dll_d_dest):
                     logger.info("Removed %s as the source does not exist", python_dll_d_dest)
                     os.unlink(python_dll_d_dest)
+
+            if not IS_PYPY:
+                # considering that on windows, python import libraries are located in
+                # the  "<root>/libs" directory, the following code will look for and
+                # copy "pythonXY.lib".
+                pythonlib_name = "python{}{}.lib".format(sys.version_info[0], sys.version_info[1])
+                pythonlib = os.path.join(os.path.dirname(sys.executable), "libs", pythonlib_name)
+                pythonlib_dest_dir = os.path.join(lib_dir, "..", "libs")
+                pythonlib_dest = os.path.join(pythonlib_dest_dir, pythonlib_name)
+                if os.path.exists(pythonlib):
+                    logger.info("Also created %s" % pythonlib_name)
+                    mkdir(pythonlib_dest_dir)
+                    shutil.copyfile(pythonlib, pythonlib_dest)
+
         if IS_PYPY:
             # make a symlink python --> pypy-c
             python_executable = os.path.join(os.path.dirname(py_executable), "python")


macisamuele:
Is there any interest on merging this PR?
I'm currently working on a small project with rust bindings on Windows and I'm not able to build it within a virtual environment. Applying those changes on top of the latest version of `virtualenv` I can finally build them.

Pinging @dstufft as one of the project moderator recently active.
If you think that this PR could be "mergeable" but it needs testing etc. let me know and I'll try to give some support

jcfr:
@macisamuele Here is a [rebased topic](https://github.com/pypa/virtualenv/compare/master...jcfr:windows-copy-import-library-rebased), it it works well, I will push force this topic.

It would indeed be nice to get this integrated.

macisamuele:
I've spent the last few days experimenting in building rust from a venv with no success, I've applied the patch from this PR on the current master and it worked on the first try 🎉

jcfr:
Great :+1:  Could you review this topic then: https://github.com/pypa/virtualenv/compare/master...jcfr:windows-copy-import-library-rebased ? A sanity check to make sure I properly rebased it.

jcfr:
Thanks @macisamuele , topic updated :smile:

gaborbernat:
Closing due to inactivity.

jcfr:
@gaborbernat I would appreciate if you could give this a little more time. Thanks for your patience.

jcfr:
Topic has been rebased and tested locally.

I confirm that the `pythonXY.lib` file is copied for all version of python:

```
$ find . | grep "\.lib$"
./venv-test-27-x64/libs/python27.lib
./venv-test-27-x86/libs/python27.lib
./venv-test-34-x64/libs/python34.lib
./venv-test-34-x86/libs/python34.lib
./venv-test-35-x64/libs/python35.lib
./venv-test-35-x86/libs/python35.lib
./venv-test-36-x64/libs/python36.lib
./venv-test-36-x86/libs/python36.lib
./venv-test-37-x64/libs/python37.lib
./venv-test-37-x86/libs/python37.lib
```

@gaborbernat I was thinking to update the test [test_install_python_bin](https://github.com/pypa/virtualenv/blob/59a5ab8ce460de950f97ffba8e1af6d79d530aa9/tests/test_virtualenv.py#L281) ? What do you think ?

virtualenv.py
```diff
@@ -1395,6 +1395,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
                 elif os.path.exists(python_dll_d_dest):
                     logger.info("Removed %s as the source does not exist", python_dll_d_dest)
                     os.unlink(python_dll_d_dest)
+
+            if not IS_PYPY:
+                # considering that on windows, python import libraries are located in
+                # the  "<root>/libs" directory, the following code will look for and
+                # copy "pythonXY.lib".
+                pythonlib_name = "python{}{}.lib".format(sys.version_info[0], sys.version_info[1])
```
gaborbernat: slight name tweak, ``python_lib_name`` (the logic applies to all other names too)

virtualenv.py
```diff
@@ -1395,6 +1395,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
...
+                pythonlib = os.path.join(os.path.dirname(sys.executable), "libs", pythonlib_name)
+                pythonlib_dest_dir = os.path.join(lib_dir, "..", "libs")
+                pythonlib_dest = os.path.join(pythonlib_dest_dir, pythonlib_name)
+                if os.path.exists(pythonlib):
+                    logger.info("Also created %s" % pythonlib_name)
```
gaborbernat: Do not materialize the string yet, leave it extra arg to logger.

virtualenv.py
```diff
@@ -1395,6 +1395,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
                 elif os.path.exists(python_dll_d_dest):
                     logger.info("Removed %s as the source does not exist", python_dll_d_dest)
                     os.unlink(python_dll_d_dest)
+
+            if not IS_PYPY:
```
gaborbernat: Why just pypy, what about jython? We should also mention why this is so special

virtualenv.py
gaborbernat: is it lib on Windows too?

virtualenv.py
```diff
@@ -1395,6 +1395,20 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
...
+                # considering that on windows, python import libraries are located in
+                # the  "<root>/libs" directory, the following code will look for and
+                # copy "pythonXY.lib".
+                pythonlib_name = "python{}{}.lib".format(sys.version_info[0], sys.version_info[1])
+                pythonlib = os.path.join(os.path.dirname(sys.executable), "libs", pythonlib_name)
+                pythonlib_dest_dir = os.path.join(lib_dir, "..", "libs")
```
gaborbernat: Shouldn't this take the lib_dir value from https://github.com/pypa/virtualenv/blob/508919a789dcb43469bdfbdcbf9d98546b4c076c/virtualenv.py#L1095?

virtualenv.py
jcfr: `.lib` is indeed the extension on Windows. [Test above](#issuecomment-454175220) was performed on windows and the directory were searched running `find`  and `grep` from Git Bash.

virtualenv.py
jcfr: The `.lib` file indeed lives in the `.libs` folder.

image

virtualenv.py
jcfr: * Jython is java-based and doesn't provide import libraries like CPython does.
* Pypy indeed provides a `libs` folder with the corresponding `pythonXY.lib` file, I will update the patch to not exclude it.

gaborbernat:

jcfr:
Hi @gaborbernat , 

Last time I worked on this PR, I rebased, tested locally and answering the questions.

To move this forward and hopefully get it integrated, let me know what would be required.

gaborbernat:
I'd like to validate this on the new rewrite branch to be honest, though not sure is advanced enough to add this.

gaborbernat:
created https://github.com/pypa/virtualenv/issues/1448 to track this

PR #1024, opened 2017-03-02
Fixed small paths typos in userguide.rst

kevinfs:
I've read that PRs should be made against `develop`, but there is no such branch anymore.
If I'm wrong in any way, feel free to redirect me / advise on what you need !
Thanks.


gaborbernat:
This has been addressed as part of another PR, thanks for the contribution!

PR #1027, opened 2017-03-08
Do not add /usr/lib64 to sys.path if it is a symlink of /usr/lib.

nonZero:
Under arch linux, /usr/lib64 is a symlink of /usr/lib.
currently site.py will cause the inclusion of both of them in sys.path.

CAUTION: This patch was tested ONLY on arch linux, and not on any other platform.


gst:
@nonZero : the patch has been tested via TravisCI on all supported pythons as you can see here: https://travis-ci.org/pypa/virtualenv/builds/209142967 ;)

gaborbernat:
Closing due to inactivity.

PR #1032, opened 2017-04-04
Enabled osx targets on Travis with allow_failures

ssbarnea:
Signed-off-by: Sorin Sbarnea <ssbarnea@redhat.com>


yan12125:
https://github.com/travis-ci/cpython-builder/pull/3 should be fixed first so that people can test Python projects properly on macOS.

PR #1037, opened 2017-04-16
Use "__BIN_NAME__" in "activate.ps1".

Antigroup:
Before it always used "Scripts", which didn't work with pypy virtual
environments because they put everything in "bin".


gaborbernat:
Done via #1229

PR #1041, opened 2017-04-18
Fix multilib support, lib64 no longer hardcoded

JamesTheAwesomeDude:
This very necessary fix, previously made by @xen0n at #516 had, sadly bit-rotted before it could be merged.

I have taken a stab at porting it to the current state of virtualenv as part of my efforts at MoonchildProductions/Pale-Moon#1035.

This seems to remove the last of the hardcoded references to lib64; `grep -r lib64` now only returns entries from `docs/changes.rst`!


JamesTheAwesomeDude:
I have read the `CONTRIBUTING.rst` file:

> virtualenv uses git-flow to coordinate development. The latest stable version should exist on the master branch, and new work should be integrated to develop.

However, there IS no `develop` branch; otherwise, I would have pushed it there.

JamesTheAwesomeDude:
Hmm, I think I may need some help getting pypy working…

rephlex:
This patch fixed my virtualenv-16.0.0 on gentoo mips n32.

Before it would not even survive
mkdir foo && virtualenv foo/
without panicking about pkgutils missing.

Kokokokoka:
This is sad, that this changeset was lost. hardcoding lib** is generally a bad idea.

gaborbernat:
@Kokokokoka if you check with version 20.x you should not have this problem (assuming your python is configured correctly).

PR #1049, opened 2017-05-30
Activate newly created virtuale env if --activate is passed

pygaur1:
while creating a new virtualenv if we pass --activate as option then it will activate virtual env just after creation.
**Example :**
`prashant:~$ virtualenv my_env --activate`
`(my_env) prashant:~$`


gaborbernat:
We don't know the users shell type so no sensible way of doing this.

PR #1063, opened 2017-07-09
Discover python.exe with -ppython on windows

### LINKED ISSUE #1062: Windows `virtualenv -p python2.7` fails even when `python2.7.exe` exists
The problem stems from [this function](https://github.com/pypa/virtualenv/blob/6d8a4ed81f40223a3f9ea1ba5d11be500d9f667a/virtualenv.py#L1542-L1565) which doesn't honor the windows-specific `PATHEXT`.

Here's a nearly-identical implementation of that function which *does* handle `PATHEXT` expansion: https://github.com/pre-commit/pre-commit/blob/e2bae300fe2794e2ece25d4cd72127238704bb1e/pre_commit/parse_shebang.py#L21-L42

`distutils.spawn.find_executable` has a similar special case, but only attempts `.exe`: https://github.com/python/cpython/blob/9648088e6ccd6d0cc04f450f55628fd8eda3784c/Lib/distutils/spawn.py#L181-L182

asottile:
Resolves #1062

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -22,6 +22,7 @@ import logging
 import zlib
 import errno
 import glob
+import distutils.spawn
 import distutils.sysconfig
 import struct
 import subprocess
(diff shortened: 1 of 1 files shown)


asottile:
@dstufft thoughts / feedback on this?

pfmoore:
See my comment on #1062. ```virtualenv -p "2.7"``` finds Python 2.7 direct from the registry entries, the same way as ```py -2.7``` works, so there should be no need for this.

Issue #1029, opened 2017-03-17
activate may fails due to unbound variables when run with set -eu


ssbarnea:
It seems that an old bug is still here: `PS1: unbound variable` when called with bash strict mode.

Clearly virtualenv needs a test using bash strict mode and an almost empty set of environment variables, so we avoid future regressions on this.

**Please note that this bug does reproduce with ANY supported Unix shell, not only `bash`, including `ksh` and `zsh`.**

Here is a simple way to replicate the bug:
```
#!/bin/bash
# same applies to any other bourne compatible shells (is not bash specific)
set -euox pipefail
pip install -U virtualenv
virtualenv xxx
unset PS1
source xxx/bin/activate
```

The workaround, while ugly, is to do prepend `PS=${PS:-} ` to the activate line, which defines PS as empty string when is not already defined, or keeps its value when is defined.

The same kind of [bug applies to Python version of venv](https://bugs.python.org/issue25351) and there is an already [open PR to fix it there too](https://github.com/python/cpython/pull/3804).

Please do not postpone/delay implementing a fix just because the same kind of bug exists in other places. Note that the **default expansion variable syntax is POSIX compatible** and not something new or fancy, the fact that this was not known to the original author should not be an excuse for not using it.

johnomics:
I am also finding this, running virtualenv 15.1.0. I'm using an environment within a Nextflow pipeline, and Nextflow runs in strict mode by default (https://github.com/nextflow-io/nextflow/issues/302). While Nextflow can be reconfigured to run without strict mode, I agree with the Nextflow developer that it would be preferable to avoid using unbound variables, if possible.

I'm not sure exactly how the `activate` script is created, but if it comes from [activate.sh](https://github.com/pypa/virtualenv/blob/263283be88061b965ead9ebff47c37021d7d3f62/virtualenv_embedded/activate.sh), then the fix might be as simple as changing `$PS1` on lines 57, 59 and 61 to `${PS1-}`. (This syntax will use the value of `PS1` if available, and the empty string otherwise. It doesn't change the value of `PS1`. [Documentation](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02)). At least, if I modify the generated `activate` script in my environment like this, the error message goes away.

johnomics:
https://github.com/pypa/virtualenv/issues/991

jakub-bochenski:
will the project maintainers even acknowledge this issue exists?

[...]

PR #1077, opened 2017-09-08
add test for activation failure

ssbarnea:
Covers case where activate may fail due
to unbound variables.

Relates to #1029

Change-Id: Ib10d364d9a273429ab7bce22aa5201571aecb349


ssbarnea:
Closing because there is another better PR addressing the same issue.

PR #1078, opened 2017-09-08
Avoid unbound PS1 failure

### LINKED ISSUE #1029: activate may fails due to unbound variables when run with set -eu

ssbarnea:
fixes #1029

Change-Id: I764297341dee69a01bd4044d6567dee1026499d2


pfmoore:
The embedded copy of the script needs refreshing. Just changing the raw script isn't sufficient - see `bin/rebuild-script.py`.

ssbarnea:
@pfmoore Thanks for the hint. I am looking now and I will also try to improve the testing, maybe I can do it in a single PR that also improves the testing.

virtualenv_embedded/activate.sh
```diff
@@ -1,5 +1,6 @@
 # This file must be used with "source bin/activate" *from bash*
 # you cannot run it directly
+set -x
```
jakub-bochenski: Not sure if people would appreciate suddenly getting debug output from activate

If we go with it we should restore the previous setting on exit at least

virtualenv_embedded/activate.sh
```diff
@@ -54,11 +60,12 @@ if ! [ -z "${PYTHONHOME+_}" ] ; then
 fi
 
 if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
-    _OLD_VIRTUAL_PS1="$PS1"
+    _OLD_VIRTUAL_PS1="${PS1:-}"
```
jakub-bochenski: I think it's better to go with `${PS1-}` we just want to guard against unset variables. Why add the side-effect of setting PS1?

And look: this construct is already employed in line 62 above --this should remove any concerns about portability

virtualenv_embedded/activate.sh
```diff
@@ -54,11 +60,12 @@ if ! [ -z "${PYTHONHOME+_}" ] ; then
...
+    # __VIRTUAL_PROMPT__ is not a variable, is replaced at compile time
     if [ "x__VIRTUAL_PROMPT__" != x ] ; then
-        PS1="__VIRTUAL_PROMPT__$PS1"
+        PS1="__VIRTUAL_PROMPT__${PS1:-}"
```
jakub-bochenski: ditto

virtualenv_embedded/activate.sh
```diff
@@ -54,11 +60,12 @@ if ! [ -z "${PYTHONHOME+_}" ] ; then
...
     else
-        PS1="(`basename \"$VIRTUAL_ENV\"`) $PS1"
+        PS1="(`basename \"$VIRTUAL_ENV\"`) ${PS1:-}"
```
jakub-bochenski: ditto

virtualenv_embedded/activate.sh
ssbarnea: @jakub-bochenski Out of curiosity, is there a difference between `${FOO-}` and `${FOO:-}`? I used the `:-` for years and that was the first time I encountered the simple minus.

jakub-bochenski: Haste makes waste -- I was wrong :)

The difference between `:-` and `-` is in treatment of null values, none of those does an assignment (you need either `=` or `:=`). Reference: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

Still I think it might be better to use `-` because
 - it's already used
 - it only affects unset variables and leaves null as null, so it's a tighter fit

pradyunsg:
@ssbarnea Travis and AppVeyor do not approve.

jakub-bochenski:
@pradyunsg please note this requires #1087 to be merged first

piotr-dobrogost:
@pradyunsg 
Could you please help move this forward?
Taking into consideration there appears to be no active maintainer I think mixing a fix with other changes (improving testing) here does not help getting this merged…

ssbarnea:
@piotr-dobrogost I think you need to comment/review https://github.com/pypa/virtualenv/pull/1087 and try to persuade the team on irc and mailing list to put some love into virtualenv.

About one year ago I started some discussions that didn't prove to be fruitless. Who knows, maybe this year we would be able to do something.

Issue #906, opened 2016-05-24
venv/bin/python does not set VIRTUAL_ENV (and PATH) environment


muelli:
I was surprised to find out that there is a difference between sourcing the bin/activate file (which seems to be only compatible to Bash and Fish) and running bin/python directly.

I expected the latter to work equally well.  It is unfortunately much easier to call a specific python interpreter  than to set up the environment when dealing with multiple virtual environments.

The way I read the current documentation, it is also the expected behaviour that it "just works" when calling scripts inside the virtual env directly. (citing https://virtualenv.pypa.io/en/latest/userguide/, emphasis mine)

> This will change your $PATH so its first entry is the virtualenv’s bin/ directory. (You have to use source because it changes your shell environment in-place.) This is all it does; it’s purely a convenience. If you **directly run** a script or the python interpreter from the virtualenv’s bin/ directory (e.g. path/to/ENV/bin/pip or /path/to/ENV/bin/python-script.py) there’s **no need for activation**.

This, however, is not completely true. If you `subprocess.call(pip)` from an activated python, you get different results from calling it from venv/bin/python.  This is probably what lead to https://github.com/fiduswriter/fiduswriter/issues/195. The difference is at least in the `VIRTUAL_ENV` and `PATH` variables. But other people also have this problem, it seems:
http://stackoverflow.com/questions/11963019/running-python-script-from-inside-virtualenv-bin-is-not-working
http://stackoverflow.com/questions/8052926/running-subprocess-within-different-virtualenv-with-python
http://stackoverflow.com/questions/3287038/cron-and-virtualenv

This issue seems to have cropped up here (https://github.com/pypa/virtualenv/issues/247) albeit in a different flavour. 

It'd be nice if this could be addressed and a "one-stop" solution be provided (rather than the two-step source and call).

Ivoz:
Yes, the subprocess is creating a **new** environment for the execution of the call, one in which your modified PATH (as set by an activate script) is no longer present.

When running directly from `venv\bin\python` you are coming across a coincidence that since `venv\bin` is the current working directory, the first `pip` found is the one find inside there.

For instance, one example of solving this would be passing the current `PATH`  through the `env=` dictionary as an argument one of the `subprocess` module functions, so that the so-called 'convenience' of the activate script flows through to the new subprocess.

muelli:
I don't follow your statement. 

>
[...]

Issue #124, opened 2011-04-21
scripts outside a virtualenv that use entry points can't find registrations inside the virtualenv


slinkp:
What the title says. If you have a python script whose shebang line points to an interpreter other than the virtualenv's interpreter, it won't find packages installed in the virtualenv that register things for entry points that it uses.

For detailed example see 
http://trac.pythonpaste.org/pythonpaste/ticket/458

Is this actually a virtualenv bug? or a setuptools bug? or what?

PR #1080, opened 2017-09-15
Clarify activation is required for environment vars

follower:
After spending many hours trying to work out why Ansible inventory scripts do not work correctly when Ansible is installed via a recent version of Homebrew (which uses virtualenv) I believe at least part of the underlying cause is the result of a lack of clarity in this section of the documentation.

As I understand it, saying "...there's no need for activation" is incorrect if a Python script spawns subprocesses or otherwise relies on `PATH` and `VIRTUAL_ENV` environment variables being modified in the same manner as the activation scripts.

When Ansible is installed within a virtualenv, inventory scripts featuring a `#!/usr/bin/env python` shebang end up using a non-virtualenv Python binary which can then not find required library files.

The original wording implies that using an activation script and then executing `python` is the same as running `/<virtualenv>/bin/python` directly--which is not the case.

I eventually discovered that according to https://github.com/pypa/virtualenv/issues/906#issuecomment-244394963 this behavior is intentional:

>  the Python executable doesn't set the PATH or VIRTUAL_ENV environment variables. That's by design, and is not a bug.

I note that is followed by the remark:

> The only thing you might be able to get support for is adding some comments to the documentation to clarify that you shouldn't be writing scripts that assume sys.executable is on $PATH, or that VIRTUAL_ENV is set 

Hence why I am proposing this patch.

### Specific example

I was trying to use the Amazon EC2 inventory script (per <http://docs.ansible.com/ansible/latest/intro_dynamic_inventory.html#example-aws-ec2-external-inventory-script>) like so:

       ansible -i ec2.py --list-hosts all

The result was an error like this (because `ec2.py` was executed via a non-virtualenv Python binary which did not have the `ansible` module installed):

       ERROR! Attempted to execute "ec2.py" as inventory script: Inventory script (ec2.py) had an execution error: Traceback (most recent call last):
         File "/<path>/ec2.py", line 135, in <module>
           from ansible.module_utils import ec2 as ec2_utils
       ImportError: No module named ansible.module_utils

### Possible approach for adding the implied functionality

Additional research suggests the implied functionality is possible without modification of the Python binary nor the use a wrapper script via the addition of a `sitecustomize.py` file but I'll discuss that in another issue later.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -52,10 +52,20 @@ separate activate files for some other shells, like csh and fish.
 This will change your ``$PATH`` so its first entry is the virtualenv's
 ``bin/`` directory. (You have to use ``source`` because it changes your
 shell environment in-place.) This is all it does; it's purely a
-convenience. If you directly run a script or the python interpreter
+convenience.
+
+If you directly run a script or the python interpreter
 from the virtualenv's ``bin/`` directory (e.g. ``path/to/ENV/bin/pip``
-or ``/path/to/ENV/bin/python-script.py``) there's no need for
-activation.
+or ``/path/to/ENV/bin/python-script.py``) then ``sys.path`` will
+automatically be set to use the Python libraries associated with the
+virtualenv. But, unlike the activation scripts, the environment variables
+``PATH`` and ``VIRTUAL_ENV`` will *not* be modified. This means that if
+your Python script uses e.g. ``subprocess`` to run another Python script
+(e.g. via a ``#!/usr/bin/env python`` shebang line) the second script
+*may not be executed with the same Python binary as the first* nor have 
+the same libraries available to it. To avoid this happening your first
+script will need to modify the environment variables in the same manner
+as the activation scripts, before the second script is executed.
 
 The ``activate`` script will also modify your shell prompt to indicate
 which environment is currently active. To disable this behaviour, see


pfmoore:
While I understand your problem, and I agree that some clarification in the docs would be useful, I'd prefer different wording. In my opinion, the Ansible scripts that use `#!/usr/bin/env python` are *wrong*, unless the Ansible documentation makes it clear that you need the correct Python interpreter on your `PATH` (in which case, running via a non-activated virtualenv's interpreter is not following their documentation). Typically when running a Python subprocess, you should use `subprocess.call([sys.executable, ...])` to be sure you're using the same interpreter as your main executable, if that's what you want.

Having said all this, I'd be happy with a comment in the documentation saying that some 3rd party scripts assume that Python is in `PATH` or that `VIRTUAL_ENV` is set - if you want to use such scripts without activating your virtualenv, you will need to ensure that you manually set the appropriate environment variables in your code.

follower:
Thanks for taking the time to respond. :)

----

### TL;DR:

**If the auto-activation of a virtual environment by the Python interpreter doesn't fit with your model of the project, I can proceed with a revision of the documentation patch.**

The remainder of the reply tries to identify where our mental models differ and either gain better understanding of your perspective or try to better communicate my perspective. :)

----

### The start 

> I agree that some clarification in the docs would be useful, I'd prefer different wording.

I had meant to mention that I wasn't attached to the exact wording I suggested, it was more intended as a starting point. :) I think the minimum would be to at least remove the phrase "there's no need for activation".


### Mental models of virtual environments and activation

I'm still somewhat confused about your perspective on the issue but it seems to me to come down to us each having a different mental model about:

 1. how an "isolated virtual environment" is defined and/or implemented; and,
 2. the concept of "activation".

#### What is an isolated virtual environment?

In terms of operation I'd expect a "virtual environment" to *replicate the same experience* as installing the application directly in the global non-virtual environment.

Additionally, I'd expect an "isolated" environment to not be affected by things outside the environment.

I think the key differences in our respective mental models hinge on how we each define what "the same experience" means:

[Okay, on re-reading this, the issue is not about what happens *in* a virtual environment (where we both seem to agree) but about what happens when a Python binary located within a virtual environment is executed directly, so, perhaps assume I wrote that instead. :) ]

 * Your perspective seems to be that--in a global installation--a subprocess Python would use the system Python, thus, in a virtual environment the subprocess Python should *also* use the system Python. 

 * My perspective is that in a global installation a subprocess Python would use the *same* Python as its parent and so in a virtual environment the subprocess Python should also use the *same* Python as its parent.

Thus both perspectives are consistent within themselves but result in different expected outcomes.

#### What is activation?

As I understand it, when an virtual environment is "activated":

 1. The correct Python library search paths for the virtual environment are set (via `sys.path`); and,
 2. The `PATH` (and other environment variables) are modified to ensure the correct Python binary (i.e. the one within the virtual environment) is used.


My initial understanding was that when the `/<virtualenv>/bin/python` binary was executed directly without manually activating the virtualenv that it would "auto-activate" the virtualenv. I now realise that what happens is actually "partial auto-activation" where only the first stage occurs:

 * Key to my confusion around this part is that I don't understand in what scenario(s) is having only a partially activated virtualenv more desirable?

#### My revised mental model

At this point I can make a slight adjustment to the explanation of my mental model of virtual environments and their activation:

 * My expectation is that directly running the Python binary which is located within the virtual environment would either not work at all or "auto-activate" the virtual environment because that means the result would be the same as manually activating the same environment.

But instead, with the current behavior there is a result that is different to *both* a global install and a manually activated virtual environment install because the partially activated virtual environment results in the subprocess using a different Python binary to its parent.

#### Why partial rather than full auto-activation?

If full auto-activation was never expected to be the outcome of running the virtualenv Python binary directly then I wonder why even partially activate it by setting the libraries? And, thus I return to the question of in what scenario is partial activation useful and safe when a full auto-activation isn't?

From your comments in #906 my understanding was that the reason for not implementing full auto-activation was a technical one and you had valid concerns around cross-platform compatibility issues around using wrapper scripts. Thus why I proposed a pure Python solution using standard mechanisms in #1081.

(Side note: I recognize the significance of maintaining backward compatibility; so, if that was desired in this case, the solution in #1081 could require a guard of some sort to ensure existing behavior didn't suddenly change--perhaps the guard could be that the `sitecustomize.py` file enabling full auto-activation is only added when specifically requested by a command line switch.)

### Consideration of the approaches you suggest

[Note: I've swapped the order of the next two quotes.]

#### Use `sys.executable`

> Typically when running a Python subprocess, you should use subprocess.call([sys.executable, ...]) to be sure you're using the same interpreter as your main executable, if that's what you want.

Yes, that's reasonable (and is specifically recommended in [PEP 394](https://www.python.org/dev/peps/pep-0394/)) and, although I didn't mention it, one of my earlier attempts to fix the issue was to follow that approach.

*However* in this case the approach of specifying the executable in the subprocess call isn't appropriate because an inventory script doesn't *have* to be a Python script, it can be any form of executable. (As can be seen in the code in question: <https://github.com/ansible/ansible/blob/2cdf31d3a2cc7d15c3362efb2e7f3566e8bdb811/lib/ansible/plugins/inventory/script.py#L169>. Note: that code is not yet released due to a recent reorganisation but the approach used is the same in the current released code.)

#### Don't use `env` shebang

> In my opinion, the Ansible scripts that use #!/usr/bin/env python are wrong, unless the Ansible documentation makes it clear that you need the correct Python interpreter on your PATH
 
There isn't necessarily a direct connection between Ansible and an inventory script (indeed, `ec2.py` was originally designed such that it could just be used directly to get a host list for test purposes) so, as per common usage/PEP-394, to me it doesn't seem unreasonable that a `env` based shebang line is used.

As it happens, the [`ansible` script](https://github.com/ansible/ansible/blob/devel/bin/ansible#L1) itself starts with a `env`-based shebang line but it would appear this gets [intentionally mangled](https://stackoverflow.com/questions/1530702/dont-touch-my-shebang) on installation. (Indeed, if it *didn't* get mangled then AFAICT the virtual environment would have to always be manually activated to work correctly anyway.)

If it were common to run `ansible` (or, indeed, any system installed script) with a Python interpreter that wasn't on the `PATH` I could see it potentially reasonable that Ansible make a note as you suggest. However, given the issue doesn't occur when Ansible is installed globally nor when it's run in a manually activated virtual environment but only in a "partially auto-activated" environment it doesn't seem to me that it's an issue on Ansible's side. (Particularly as the same issue would occur with any other Python-based tool that worked the same way.)

#### Don't use a non-activated virtualenv's interpreter

> (in which case, running via a non-activated virtualenv's interpreter is not following their documentation).

The use of the phrase "non-activated virtualenv's interpreter" suggests that it was never your intention that running such an interpreter would auto-activate the environment? Is there a particular reason why a "partial activation" (i.e. setting `sys.path`) fits your mental model of how virtualenv should work but a "full activation" (i.e. setting the Python interpreter process's environment variables) doesn't?

(The only concern that comes to mind is that somehow a changed `PATH` might interfere with correct operation of some other type of subprocess but if this were the case wouldn't doing the same thing in a manually activated virtual environment also interfere? I guess I'd see it as being more consistent if the virtualenv Python removed its own parent directory from the `PATH` after it was started if this was a concern.)

#### Do update the documentation

> Having said all this, I'd be happy with a comment in the documentation saying that some 3rd party scripts assume that Python is in PATH or that VIRTUAL_ENV is set - if you want to use such scripts without activating your virtualenv, you will need to ensure that you manually set the appropriate environment variables in your code.

Sure...or, they could drop the handy pre-written `sitecustomize.py` into the virtualenv's `site-packages`. :D

But, also, I appreciate you being open to a revised documentation patch. :)

### A summary and proposed next step

In summary, because of my mental model I don't understand why if a virtual environment *can* be auto-activated (in a non-hacky way) by the Python binary that is located within it, what scenario benefits from *not* doing this and only using partial activation?

However, I do understand that you are developing the project to fit the mental model that you have of it and which seems reasonable; so, you're obviously entitled to not implement full auto-activation if you don't see it as consistent with your mental model and having value for the project.

So, if that's still the case, then I can update the documentation patch and will instead suggest the `sitecustomize.py` approach to Homebrew to see if they'll incorporate it.

Thanks again for your time and work on the project!

pfmoore:
Wow, that's a comprehensive reply! I don't have the time right now to read it fully and reply - I'll try to do that later. However, one point I think you're missing is that when you say

> "I'd expect a "virtual environment" to replicate the same experience as installing the application directly in the global non-virtual environment."

you've missed an important point - a non-activated virtualenv *is not on `PATH`* and behaves exactly as the system Python would *if that weren't on `PATH` either*. It's just that it's a *lot* rarer for anyone to run the system Python without it being on `PATH`.

I'll read your full reply and respond in more detail later.

pfmoore:
OK, I've now read your reply more fully and I see why you're finding the current behaviour difficult to follow. Let me try to clarify.

> * Your perspective seems to be that--in a global installation--a subprocess Python would use the system Python, thus, in a virtual environment the subprocess Python should also use the system Python.
> * My perspective is that in a global installation a subprocess Python would use the same Python as its parent and so in a virtual environment the subprocess Python should also use the same Python as its parent.

This is the first point where I think your view is going astray. There's no such thing as "a subprocess Python". All there is, is "a subprocess". Python's subprocess module simply calls the OS to start a new process (`execv` on Unix, `CreateProcess` on Windows). Whether that subprocess runs Python, a shell, or anything else, is entirely down to the OS, and Python has no interest in, or control over, what happens.

When you're talking about a "Python subprocess", your examples are scripts that start with a shebang line `#!/usr/bin/env python`. Those are handled by the OS, which on Unix reacts to the first 2 bytes `#!` and reads the line, invoking the executable `/usr/bin/env` to run Python - which it does by searching the `PATH` environment variable for a python executable (because that's what the `env` program does). The one it finds is the system Python, because you've done nothing to tell it to do anything different. (The situation on Windows is somewhat different, but I'll ignore that here). I don't know if I'm explaining things you already know - if so, I apologise - but the net effect is that the Python program that invokes the subprocess has *no control* over what the OS does, except by the normal means of setting environment variables like `PATH`. (The Python program calling `subprocess` could modify the `PATH` environment variable if it chose to, but the Python *language* certainly shouldn't do so without being asked).

> As I understand it, when an virtual environment is "activated":
> 1. The correct Python library search paths for the virtual environment are set (via sys.path); and,
> 2. The PATH (and other environment variables) are modified to ensure the correct Python binary (i.e. the one within the virtual environment) is used.

Not correct.

Specifically activation doesn't do (1) at all. The standard Python executable sets `sys.path` and it does so based on various things (the location of the executable). The `site.py` file shipped with virtualenv does further things to modify this. But *none* of this is dependent on activation - it happens for all virtualenvs.

What activation does is (1) add the `bin` directory of the virtualenv to the front of `$PATH`, and (2) set the `VIRTUAL_ENV` environment variable. Apart from some cosmetics (setting the prompt) that's it. And of those, the *only* change that has any global effect is the setting of `PATH` (i.e., unless a program explicitly checks if `VIRTUAL_ENV` is set, it won't care).

So, the specific effect of all this is that the *only* practical difference activating a virtual environment has is in how `PATH` is set.

Whether your scripts are written to assume that `python` is a command available on `PATH`, and it points to the "correct" Python, is a matter for those scripts to get right. It sounds like the Ansible scripts you're referring to do make that assumption, but the Homebrew approach of "using a virtual environment" is somehow violating that assumption if you don't activate the virtualenv. That's an issue that will need to be resolved between those two parties - it's not something that virtualenv can (or should, in my opinion) "fix". The resolution may be as simple as saying "running the Homebrew scripts without activating the virtualenv is not supported" or it may be something complex like patching the scripts - I can't say.

In my view, the real confusion here comes from the term "activating". It *sounds* like it's a really clever bit of machinery, but it isn't - it's just adding the virtualenv to `PATH`. You don't install git and then "activate" it, you simply add the directory you pit the git executable in to your `PATH` (or more likely, git installs itself to somewhere that's by convention already on your `PATH`, but you know what I mean...)

I hope this helps clarify.

follower:
Thanks for reading my comments and taking the time to give a detailed reply, I appreciate you doing so. :) 

I'll preface my reply by saying that while I was writing up an issue related to this for Homebrew I did gain some clarity with regard to the more common non-Ansible use case.

In the interests of enabling you to prioritize your time I'll now skip to my conclusion and you can read my "working" & other thoughts after that, if it is still of interest...

### In conclusion

My suggestion for where to go from here:

 1. Definitely a documentation change. (Which I now feel I'm in an improved position to write due to my clearer understanding of how the different pieces fit together.)
 2. I would still like to advocate for, say, a command-line option on creation of a virtualenv that would enable "auto-activation" of the virtualenv via `sitecustomize.py` (when the virtualenv's Python interpreter binary is executed directly) for those who are looking for a wrapper script and/or using an "application" that is implemented like Ansible.
 3. If (2) is not an option, then I'd like the documentation to at least point to a suitable `sitecustomize.py` based on the approach shown in #1081. (Actually, on reflection, I guess it would be more prudent to not implement (2) until such a `sitecustomize.py` has seen some extended use--and it seems like Homebrew-installed Ansible, at a minimum, needs it, so that could be a starting point. But at some point it would be nice if the script could at least be distributed as part of `virtualenv`.)

And, if nothing else, hopefully our conversation will be helpful to others if someone else starts down this same road as confused as I was. :)

Thanks again for your time.

----

[By now I have no idea about the most logical flow of the rest of this, so you might want to skip to the "Clarification in regard to your initial reply" and later sections and then return here.]

### My improved understanding around the non-Ansible use case

As prefaced above, I can now see how in a "typical" "application" installation scenario, one wouldn't want the fact that a particular application is written in Python to affect the execution of an arbitrary script by that application that also happened to be written in Python.

Say I have an installed tool `every-ten-minutes` (that happens to be written in Python and has a `virtualenv`-based shebang line) which takes an executable path and runs that executable every ten minutes. Then, let's say, I happen to write a Python script called `print-hello` (with an `env`-based shebang line) that prints "hello" to the console and I want to supply that as the executable to `every-ten-minutes` (i.e. `every-ten-minutes print-hello`). I see the logic in not setting the `PATH` for `print-hello` in this situation because clearly it may depend on default/system Python libraries not available within the `every-ten-minutes` virtualenv.

And, it would seem that this is the most typical use case for a "from the outside" use of a virtualenv.


#### Is "lack of consistent behavior" a bug?

*But* if I try to execute `every-ten-minutes print-hello` from *within* an activated virtualenv there's no guarantee I'll get the same result because the child process will inherit the modified `PATH`.

 * Is this behavior intentional?
 * Is the difference in behavior intentional?
 * Is it actually a bug that should be fixed by ensuring the child process receives the *original unmodified* `PATH` instead? That way both forms of virtualenv use would behave in the same way.

Note: I'm not *actually* advocating changing to use the original `PATH` for child processes of an activated virtualenv Python. If people haven't been clamoring for this to be changed I assume it's not a big deal.

*But* I think it does show that it's not entirely unreasonable to anticipate that the same behavior might occur with a "from the outside use" as well.

It does all make me wonder whether people generally just don't use the same environment alongside Python scripts from both within and without. (Although requests for wrapper scripts suggest that it *does* happen sometimes.)


#### The virtualenv + Ansible + inventory scripts (+ Homebrew) edge case

So, it would appear that virtualenv + Ansible + inventory scripts (+ Homebrew) now just happens to be a really gnarly edge-case that's making multiple abstractions leak or something. :) (In Homebrew terms it seems a little like an "application" which *also* supplies its own "bindings".)

Somewhat amusingly, [Homebrew documentation](https://docs.brew.sh/Python-for-Formula-Authors.html) specifically uses Ansible as an example when it says:

> Homebrew draws a distinction between Python **applications** and Python **libraries**. The difference is that users generally do not care that applications are written in Python; *it is unusual that a user would expect to be able to `import foo` after installing an application*. Examples of applications are `ansible` and `jrnl`.

:D *sigh*

However it seems this approach is intended to stick around, according to <https://github.com/ansible/ansible/issues/17147#issuecomment-241029301>: "We plan to make more use of the ansible libraries in the dynamic inventory scripts over time" 


### Clarification in regard to your initial reply 

> > "I'd expect a "virtual environment" to replicate the same experience as installing the application directly in the global non-virtual environment."
> 
> you've missed an important point - a non-activated virtualenv is not on PATH and behaves exactly as the system Python would if that weren't on PATH either. It's just that it's a lot rarer for anyone to run the system Python without it being on PATH.
> 

Apologies, I believe I did actually understand your point originally, and had intended *that* sentence to be read unaltered as "I'd expect [an activated] 'virtual environment' to replicate the same experience as installing the application directly in the global non-virtual environment."

But I think my "...on re-reading this, the issue is not about what happens in a virtual environment..." remark may have confused things a little. The "this" I was referring to applied to the bullet points following the statement, not the paragraphs proceeding it--so my remark was intended to be read as "...on re-reading [the following bullet points], the issue is not about what happens in a virtual environment...". (Writing non-linearly is hard...)


### Defining terms used

> > * Your perspective seems to be that [...snip...].
> > * My perspective is that in a global installation a subprocess Python would use the same Python as its parent and so in a virtual environment the subprocess Python should also use the same Python as its parent.
> > 
> 
> This is the first point where I think your view is going astray. There's no such thing as "a subprocess Python".

Alas, my use of the phrase "a subprocess Python" was *intended* to be short-hand to refer to "the Python interpreter binary which was first found on the `PATH` by the `env` tool and executed as the newly created child process created as a result of the call to a function of the `subprocess` module" but apparently I need to define my terms more clearly. :)

So I believe my understanding of this bit does match yours.

> I don't know if I'm explaining things you already know - if so, I apologise

Graciously accepted but no need for an apology, it made sense to clarify that I understood--after all, you had no way of knowing that I hadn't included *every* background detail I knew, as much as it might've seemed the case by volume of text. :D

> The Python program calling subprocess could modify the PATH environment variable if it chose to, but the Python language certainly shouldn't do so without being asked

It's possibly a moot point but I'd see value in drawing a distinction between "the Python language" (by which I assume you mean the Python interpreter binary) and "the Python site configuration" in this case. I'd agree the language wouldn't be expected to have modified the `PATH` but in a special circumstance I think it wouldn't be unreasonable that it might be possible for "the Python site configuration" to have modified the `PATH`. (Of course, given #1081 I'm obviously biased in this regard. :) ) 


#### Background (I): The distinction between "use" and "activate" 

It seems to me that some of the confusion around talking about virtualenv is due to the project drawing a distinction between "using a virtualenv" versus "activating and then using a virtualenv". Coming from outside the project and being unaware of virtualenv's implementation details, the existence of and need for this distinction is not obvious.

(Side note: I'm quoting out of context but this quote I found earlier while closing some tabs seems to intersect with this aspect: "Using the virtualenv's python interpreter is how you use a virtualenv." via <https://github.com/pypa/virtualenv/issues/124#issuecomment-1040000>)


#### Background (II): "Intention" versus "Implementation"

I think another factor leading to confusion is lack of shared understanding around "intention" versus "implementation detail". (Which also relates to the lack of awareness about "use" vs "activation".)

For example, I had assumed that because the `PATH` value is modified when one "uses" (an activated virtualenv) that this `PATH` modification was considered an "intentional" "part" of the virtualenv but it seems to turn out this is just an "implementation detail" related to how the correct Python binary is found.

Another example is that I assumed that the split between modifications made in the activate script and modifications made in `site.py` was an "implementation detail" but as I understand it is "intentional" that only the modifications made in `site.py` are considered part of virtualenv "use".

Additionally, I assumed that the need to manually "activate" a virtualenv was an "implementation detail" related to there being no easy way to perform the task automatically when in a shell, rather than an "intentional" distinction between "activation" and "use".


#### Background (III): Defining the "Encapsulation" and "Isolation" use cases

Broadly, my impression is that the two primary use cases for virtualenv are:

 * "Passively" using a virtualenv "from the outside"  (i.e. not activated) for purposes of encapsulation.

 * "Actively" using a virtualenv "from the inside" (i.e. activated) for purposes of isolation (from "sibling" installs, shall we say).

It seems where a use case "crosses the boundaries" of these two categories is where issues occur, e.g.:

 * The Ansible inventory script use case. (Needs "auto activation".)
 * The cronjob/web-server use case. (Needs a "wrapper" script.) 


### Mental models revisited

> > As I understand it, when an virtual environment is "activated":
> > 
> > - The correct Python library search paths for the virtual environment are set (via sys.path); and,
> > - The PATH (and other environment variables) are modified to ensure the correct Python binary (i.e. the one within the virtual environment) is used.
> > 
> 
> Not correct.
> 
> Specifically activation doesn't do (1) at all. The standard Python executable sets sys.path and it does so based on various things (the location of the executable). The site.py file shipped with virtualenv does further things to modify this. But none of this is dependent on activation - it happens for all virtualenvs.
> 
> What activation does is (1) add the bin directory of the virtualenv to the front of $PATH, and (2) [...snip...]
> 
> So, the specific effect of all this is that the only practical difference activating a virtual environment has is in how PATH is set.
> 

Essentially the fundamental mismatch between our mental models of virtualenv seems to come down to the impact of the "use" vs "activation" and "intention" vs "implementation" factors as evidenced in the above quoted section.

So, it's definitely helpful to see it spelled out like that.

[Quote moved from here to next section]

> In my view, the real confusion here comes from the term "activating".

I would agree it's a significant contributing factor. (per "Background (I)")

> It sounds like it's a really clever bit of machinery, but it isn't - it's just adding the virtualenv to PATH.

While the specific implementation details might be straightforward, I feel like the task it is performing ("ensure the correct virtualenv Python binary is selected") is actually a significant part of virtualenv's entire functionality.


### Resolution for the Ansible + virtualenv (+ Homebrew) use case

> Whether your scripts are written to assume that python is a command available on PATH, and it points to the "correct" Python, is a matter for those scripts to get right. It sounds like the Ansible scripts you're referring to do make that assumption, but the Homebrew approach of "using a virtual environment" is somehow violating that assumption if you don't activate the virtualenv.
>
> That's an issue that will need to be resolved between those two parties - it's not something that virtualenv can (or should, in my opinion) "fix".

While Homebrew can definitely be *a* location for a "fix", I do feel there is a wider, more general Ansible + virtualenv issue--after all, installation via Homebrew isn't the only scenario in which Ansible ends up in a virtualenv.

And, Ansible *does* successfully work in either a global installation or a manually activated virtualenv, so I suspect they might have minimal motivation to make any changes. (Although maybe "installing" inventory scripts or shipping the `contrib` files are options.)

Given that there are other actual scenarios (activation via wrapper script) and potential scenarios (an application that acts in the same manner as Ansible) that are/can be also affected I feel like the common element is virtualenv. Which suggests it might be a good place for a "fix" (or, at least, to store a "fix" :) ).

Perhaps, even if the need for a "fix" is based on a fundamental misunderstanding of how virtualenv is intended to be used and is implemented. :D  

> The resolution may be as simple as saying "running the Homebrew scripts without activating the virtualenv is not supported" or it may be something complex like patching the scripts - I can't say.

I'm certainly hoping it's as simple as dropping a `sitecustomize.py` into `site-packages` on installation when needed but I accept that might not be desirable as far as the virtualenv project is concerned. :)

Once again, thanks for your time, reading this far and your work on the project!

pfmoore:
Well, I think we still disagree somewhat. With regard to your proposal:

1. I'm OK if you want to propose a documentation change. However, it needs to be worded in a way that matches the common understanding of how virtualenv works. Your view seems pretty unusual, at least in the sense that I've never heard anyone explain virtualenv the way you do. So I wouldn't want to confuse other users by introducing terms (like "auto-activation") that only you have ever used.
2. Sorry, but I'm against any sort of "auto-activation" flag. Reasons:
  a. There's not really any such thing as "auto-activation", so it doesn't much make sense as a concept.
  b. The `sitecustomize.py` file is *solely* for the user to manage, so virtualenv shouldn't be involved with it at all.
  c. Silently modifying the environment what starting Python seems like a pretty bad idea to me, so even if it works around the issues with ansible in Homebrew, I'd consider it a *bad* workaround. I don't have a better suggestion, as I know nothing much about either system, but I'd strongly suggest finding something better.
3. I don't think we should be documenting `sitecustomize.py` hacks like you're suggesting. Anyone with enough understanding to know the risks and issues with the approach you suggest is presumably going to know enough to be able to work out the approach themselves. And if they *don't* know the risks involved, I don't wan the virtualenv docs pointing them at something that could potentially get them into trouble.

I'm sorry if this isn't the response you were hoping for, but I'm afraid I really do feel that the issue here is with how Ansible/Homebrew are interacting, and not with virtualenv. Note by the way that the "problems" you are seeing would also appear if the core Python `venv` module was used in place of virtualenv - which to me clearly points to it not being a problem with virtual environments, but rather with users' (in this case Ansible's and Homebrew's) understanding of the implications of them.

ssbarnea:
@pfmoore It took me a lot of time to read this thread. TBH, I encountered this issue before myself. I kinda understand why the user raised the bug but I also understand your point of view: effectively virtualenv have no idea (or control) over what script it will be installed inside its bin directory.

Sadly it remains up to each script to implement the logic for being able to replicate activation tricks whenever this is needed, ansible being a very good example. Shortly I would advise against using any virtualenv installed tools without doing an activate first.

Other than this, looking at the proposed documentation change, I have nothing against it, it just describes more clear the behaviour, hopefully having less confusions in the future.

PR #1082, opened 2017-09-16
Update activate.fish

ofek:
https://github.com/ofek/hatch/issues/21


ssbarnea:
@pfmoore seems like reasonable to approve.

pfmoore:
The embedded copy of the script needs refreshing. Just changing the raw script isn't sufficient - see `bin/rebuild-script.py`.

ofek:
Sorry, but it's not immediately clear to me what needs to be done there. What should be changed in that file?

ssbarnea:
@ofek Goop point. I looked yesterday at the file and didnt see anything to do about it. Still, I am now trying to run the full test suite on both MacOS and Linux in order to spot any problems. I will add tests for our bug because is very easy to do so, just enabling strict mode and calling activate in isolated mode (env -i) is a very good way to replicate it.

I would go so far to even add a linter like bashate for the bash scripts but this would add an extra test dependency. I will do it only if I get feedback that this is desired, otherwise i will just use it myself.

Issue #1050, opened 2017-05-31
Fails to copy python3.dll to Virtualenv under Windows


jdmansour:
When creating a Virtualenv under Windows (using Python 3.6), it seems to work fine at first, but some modules (notably PyQt5) complain with `ImportError: DLL load failed: The specified module could not be found.`. The cause is that it fails to find `python3.dll` (not `python3.6.dll`). Manually copying it to the Scripts directory fixes the problem.

Another option would be to add the original Python directory to the PATH, but I think that might cause problems. Better to keep the Virtualenv separate from the main installation, at the expense of copying one more file.

I wonder how this is handled under Linux, I never had that problem there?

Here somebody mentions the problem in a different isue: https://github.com/pypa/virtualenv/issues/796#issuecomment-291014201
I think this issue is related, but for Linux: https://github.com/pypa/virtualenv/issues/1015

gst:
effectively this really looks like 1015 but on Windows.

> I wonder how this is handled under Linux, I never had that problem there?

simply Linux system pythons (python coming with the Linux distributions) aren't compiled with that option (relative path to the shared library).
you have to compile one by hand and provide the option.

gst:
@jdmansour fyi: python3  has its embedded `venv`  module.. which is normally recommended over virtualenv

jdmansour:
@gst Interesting, I always thought that the builtin `venv` was identical to `virtualenv`. I just tried it out, and `python -m venv myvenv` does copy `python3.dll`. So I guess it would be good if `virtualenv` did the same for consistency (if you are still maintaining it, given there is `venv` in Python 3)?

One added complication is that I used `virtualenvwrapper` (or `virtualenvwrapper-win`) which seems to insist on using `virtualenv` the package.

gst:
I m a user like you here ;) venv vs virtualenv: they are not the same yes.
If you are using python3 just use venv and all should be good I guess (had not the chance to use it/python3 recently).

brupelo:
@dstufft Hi Donald, sorry to ping you but I've seen in the commit history you're a major contributor of virtualenv. Any possible solution about trivial issue?

Running virtualenv 15.1.0 on a windows platform will show this issue after installing pyqt5. Copying 
 manually the missing python3.dll to the Script folder will solve it.

So, wouldn't something like copying python3*.dll on the virtualenv Scripts folder if platform==windows solve it?

PR #1083, opened 2017-09-20
Fix issue #1050 - Copy additional python dll to the scripts folder so…

brupelo:
… libraries like pyqt5 can import succesfully

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1241,24 +1241,33 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
             elif os.path.exists(python_d_dest):
                 logger.info('Removed python_d.exe as it is no longer at the source')
                 os.unlink(python_d_dest)
+
             # we need to copy the DLL to enforce that windows will load the correct one.
             # may not exist if we are cygwin.
-            py_executable_dll = 'python%s%s.dll' % (
-                sys.version_info[0], sys.version_info[1])
-            py_executable_dll_d = 'python%s%s_d.dll' % (
-                sys.version_info[0], sys.version_info[1])
-            pythondll = os.path.join(os.path.dirname(sys.executable), py_executable_dll)
-            pythondll_d = os.path.join(os.path.dirname(sys.executable), py_executable_dll_d)
-            pythondll_d_dest = os.path.join(os.path.dirname(py_executable), py_executable_dll_d)
-            if os.path.exists(pythondll):
-                logger.info('Also created %s' % py_executable_dll)
-                shutil.copyfile(pythondll, os.path.join(os.path.dirname(py_executable), py_executable_dll))
-            if os.path.exists(pythondll_d):
-                logger.info('Also created %s' % py_executable_dll_d)
-                shutil.copyfile(pythondll_d, pythondll_d_dest)
-            elif os.path.exists(pythondll_d_dest):
-                logger.info('Removed %s as the source does not exist' % pythondll_d_dest)
-                os.unlink(pythondll_d_dest)
+            py_executable_dlls = [
+                (
+                    'python%s.dll' % (sys.version_info[0]),
+                    'python%s_d.dll' % (sys.version_info[0])
+                ),
+                (
+                    'python%s%s.dll' % (sys.version_info[0], sys.version_info[1]),
+                    'python%s%s_d.dll' % (sys.version_info[0], sys.version_info[1])
+                )
+            ]
+
+            for py_executable_dll, py_executable_dll_d in py_executable_dlls:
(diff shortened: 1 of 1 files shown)


pfmoore:
LGTM. Thanks for the contribution!

brupelo:
@pfmoore Thanks to you to take care of the merging ;)

May I ask you when it'll be possible to have available the new wheel on pypi? Who's the one/s in charge of that decission? Asking you this cos I'm developing a ST plugin to manage virtualenv on windows but I won't be able to release it till the new wheel is cooked. I mean, it's not nice to host the wheel by myself and saying the users "hey, just do pip install -U http://myweirdhost/virtualenv-15.2.0.dev0-py2.py3-none-any.whl" ;)

Anyway, thanks guys for this really nice project, keep it going!

pfmoore:
It'll be in the next release of virtualenv, which will probably be released at the same time as the next release of pip - but there's no specific timescale for that. Sorry I can't be more definite.

brupelo:
I understand, looking at the pip [changelog](https://pip.pypa.io/en/stable/news/) and virtualenv [changelog](https://virtualenv.pypa.io/en/stable/changes/) i can see the last release dates are in 9.0.1 (2016-11-06) and 15.1.0 (2016-11-15) respectively, almost one year since, that's quite a long time!

That proves both projects are stable and mature enough... on the other hand, IMHO it's always a nice thing projects have a more short release lifecycle, so they get contributors more passionate about it and a bigger driving force ;)

pfmoore:
We're aware of that principle, but we're a very small group with limited time, and pip is currently working through some quite significant architectural changes. We could in theory make a release of virtualenv independently of pip, but (although I'm not the guy who does the releases) I believe that is a bit tricky to manage. And while I understand it's blocking you, I don't know that this issue is sufficiently critical (#1050 was the only report of the problem that we've ever seen, as far as I am aware) to warrant an emergency release.

brupelo:
@pfmoore Paul, ok, fair enough, I totally understand your point. I'll be waiting patiently then till virtualenv is released on pypi. 

And yeah, it's as you're saying, the issue is definitely not critical, it's just annoying. For instance, let's say you're an user who'll use one or two global venvs to code their pyqt5 apps/widgets, when he realizes about the bug he'll just spend few minutes time googling a bit or asking about the issue on some forums/irc channel and then he'll just copy the necessary dlls to load pyqt5 (or similar libs) succesfully. On the other hand, if your an user who's using a proper plugin in his favourite IDE to manage virtualenvs and is creating very often new virtualenvs+requirements per project/release the issue becomes annoying.

Anyway, moving on. Thanks for your time and explanations ;)

pfmoore:
I suspect the issue may be that pyqt5 has started linking to `python3.dll` rather than `python36.dll`. I'm not sure why they chose to do that, but it may be that's what's made this issue start to come up more often. I'm not saying it's a bad thing that they do that, but historically the stable ABI hasn't been used much, which is likely why tools are having to catch up (I suspect pyqt5 have had to customise their build process to do this, AFAIK setuptools doesn't support using the stable ABI out of the box yet).

See https://github.com/pypa/pip/issues/4445 for another aspect of this issue, which I found via a Google search (and explains why the fact that it's pyqt5 seemed familiar to me :smile:)

brupelo:
Interesting. Btw, also worth mentioning the right way to test virtualenv would be assuming PATH doesn't contain any system-wide python folder, otherwise these types of bugs will be missed silently.

For instance, if i would have python3.6.2 in my PATH python3.dll would be reachable in the first place and pyqt5 would have been loaded correctly (ie: where python3.dll), i think is fair to assume a virtualenv won't rely on PATH containing any python paths.

Also, as you can see, the fix provides a manual list of python dlls instead of something more generic like copying the glob of `*.dll` or `python*.dll`, that way you can have a finer control over the copying process and being minimalistic :)

pfmoore:
Ha. The virtualenv test suite is woefully limited (we don't even *have* CI on Windows). That's an acknowledged issue, but no-one has ever had the time to address it. It's one of the major reasons it's so hard to get fixes in - the huge diversity of systems that virtualenv is used on, coupled with the fact that virtualenv is basically a collection of hacks around quirks in those systems, and the fact that we have no tests to verify that we haven't broken stuff, leaves every PR at risk of breaking something that someone relies on. That's why (for example) I won't touch PRs for virtualenv on Unix, as I don't use Unix myself.

https://xkcd.com/1172/

ssbarnea:
@pfmoore I think that virtualenv has CI for both Windows and Linux. The Windows file `appveyor.yml` was added like 2 years ago, and it does run.

Regarding py36 support there is an open PR for enabling it on both Windows and Linux, and surprise, it works.  See https://github.com/pypa/virtualenv/pull/1089

PR #1086, opened 2017-10-03
Link to Python 3 venv

galund:
The lack of a pointer here to `venv` is a source of confusion to people coming fresh to Python, who will often see this documentation quoted or referenced from existing material.

I think this would be an important enough change to warrant updating the stable release.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,7 +11,9 @@ Dev IRC: #pypa-dev
 Introduction
 ------------
 
-``virtualenv`` is a tool to create isolated Python environments.
+``virtualenv`` is a tool to create isolated Python environments. Since
+Python 3.3 it has been replaced by the `venv
+module <https://docs.python.org/3/library/venv.html>`_.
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -103,6 +105,9 @@ Compare & Contrast with Alternatives
 
 There are several alternatives that create isolated environments:
 
+* Python 3's `venv module <https://docs.python.org/3/library/venv.html>`_
```
gaborbernat: I would like to reformulate this a bit. Virtualenv is still the de-facto tool to use if you want cross python virtual environment creation; custom script activations (powershell, xonosh) or support for pypy, jython and co. It also has a few features desire-able in some situations.

asottile:
Another thing to add here, some OS distributions choose not to provide the `venv` module by default so often `virtualenv` is a more portable choice.

(notably debian requires you to install `python3-venv` in order to use `-mvenv`)

PR #1087, opened 2017-10-03
travis: enable use of test_activate.sh

ssbarnea:
This will also run test_activate.sh which is an essential part of the testing, obviously outside tox because inside would not work.

### DIFF
--- a/tests/test_activate.sh
+++ b/tests/test_activate.sh
@@ -10,7 +10,16 @@ rm -rf ${TESTENV}
 
 echo "$0: Creating virtualenv ${TESTENV}..." 1>&2
 
-${VIRTUALENV} ${TESTENV} | tee ${ROOT}/tests/test_activate_output.actual
+PY_BLEND=$(python -c "import sys; print('pypy' if '__pypy__' in sys.builtin_module_names else 'python');")
+cat >${ROOT}/tests/test_activate_output.expected <<EOL
+New ${PY_BLEND} executable in /tmp/test_virtualenv_activate.venv/bin/python
+Installing setuptools, pip, wheel...done.
+EOL
+
+# "Path not in prefix" warning is not really an error, being expected under pypy
+# > Path not in prefix '/home/travis/virtualenv/pypy2.7-5.8.0/include' '/opt/python/pypy2.7-5.8.0'
+
+${VIRTUALENV} ${TESTENV} 2>&1 | sed '/^++\|^Also creating executable in\|^Using real prefix\|^Path not in prefix/d' | tee ${ROOT}/tests/test_activate_output.actual
 if ! diff ${ROOT}/tests/test_activate_output.expected ${ROOT}/tests/test_activate_output.actual; then
     echo "$0: Failed to get expected output from ${VIRTUALENV}!" 1>&2
     exit 1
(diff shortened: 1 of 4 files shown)


ssbarnea:
@pfmoore I hope you will not mind reviewing this PR which enables testing /test/test_activate.sh script which... surprise was not used as part of the gating :)

This would allow us to improve the gating and avoid potential surprises from other PRs.

ssbarnea:
Please note that this patch needs https://github.com/pypa/virtualenv/pull/1089 merged first in order to pass the testing gates.

ssbarnea:
@jakub-bochenski Can you please review this? That's is now No1 priority because enables the validation that the encoded script is exactly the same thing as the unencoded one. 

We really need to be sure that our gates are validating the that, as humans will probably not look at these parts.

jakub-bochenski:
@ssbarnea I'd already but I don't feel too confident in this area; I can give it a shot but I'd prefer somebody more familiar with this code reviewed it

tests/test_activate.sh
```diff
@@ -10,7 +10,16 @@ rm -rf ${TESTENV}
...
-${VIRTUALENV} ${TESTENV} | tee ${ROOT}/tests/test_activate_output.actual
+PY_BLEND=$(python -c "import sys; print('pypy' if '__pypy__' in sys.builtin_module_names else 'python');")
+cat >${ROOT}/tests/test_activate_output.expected <<EOL
+New ${PY_BLEND} executable in /tmp/test_virtualenv_activate.venv/bin/python
+Installing setuptools, pip, wheel...done.
+EOL
+
+# "Path not in prefix" warning is not really an error, being expected under pypy
+# > Path not in prefix '/home/travis/virtualenv/pypy2.7-5.8.0/include' '/opt/python/pypy2.7-5.8.0'
+
```
krosinski: How about checking venv creation exit code?

krosinski:
@ssbarnea the changes look good for me. Validation virtualenv creation standard output may impose a lot of maintenance effort in the future - if it's possible to relay only on exit code I'd got for it.

tests/test_activate.sh
ssbarnea: @krosinski I do agree with you that checking the stdout/stderr output makes no sense, that's not really an API. 

I would check for behaviour which means: exit code PLUS validating that the environment was activated, which to me it means checking the `which python` returns the venv version of python.

Overall test_activate.sh is a mess and last week I started to rewrite it as pytest unittests and also to test it for all 4 supported shells:(bash, csh, fish and powershell. You can see the WIP work at https://github.com/pypa/virtualenv/pull/1090 but it will take some time to get it ready as increasing the number of shells multiplied by number of supported platforms and python versions does uncover other bugs.

I think it will be better if we first mere a gradual improvement in test_activate.sh because otherwise we will block testing other fixes. I would prefer to avoid the case where a PR takes so much time that it never reaches the merge status.

If you are OK, I would not mind removing the output testing.

ssbarnea:
@dstufft do you think you could do cast your magic here? Thanks.

ssbarnea:
@asottile can this be merged, please?

asottile:
> @asottile can this be merged, please?

unfortunately I'm not part of pypa

ssbarnea:
Lets give it another try, maybe @dstufft @Ivoz @pjenvey or @mattip can have alook at it.

tests/test_activate.sh
```diff
@@ -10,7 +10,16 @@ rm -rf ${TESTENV}
 
 echo "$0: Creating virtualenv ${TESTENV}..." 1>&2
 
-${VIRTUALENV} ${TESTENV} | tee ${ROOT}/tests/test_activate_output.actual
+PY_BLEND=$(python -c "import sys; print('pypy' if '__pypy__' in sys.builtin_module_names else 'python');")
+cat >${ROOT}/tests/test_activate_output.expected <<EOL
```
mattip: You could use [platform.python_implementation()](https://docs.python.org/2/library/platform.html#platform.python_implementation) instead of probing `sys.builtin_module_names`

tests/test_activate.sh
```diff
@@ -63,8 +72,13 @@ echo "$0: Executing a simple Python program..." 1>&2
 TESTENV=${TESTENV} python <<__END__
 import os, sys
 
-expected_site_packages = os.path.join(os.environ['TESTENV'], 'lib','python%s' % sys.version[:3], 'site-packages')
-site_packages = os.path.join(os.environ['VIRTUAL_ENV'], 'lib', 'python%s' % sys.version[:3], 'site-packages')
+if '__pypy__' in sys.builtin_module_names:
```
mattip: better as `'PyPy' == platform.python_imlementation()`

mattip:
took a look, had a little nit comment, but I too am not a committer here

ssbarnea:
@gaborbernat Can you please merge this long standing improvement on virtualenv CI testing?

This also improves security as it will fail to run jobs where the embedded compiled code differs from the non-embedded one.

PR #1089, opened 2017-10-04
travis: remove py33 and added py36

ssbarnea:
py33 reached EOL and the Travis version is broken
as it missing zlib encoder, making impossible
to test bin/build_script.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,17 @@ language: python
 
 matrix:
   include:
+    # Based on https://docs.python.org/devguide/#status-of-python-branches
     - python: 2.6
       env: TOXENV=python2.6
     - python: 2.7
       env: TOXENV=python2.7
-    - python: 3.3
-      env: TOXENV=python3.3
     - python: 3.4
       env: TOXENV=python3.4
     - python: 3.5
       env: TOXENV=python3.5
+    - python: 3.6
+      env: TOXENV=python3.6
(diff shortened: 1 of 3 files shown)


ssbarnea:
@dstufft and @jakub-bochenski - Thank you! Remind me to buy a beer when we meet.

As expected,  now the https://github.com/pypa/virtualenv/pull/1087 is passing.

agronholm:
Python 2.6 has reached EOL 4 years ago. Why was that still left in?

ssbarnea:
@agronholm You are right but I would avoid removing it unless we found something broken with it. Usually you drop support when you cannot support it anymore not for the sake of it.

As currently the few tests that we have are passing on py26, I see no reason to drop. If they will find a bug that we can avoid inside virtualenv we will keep it. Once we discover that there is no way to keep supporting py26, we will drop it.

py33 on Travis was such an example, as it missing the zlib library and nobody is going to ever add it to py33. So is was more of being forced to drop support for it than just doing it.

dstufft:
Python 2.6 isn't supported in pip anymore, so when I pull in the next version of pip into virtualenv I'll end up dropping py26 anyways.

ssbarnea:
@dstufft I doubt many will cry when you will do this. I woudnt for sure, I was dreaming about getting pip >= 9 wide spread. The doctor feauture is very useful.

PR #1094, opened 2017-10-20
Adapt for changes in next pip release

### LINKED ISSUE #1093: pip 10 will no longer expose pip.main
In pip 10, the `pip.main` function is no longer available, as all internal APIs have been moved to an `_internal` namespace. Virtualenv uses `pip.main` at https://github.com/pypa/virtualenv/blob/master/virtualenv.py#L855 so this needs changing.

pradyunsg:
Fixes #1093 

This is really a special case hack. The only reason that I'm okay with importing from pip's internals here is because we're using pip to install pip and this is really the cleanest way to allow for the bootstrapping to work. Further, the script doesn't break any of pip's internal assumptions (like sole control of logging) so this is fine.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -858,7 +858,10 @@ def install_wheel(project_names, py_executable, search_dirs=None,
         import tempfile
         import os
 
-        import pip
+        try:
+            from pip._internal import main as _main
+        except ImportError:
+            from pip import main as _main
 
         cert_data = pkgutil.get_data("pip._vendor.requests", "cacert.pem")
         if cert_data is not None:
@@ -874,7 +877,7 @@ def install_wheel(project_names, py_executable, search_dirs=None,
                 args += ["--cert", cert_file.name]
             args += sys.argv[1:]
 
-            sys.exit(pip.main(args))
+            sys.exit(_main(args))
         finally:
             if cert_file is not None:
                 os.remove(cert_file.name)
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -860,10 +860,11 @@ def install_wheel(project_names, py_executable, search_dirs=None,
 
         try:
             from pip._internal import main as _main
+            cert_data = pkgutil.get_data("pip._vendor.certifi", "cacert.pem")
         except ImportError:
             from pip import main as _main
+            cert_data = pkgutil.get_data("pip._vendor.requests", "cacert.pem")
 
-        cert_data = pkgutil.get_data("pip._vendor.requests", "cacert.pem")
         if cert_data is not None:
             cert_file = tempfile.NamedTemporaryFile(delete=False)
             cert_file.write(cert_data)


pradyunsg:
FWIW, I tested this locally:

```
echo "Setup"
hg clone https://bitbucket.org/pypa/distlib
git clone https://github.com/pypa/pip
git clone https://github.com/pypa/virtualenv
pushd pip
rm -rf pip/_vendor/distlib
cp -r ../distlib/distlib pip/_vendor/
python setup.py bdist_wheel
popd
pushd virtualenv
rm virtualenv_support/pip-9.0.1-py2.py3-none-any.whl
cp ../pip/dist/pip-10.0.0.dev0-py2.py3-none-any.whl virtualenv_support

echo "Testing"
python ./virtualenv.py 'v e n v' --no-download
source ./v\ e\ n\ v/bin/activate
pip --version
```

If the last command gives 10.0.0.dev0 means all is well. :)

pfmoore:
LGTM

pradyunsg:
Doesn't seem to be ready.

```
Using base prefix '/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6'
New python executable in /Users/pradyunsg/.venvwrap/venvs/tmp-f40cf84096a4d12/virtualenv/v e n v/bin/python3.6
Also creating executable in /Users/pradyunsg/.venvwrap/venvs/tmp-f40cf84096a4d12/virtualenv/v e n v/bin/python
  Complete output from command "/Users/pradyunsg/.ve... e n v/bin/python3.6" - setuptools pip wheel:
  File "<stdin>", line 12, in <module>
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/pkgutil.py", line 634, in get_data
    return loader.get_data(resource_name)
OSError: [Errno 0] Error: 'pip/_vendor/requests/cacert.pem'
  File "./virtualenv.py", line 2334, in <module>
    main()
  File "./virtualenv.py", line 712, in main
    symlink=options.symlink)
  File "./virtualenv.py", line 946, in create_environment
    download=download,
  File "./virtualenv.py", line 903, in install_wheel
  File "./virtualenv.py", line 796, in call_subprocess
    % (cmd_desc, proc.returncode))
OSError: Command "/Users/pradyunsg/.ve... e n v/bin/python3.6" - setuptools pip wheel failed with error code 1
```

pradyunsg:
Cool. This patch now works. :)

```shell
git clone https://github.com/pypa/pip --depth=1 --quiet
git clone https://github.com/pypa/virtualenv --depth=1 --quiet

pushd virtualenv > /dev/null
echo "Applying the Patch"
wget --quiet https://github.com/pypa/virtualenv/pull/1094.diff
git apply 1094.diff > /dev/null
rm 1094.diff > /dev/null

echo "Testing 9.0.1"
python ./virtualenv.py 'testing-env' --no-download > /dev/null

source ./testing-env/bin/activate > /dev/null
pip --version

echo "Building wheel for pip dev"
popd > /dev/null

pushd pip > /dev/null
python setup.py -q bdist_wheel 2> /dev/null
popd > /dev/null

deactivate > /dev/null

pushd virtualenv > /dev/null
rm virtualenv_support/pip-9.0.1-py2.py3-none-any.whl
cp ../pip/dist/pip-10.0.0.dev0-py2.py3-none-any.whl virtualenv_support

echo "Testing 10.0.0"
python ./virtualenv.py 'testing-env' --no-download > /dev/null

source ./testing-env/bin/activate > /dev/null
pip --version
deactivate > /dev/null
popd > /dev/null
```

Gives me:

```
Applying the Patch
Testing 9.0.1
pip 9.0.1 from /Users/pradyunsg/.venvwrap/venvs/tmp-de0674286021596/virtualenv/testing-env/lib/python2.7/site-packages (python 2.7)
Building wheel for pip dev
Testing 10.0.0
pip 10.0.0.dev0 from /Users/pradyunsg/.venvwrap/venvs/tmp-de0674286021596/virtualenv/testing-env/lib/python2.7/site-packages/pip (python 2.7)
```

msarahan:
This works fine, but can I ask why virtualenv is continuing to rely on private vendorings in pip?  This seems like very tight coupling that can cause problems (and has, here necessitating this change).  

Why not add certifi as an actual dependency and get the certs in a "normal" way, rather than reaching into pip's internals?

pradyunsg:
@msarahan I'll quote @dstufft from over at pip:

> This mess with virtualenv/get-pip.py/ensurepip essentially flows from the fact that we're trying to use pip to install pip, so there doesn't yet exist a command to subprocess-- this is effectively a use case that is entirely unsupported unless you're one of us.

msarahan:
Let me explain how I encountered this, in the hopes of illustrating that this is a fragile fix.

I work for Anaconda.  We struggled with the vendoring of requests in pip, which was itself vendoring other things.  The long path lengths caused issues on Windows.  Our first attempt was to un-vendor pip.  That ended up being very fragile for two reasons:

1. We had a buggy old package that when removed was keeping some folders around that convinced pip that it was still vendored
2. People were installing new, incompatible versions of html5lib

So, I set off to re-vendor pip.  I saw that on pip's master branch, requests is no longer vendored, matching some newer version of requests.  Great!  I spliced that into pip 9.0.1, and to make sure that things worked, I ran pip's test suite.

No tests worked at all, because virtualenv crashed with arcane tracebacks like:

```
_____________________ ERROR at setup of test_check_clean ______________________

tmpdir = Path('C:\\Users\\builder\\AppData\\Local\\Temp\\pytest-of-msarahan\\pytest-0\\test_check_clean0')
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x00000127B46F3780>
isolate = None

    @pytest.fixture
    def virtualenv(tmpdir, monkeypatch, isolate):
        """
        Return a virtual environment which is unique to each test function
        invocation created inside of a sub directory of the test function's
        temporary directory. The returned object is a
        ``tests.lib.venv.VirtualEnvironment`` object.
        """
        # Force shutil to use the older method of rmtree that didn't use the fd
        # functions. These seem to fail on Travis (and only on Travis).
        monkeypatch.setattr(shutil, "_use_fd_functions", False, raising=False)

        # Copy over our source tree so that each virtual environment is self
        # contained
        pip_src = tmpdir.join("pip_src").abspath
        shutil.copytree(
            SRC_DIR,
            pip_src,
            ignore=shutil.ignore_patterns(
                "*.pyc", "__pycache__", "contrib", "docs", "tasks", "*.txt",
                "tests", "pip.egg-info", "build", "dist", ".tox", ".git",
            ),
        )

        # Create the virtual environment
        venv = VirtualEnvironment.create(
            tmpdir.join("workspace", "venv"),
>           pip_source_dir=pip_src,
        )

tests\conftest.py:145:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\lib\venv.py:45: in create
    obj._create(clear=clear)
tests\lib\venv.py:55: in _create
    no_wheel=True,
C:\Miniconda3\lib\site-packages\virtualenv.py:945: in create_environment
    download=download,
C:\Miniconda3\lib\site-packages\virtualenv.py:901: in install_wheel
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd = ['C:\\Users\\builder\\AppData\\Local\\Temp\\pytest-of-msarahan\\pytest-0\\test_check_clean0\\workspace\\venv\\Scripts\\python.exe', '-', 'setuptools']
show_stdout = False, filter_stdout = None, cwd = None
raise_on_returncode = True
extra_env = {'JYTHONPATH': 'C:\\Miniconda3\\lib\\site-packages\\virtualenv_support\\setuptools-28.8.0-py2.py3-none-any.whl;C:\\Min..._support C:\\Miniconda3\\lib\\site-packages\\virtualenv_support', 'PIP_NO_INDEX': '1', 'PIP_ONLY_BINARY': ':all:', ...}
remove_from_env = None
stdin = b'\nimport sys\nimport pkgutil\nimport tempfile\nimport os\n\nimport pip\n\ncert_data = pkgutil.get_data("pip._vendor....argv[1:]\n\n    sys.exit(pip.main(args))\nfinally:\n    if cert_file is not None:\n        os.remove(cert_file.name)\n'

    def call_subprocess(cmd, show_stdout=True,
                        filter_stdout=None, cwd=None,
                        raise_on_returncode=True, extra_env=None,
                        remove_from_env=None, stdin=None):
        cmd_parts = []
        for part in cmd:
            if len(part) > 45:
                part = part[:20]+"..."+part[-20:]
            if ' ' in part or '\n' in part or '"' in part or "'" in part:
                part = '"%s"' % part.replace('"', '\\"')
            if hasattr(part, 'decode'):
                try:
                    part = part.decode(sys.getdefaultencoding())
                except UnicodeDecodeError:
                    part = part.decode(sys.getfilesystemencoding())
            cmd_parts.append(part)
        cmd_desc = ' '.join(cmd_parts)
        if show_stdout:
            stdout = None
        else:
            stdout = subprocess.PIPE
        logger.debug("Running command %s" % cmd_desc)
        if extra_env or remove_from_env:
            env = os.environ.copy()
            if extra_env:
                env.update(extra_env)
            if remove_from_env:
                for varname in remove_from_env:
                    env.pop(varname, None)
        else:
            env = None
        try:
            proc = subprocess.Popen(
                cmd, stderr=subprocess.STDOUT,
                stdin=None if stdin is None else subprocess.PIPE,
                stdout=stdout,
                cwd=cwd, env=env)
        except Exception:
            e = sys.exc_info()[1]
            logger.fatal(
                "Error %s while executing command %s" % (e, cmd_desc))
            raise
        all_output = []
        if stdout is not None:
            if stdin is not None:
                proc.stdin.write(stdin)
                proc.stdin.close()

            stdout = proc.stdout
            encoding = sys.getdefaultencoding()
            fs_encoding = sys.getfilesystemencoding()
            while 1:
                line = stdout.readline()
                try:
                    line = line.decode(encoding)
                except UnicodeDecodeError:
                    line = line.decode(fs_encoding)
                if not line:
                    break
                line = line.rstrip()
                all_output.append(line)
                if filter_stdout:
                    level = filter_stdout(line)
                    if isinstance(level, tuple):
                        level, line = level
                    logger.log(level, line)
                    if not logger.stdout_level_matches(level):
                        logger.show_progress()
                else:
                    logger.info(line)
        else:
            proc.communicate(stdin)
        proc.wait()
        if proc.returncode:
            if raise_on_returncode:
                if all_output:
                    logger.notify('Complete output from command %s:' % cmd_desc)
                    logger.notify('\n'.join(all_output) + '\n----------------------------------------')
                raise OSError(
                    "Command %s failed with error code %s"
>                   % (cmd_desc, proc.returncode))
E               OSError: Command C:\Users\builder\App...v\Scripts\python.exe - setuptools failed with error code 1

C:\Miniconda3\lib\site-packages\virtualenv.py:797: OSError
```

No mention of cacert.pem in there, though there is the stdin bit, which is what eventually led me to the right place.  This wasted a fair amount of my time.  Your fix here addresses the thing that bit me, but it leaves you vulnerable to similar future breakages.

This reacharound use of pip internals is hostile to developers.  Please consider directly vendoring certifi, or otherwise vendoring cacert.pem directly.

dstufft:
We can't depend on certifi because one of the use cases we support is running ``virtualenv.py`` with no installation. We also don't want to bundle our certificate file because it's one more that needs to be updated, and that makes it more likely it will forget to be updated.

While the particular error was not great, depending on certifi still wouldn't have allowed things to "just work", since you still would have had to adapt the imports to handle the correct new path (although I do acknowledge that has a better and easier to understand error message).

Ultimately, I don't think the cost is worth it, virtualenv is tied to the internals of pip, and that is why we bundle a specific version of pip.

msarahan:
> that is why we bundle a specific version of pip.

this bundling imposes a pretty heinous circular dependency across your build and test phases.  If pip didn't depend on virtualenv for testing, or if virtualenv used its bundled version of pip when testing pip, I'd totally agree with you.  I get that virtualenv uses pip internals, and that's not what I'm arguing against.  It's pretty weak to say that it's easy to forget to update a cert vendor.  That's the kind of thing that can be readily automated into a release process.

dstufft:
The pip test suite uses the ``--no-pip`` flag to virtual environment, so this bit of code never gets called as part of the pip test suite I *think*, but even if it did it would use it's bundled version to create the virtual environment, and then the pip test suite manually installs pip into that virtual environment on it's own.

msarahan:
How does my traceback arise, then?  That traceback comes because virtualenv is using my development install of pip, which I changed the vendorings in.  It is not using any virtualenv-internal version of pip.  If it were, it would not have broken for me.

dstufft:
I don't know what specifically you've patched to cause that, but the pip test suite is currently passing with a virtualenv without this pull request added, using the old APIs and locations. So either my recollection of how it works is wrong AND the Travis and Appveyor environments hit some edge case that somehow didn't expose the bug OR there's something triggered by the Anaconda patches or the environment in which you're running tests in.

I don't have time at the moment to dig in more, but I suspect it's the latter since it's hard for me to think of a scenario that would cause our CI to continue to pass if virtualenv didn't use it's bundled version.

pradyunsg:
> if virtualenv didn't use it's bundled version.

It does.

msarahan:
Here's our patches:

1. Move src/pip/_vendor from master branch to pip/_vendor in 9.0.1 branch: https://github.com/AnacondaRecipes/pip-feedstock/blob/master/recipe/0001-update-vendorings-to-match-master.patch
2. fix up references to vendored requests dependencies that are now pip-vendored dependencies: https://github.com/AnacondaRecipes/pip-feedstock/blob/master/recipe/0002-fix-up-references-to-vendored-requests-deps.patch

@pradyunsg I have evidence that it is using my development version of pip (because it's looking in the new vendored requests folder, and not finding cacert.pem).  Where should it be looking?  How do you expect it to look there?  What evidence can I show you that in my case, it is not using whatever bundled pip version you expect?

dstufft:
Have you patched virtualenv at all?

msarahan:
no, I have not touched virtualenv, and I hope to not require that.

pradyunsg:
With pip 10, an unpatched version of virtualenv won't work. This circular dependency is just the nature of how tightly pip-virtualenv are coupled. Any major changes in one are bound to affect the other. 

This patch fixes that issue.

> no, I have not touched virtualenv, and I hope to not require that.
> [snip]
> @pradyunsg I have evidence that it is using my development version of pip

I don't doubt you there but I also don't have time to dig into it. Sorry! :|

Off the top of my head, if you've vendored a newer version of requests with pip 9, you'd have to patch virtualenv to look in `pip/_vendor/certifi` for `cacert.pem`, instead of inside requests.

msarahan:
> This patch fixes that issue.

No, this patch does not fix the tight coupling.  It very slightly loosens the coupling to a wider range of versions.

> you'd have to patch virtualenv to look in pip/_vendor/certifi for cacert.pem, instead of inside requests

My approach has been to copy (would like to symlink, but this is windows) the vendored certifi/cacert.pem into _vendor/requests.  I really don't want to patch two things for one change.  This is why I'm complaining about the coupling.  If you take this opportunity to reduce the coupling, it fixes this problem in the future.  No, it won't fix past virtualenv versions.  It will, however, ensure that virtualenv doesn't break in the future because pip changed the location of cacert.pem.

pradyunsg:
> No, this patch does not fix the tight coupling. It very slightly loosens the coupling to a wider range of versions.

Oops. I missed a line! >.<

> With pip 10, an unpatched version of virtualenv won't work. This circular dependency is just the nature of how tightly pip-virtualenv are coupled. Any major changes in one are bound to affect the other.

pip 10 moves `main` to `pip_internal` and the location of `cacert.pem` changes due to changes in `requests` which breaks virtualenv.

> This patch fixes that issue.

pradyunsg:
> If you take this opportunity to reduce the coupling, it fixes this problem in the future.

I won't want to make that call.

Feel free to propose a PR for that if you want to see it changed, so that there can be a discussion about that there. This PR is really not the correct place for discussing that, especially because this is originating from the patching of pip 9.

msarahan:
Thanks.  I will work on a PR alternate to this PR, and cross reference this PR there.  FWIW, I will also try to figure out why virtualenv isn't using its bundled pip, and if there's any way to correct that.

pradyunsg:
IIRC, you need to use `--no-download` with virtualenv to _force_ it to use the bundled versions.

> I will work on a PR alternate to this PR

I'm skeptical of that but go ahead.

On that note, I'm gonna stop with this and get to this big report I have to write.

msarahan:
> IIRC, you need to use --no-download with virtualenv to force it to use the bundled versions.

Thanks.  The issue may only be that pip's test suite isn't doing that correctly.  I'll be very, very happy if that's the case, rather than some crummy subtle PYTHONPATH-type module load order issue.

> On that note, I'm gonna stop with this and get to this big report I have to write.

Yes, we all have more important things to do.  I wish I hadn't needed to waste the time figuring out this problem, and I hope I can help make sure it isn't a problem for anyone else in the future.

msarahan:
This is a crummy, subtle PYTHONPATH/sys.path issue.  I have been calling my tests from the root of a cloned git pip repo.  From the python docs at https://docs.python.org/3/using/cmdline.html,

> ``-``
  Read commands from standard input (sys.stdin). If standard input is a terminal, -i is implied.
  If this option is given, the first element of sys.argv will be "-" and the current directory will be added to the start of sys.path.

The key bit there is 

> the current directory will be added to the start of sys.path

So it doesn't matter how much virtualenv tries to put pip first.  My git clone is the front of sys.path, and when this happens in virtualenv.py::install_wheel:

```
    SCRIPT = textwrap.dedent("""
        import sys
        import pkgutil
        import tempfile
        import os

        import pip
```

It will currently always be importing pip from my git clone's pip folder.

So, I'm sorry for the noise.  The fix here isn't that the cacert.pem stuff needs to change, it's that we need a way to help ensure that virtualenv is using its bundled pip in preference to the front of sys.path here, or a better documented way of how to run tests to avoid this bad behavior.  Maybe this is all a moot point because it's a crappy artifact of how I'm trying to graft the newer vendored dependencies into the old folder structure - you all have already avoided this with the shift to the src folder.  I probably need to adapt more of master to 9.0.1 in order for this to work.

Anyway, thanks for humoring me.

msarahan:
For the record, the ways that I was trying to run tests:

* tox -e py26 (in CentOS6 docker image, no conda at all)
* python setup.py test (on windows)
* py.test tests

all executed in the root of my git clone of the pip repo.

pradyunsg:
Much like what I was expecting, it's an issue with your testing methodology. Glad that's sorted out.

One way to run pip's test is to use tox.

msarahan:
umm... you can say it's my testing methodology, but I wouldn't say I'm doing anything crazy or out of line here.  I did use tox, and I did encounter this same error.  This is something of an unfortunate coincidence in package naming, and a frequent point of confusion among python newbies (and an easy way to bite not-so-newbies).  I do think that this is reasonably unique to my hack here, and that pip has addressed the issue adequately with its source tree restructure (nesting things under the src folder), and that my discussion here need not continue.  Sorry for the interruption.

dstufft:
FWIW it's possible that this problem would occur in other situation too for actual end users if their environment is significantly wonky. I wouldn't feel bad about accepting a PR that attempts to better guard virtualenv against errant pip's on the ``sys.path`` taking precedence over our bundled version. Perhaps by directly manipulating ``sys.path`` in our little "script" instead of relying on ``PYTHONPATH`` handling.

dstufft:
> actual end users

Sorry, I didn't mean to imply you weren't an end user, specifically I meant in non-testing situations.

pfmoore:
FWIW, I routinely got weird situations as a result of working in a pip checkout and getting the "local" pip rather than the installed one. So @msarahan I feel your pain - but I believe that the move to a `src` directory has fixed it (it has for me, at least).

Glad you got to the bottom of your issue :-)

pradyunsg:
@pfmoore @dstufft Will we merge this now or will we wait until we're closer to pip 10's release?

I don't think any changes are needed.

Issue #864, opened 2016-02-03
The -p N.M shortcut doesn't work for Python 3.5 on Windows


pfmoore:
The `--python` command line can take an abbreviated interpreter name. On Unix this is one of the versioned alias names (`python2.7`, `python3`, ...) and is looked up on `PATH`.

On Windows the versioned names don't exist so the registry is searched for the Python version specified. So, `-p 2.7` finds the location of Python 2.7 in the registry and uses that.

However, the registry layout is not well documented, and is complex (user installs, 32-bit vs 64-bit). It also seems to have changed in Python 3.5, and `-p 3.5` does not work. Rather than attempt to fix this, though, I propose that the feature be marked as deprecated, and left as it stands. It's not clear that it was ever widely used.

People wanting to use an abbreviated interpreter name would be better to run `py -N.M -m virtualenv` (assuming virtualenv is installed under Python N.M) or `py -N.M C:\path\to\virtualenv` to run a standalone copy of virtualenv. This uses the (fully supported) Python launcher to select the required Python version.

Ivoz:
+1, never even knew that was a thing.

pfmoore:
IIRC, it's a thing because I added it because I was too lazy to type in the full interpreter. Then  I never actually used it. Go figure.

Yeah, I'll see if I can work up a proper deprecation note. For now, I mainly raised the issue so we had somewhere that said "it's broken and not going to get fixed" to justify me skipping the test I just added for the feature :-)

jurko-gospodnetic:
@asottile - this is actually important for your `pre-commit` project as it wants to be able to easily request a specific Python version to run a pre-commit hook, but does not want to duplicate the effort of detecting the installed Python executable location.

PR #1112, opened 2017-12-11
improve windows python install detection

jurko-gospodnetic:
This fixes issue #864.

Updated detecting Windows Python installations:

- now correctly detects current user installation and not just those installed at the system level
- for Python versions 3.5+ (which track 32-bit & 64-bit installations separately), recognize version tags X.Y, X.Y-32 & X.Y-64 where X.Y represents the 64-bit installation if available or 32-bit otherwise

Touched up & improved `resolve_interpreter()` tests:

- tests no longer permanently modify the `virtualenv` module under test (`virtualenv.is_executable` changes were leaking from some tests)
- tests now check that get_installed_python() results unrelated to the given version tag do not affect the result
- simple patch function's return values now given in `@patch` decorators to make the test code more compact

Tested `resolve_interpreter()` with registered python installations.

Tested `get_installed_pythons()` on and off Windows platform.

# Open issues

- [x] look into #582 changes to see if old Python installers perhaps give us a better way to detect 32-bit installations on 64-bit OSs
image
    - looked into this and it seems that nothing relevant gets stored in the `WOW6432Node` registry hive for either 32 or 64-bit installations on a 64-bit OS, at least for Python versions 2.7+, which are the only ones supported by virtualenv atm (tested with 2.7.13, 3.4.0, 3.5.0 & 3.6.3)
- [x] see what the `virtualenv_embedded` stuff is about and whether the changes done in this pull request need to be moved to different location
    - seems `virtualenv.py` is the only source file containing the WIndows installation detection logic

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -15,11 +15,12 @@ def test_version():
     assert virtualenv.virtualenv_version, "Should have version"
 
 
-@patch('os.path.exists')
-def test_resolve_interpreter_with_absolute_path(mock_exists):
+@patch('virtualenv.is_executable', return_value=True)
+@patch('virtualenv.get_installed_pythons', return_value={'foo': 'bar'})
+@patch('os.path.exists', return_value=True)
+def test_resolve_interpreter_with_absolute_path(
+        mock_exists, mock_get_installed_pythons, mock_is_executable):
     """Should return absolute path if given and exists"""
-    mock_exists.return_value = True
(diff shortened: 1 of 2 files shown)


jurko-gospodnetic:
Travis failure is just due to the tests getting run on a no longer supported Python 2.6 platform - fixed in pull request #1113.

gaborbernat:
Put this inside https://github.com/pypa/virtualenv/pull/1219, @jurko-gospodnetic thanks!

PR #1113, opened 2017-12-11
update testing notes and tox configuration

jurko-gospodnetic:
- added docs on running the internal test suite using `tox`
- now you can pass in additional pytest arguments when calling tox by listing them on the command-line after `--`, e.g. `tox -e python3.6 -- -x -v --pdb`
- removed Python 2.6 from all lists of tested Python environments as running virtualenv in that environment will just report an error about that Python version not being supported
- added a test for creating a Python environment in a path with spaces
- added a test for creating a Python environment while locating the needed wheel sources in a path with spaces

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -56,7 +56,7 @@ and then run the tests with a specific Python version using a call like::
 
     $ python -m tox -e python3.6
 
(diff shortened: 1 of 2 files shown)


cclauss:
Related to #1138

PR #1118, opened 2018-01-05
Avoid duplicate prepend to PATH

wrp:
If the activate script is called multiple times, the virtual env path gets prepended multiple times to the PATH.  This is undesirable behavior.


Ivoz:
Why?

wrp:
Because having duplicate entries in a PATH is annoying and wasteful and
renders the activate "script" non-idempotent. If the virtual environment is
already in the PATH for some other reason, activate will currently make it
a duplicate entry.  However, this is based on the premise that virtualenv
behaves in a sane manner, but frankly the practice of sourcing the main
script to begin the environment is so fundamentally flawed that there's
really no chance of saving this software.  I've closed the pull request
because I really don't think there's much point.  Rather than sourcing
activate to begin and setting up a fragile 'deactivate' function in the
host shell, it would be *much* cleaner to start a new shell and then exit
it to exit the virtual environment.  AFAICT, the only reason to source the
activate script instead of starting a new shell is to preserve the prompt,
and that's not really a good trade off for the fragility.

tobiasherp:
@wrp: Well, it's not only the `PROMPT`; I can think of more possible reasons not to create a new shell, e.g. the command history, or other shell functions. Some of my shell scripts and shell functions expect an active virtual environment, and others don't; in such cases, I `. activate` or `deactivate`, respectively, and simply repeat the command from the history.
I can imagine situations which make it desirable to have a completely fresh shell which could be created by another method, e.g. a `createshell` script; it should have a slightly different prompt convention, and it's `deactivate` function (it should have one, I say) should simply emit an error message about exiting the shell instead, and fail.

PR #1125, opened 2018-01-24
Cleanup pulled from #628

slide:
Some general clean-up items pulled forward from https://github.com/pypa/virtualenv/pull/628 in preparation for a PR to support IronPython.


virtualenv.py
```diff
@@ -54,7 +54,8 @@
 
 is_jython = sys.platform.startswith('java')
 is_pypy = hasattr(sys, 'pypy_version_info')
-is_win = (sys.platform == 'win32')
+is_win = sys.platform == 'win32' or os.name == 'nt' or getattr(os, '_name', None) == 'nt'
+is_cpython = not is_jython and not is_pypy
```
Ivoz: I'd be much happier with an affirmative definition of cpython, rather than "well it hasn't identified itself as alternatives... so probably cpython?" complement approach. If not possible, then you could rearrange the branching logic to be easier on the eyes through negating the conditions

virtualenv.py
```diff
@@ -75,7 +76,7 @@
 
 # Return a mapping of version -> Python executable
 # Only provided for Windows, where the information in the registry is used
-if not is_win:
+if not is_win or not is_cpython:
```
Ivoz: For example, this could be re-arranged to simply be `if is_win:` <following block of code> and `else:` the simple two lines returning an empty dict.

I am guessing you're wanting to add a `is_cpython` to differentiate ironpython in the future, as a python on Windows, from cpython on Windows.

However, virtualenv doesn't know ironpython yet exists? So there is no reason/*need* to really add a `is_cpython` flag in this particular PR. I think the extra logic could go in the one supporting IronPython itself.

virtualenv.py
```diff
@@ -1232,7 +1233,7 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
         executable = sys.executable
         shutil.copyfile(executable, py_executable)
         make_exe(py_executable)
-        if is_win or is_cygwin:
+        if is_cpython and (is_win or is_cygwin):
```
Ivoz: Same feeling here as above.

virtualenv.py
slide: I haven't found a way to determine that cpython is the one being run on. I will try and research this, as it would definitely be cleaner.

slide:
Hmmm, not sure why it pulled in the 3.6 stuff. I'll resubmit a cleaner PR

PR #1131, opened 2018-02-11
Update setuptools to most recent python2.6 compatible version

### LINKED ISSUE #1121: Package using setup.cfg metadata fails within virtualenv due to outdated setuptools
I'm currently in the process of refactoring a lot of Python code into multiple proper Python packages. What I noticed quickly is that each modules "setup.py" was a slightly modified copy&paste. Fortunately I found that setuptools allows to factor everything variable into a separate setup.cfg:

  https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

The feature needs a setuptools >= December 2016. Not a problem in early 2018 I thought ;-)

When I install the package inside a virtual environment (e.g. to run tox tests) all kinds of odd effects happen. This is a demo of the effect:

```
E:\temp>mkdir venv

E:\temp>c:\Python36\python.exe -m venv venv

E:\temp>venv\Scripts\activate
(venv) E:\temp>pip install e:\Dev\Pype\cstutil\dist\cstutil-0.0.zip
Processing e:\dev\pype\cstutil\dist\cstutil-0.0.zip
Installing collected packages: UNKNOWN
  Running setup.py install for UNKNOWN ... done
Successfully installed UNKNOWN-0.0

(venv) E:\temp>pip install --upgrade setuptools
Collecting setuptools
  Using cached setuptools-38.4.0-py2.py3-none-any.whl
Installing collected packages: setuptools
  Found existing installation: setuptools 28.8.0
    Uninstalling setuptools-28.8.0:
      Successfully uninstalled setuptools-28.8.0
Successfully installed setuptools-38.4.0

(venv) E:\temp>pip install e:\Dev\Pype\cstutil\dist\cstutil-0.0.zip
Processing e:\dev\pype\cstutil\dist\cstutil-0.0.zip
Collecting six (from cstutil==0.0)
  Using cached six-1.11.0-py2.py3-none-any.whl
Requirement already satisfied: setuptools>=30.3.0 in e:\temp\venv\lib\site-packages (from cstutil==0.0)
Installing collected packages: six, cstutil
  Running setup.py install for cstutil ... done
Successfully installed cstutil-0.0 six-1.11.0
```

As you can see the old setuptools version just ignores the metadata. What makes matters worse is that the requirement "setuptools >= 30.3.0" of the package also doesn't work! 

After several "pip install --upgrade setuptools" runs and a lot of scratching my head I discovered that the old setuptools come from within virtualenv.  Is there a chance that the version can be bumped?

Using "install_requires" as parameter in setup.py would allow me to express the setuptools-dependency in a way setuptools 28 understands. Then however setting the (package specific) list of other dependencies would need to go into setup.py again -- defying the whole idea of pushing this information into setup.cfg.

gluap:
The version of setuptools 28.8.0 bundled with virtualenv 15.1.0 is quite outdated. Since virtualenv still supports python2.6, it makes sense to upgrade to the latest setuptools version with python2.6 support: [setuptools 36.8.0](https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v3700). The next upgrade would then only be necessary once virtualenv drops python2.6 support, too.

Currently virtualenv 15.1.0 will create a virtualenv with the following installed:
```
paulg$ pip list installed
pip (9.0.1)
setuptools (28.8.0)
virtualenv (15.1.0)
```

The outdated version of setuptools caused me some irritation when working with packages making use of `setup.cfg` (introduced in [setuptools 30.3.0](https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v3030)). Once I realized that older setuptools versions were introduced by the bundled wheels, the obvious workaround is updating the setuptools in each and every virtualenv right after creation. However with increasing popularity of the `setup.cfg` feature more and more packages will cease to work with vanilla virtualenvs, so in order to minimize surprise I think it makes sense to bundle a more recent setuptools version. This would fix #1121 (which has been closed by @ThimoNeubauer without fix but the closing remark that he also would prefer an updated setuptools over manually updating each created virtualenv).

The wheel in this PR is the official version available from pypi, the md5 should be 7539ecafc99fbf5e62bbab6a4ee90a65:
https://pypi.python.org/pypi/setuptools/36.8.0


gluap:
Note: appveyor failing is likely a symptom of tox ditching 2.6 support in recent versions and not caused by an error introduced with this PR. The same appveyor error is seen in other PRs for the same reason. I'd make fixing this part of the PR but @jurko-gospodnetic already did it thoroughly in #1113.

If #1113 is adopted I'd suggest going straight for the newest setuptools instead of sticking with 2.6 suport and update the PR accordingly.

gluap:
Thanks for updating!

PR #1133, opened 2018-02-20
enable pip cache in appveyor and travis builds

thijstriemstra:
The reason for this pull request is the same reason caching was added to pip: we want to minimze the amount of traffic and load on the pypi site. Downloading any package(s) from pypi more than once is a waste of resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install: pip install tox
 script: tox
 
 sudo: false # Use container based builds
+cache: pip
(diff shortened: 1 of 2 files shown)


thijstriemstra:
Can someone let me know if this trivial change is ok?

pradyunsg:
It seems fine to me...

xavfernandez:
@thijstriemstra thanks :)

PR #1134, opened 2018-02-21
Cleanup pulled from #628

slide:
- Use platform.python_implementation() to determine if running on CPython
- Cleanup file copy code
- Remove dead code (is_executable_file)

The check on line 80 is required because of the change to is_win including the jython os._name check (line 58), this cleanup up the code on line 1448 though


slide:
Is there anything that I need to do on this to get it accepted?

gaborbernat:
Looking at the master somewhere this is already merged.

Issue #1128, opened 2018-01-26
execsitecustomize in site.py doesn't clear exceptions


zhangyangyu:
`execsitecustomize` in `site.py`, also `execusercustomize`, are somewhat outdated that they only silence `ImportError`. They now in virtualenv are just what they are in Python2.5. In Python2.5, although they don't catch exceptions, but `initsite` in `pythonrun.c` clears all exceptions. So any error in `sitecustomize.py` won't block the process to run. But in Python2.6+, `initsite` no longer clears exceptions, exceptions are catched in `site.py`, so exceptions raised in `sitecustomize.py` still won't block the process to run. But with virtualenv, new interpreters and old `site.py` means the process dies immediately at start if there is any error in `sitecustomize.py`. This behavior is not consistent with system one and the doc.

PR #1145, opened 2018-03-10
Fix execsitecustomize/execusercustomize in site.py

zhangyangyu:
Don't exit the process when exceptions other than ImportError are
raised while tring to import sitecustomize/usercustomize.

Fixes::#1128


gaborbernat:
Closing due to inactivity.

PR #1146, opened 2018-03-15
PowerShell activation script adds paths incorrectly

idavis:
Running `<envname>\Scripts\activate.ps1` does not construct the path to scripts correctly causing the scripts to not work on the path as expected. The `$BASE_DIR` and `$env:PATH` should now be set in a platform agnostic way.


idavis:
I'm not sure how to regen the virtualenv.py file yet.

gaborbernat:
Closing due to inactivity.

PR #1149, opened 2018-03-17
Upgrade bundled dependencies

dstufft:


dstufft:
Merging this, the failures on Appveyor are for 2.6 only, and are because of dependencies in the test chain not supporting 2.6. It works fine on Travis 2.6.

PR #1170, opened 2018-04-26
Restore support for Python 2.6

sergeykolosov:
Fixed Python 2.6 support broken after pip 10.0 release.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -937,7 +937,10 @@ def create_environment(home_dir, site_packages=False, clear=False,
             to_install.append('setuptools')
 
     if not no_pip:
-        to_install.append('pip')
+        if sys.version_info[:2] == (2, 6):
+            to_install.append('pip<10')
+        else:
+            to_install.append('pip')
 
     if not no_wheel:
         if sys.version_info[:2] == (2, 6):


avylove:
@ssbarnea I don't see why 2.6 support was removed in the first place. People are willing to submit pull requests to keep it working and as of yet there haven't been any conflicts with new code. The reality is a whole lot of people still have to support 2.6 code and will until RHEL 6 goes EOL. A lot of that code is actually new code written for Python 3, but that has to be 2.6 and 2.7 compatible. It makes it a whole lot easier to test that code against all those different versions of Python if tox can create all those environments.

avylove:
@gaborbernat Not sure you understand the use case. This is not a package dependency issue. virtualenv is the utility that allows you to test against other versions. So by taking away the ability to test against versions of Python people actually use, the message is either A.) Don't use virtualenv or B.) Don't upgrade virtualenv for the next few years. Neither seem like reasonable options. In the long term, a better design is probably needed, so there are fewer dependencies on the target Python interpreter. In the meantime, there's no reason to remove support for 2.6 if it doesn't break something else. Most of the people this affects do not want to be writing code for 2.6, but that's the reality and these kind of decisions just make life harder for people stuck in the middle. They are also probably not launching virtualenv with 2.6, (In my case my main interpreter is 3.6 or 3.7) but since it's one file called in both the launching environment and target environment, what can you do?

pradyunsg:
> In the meantime, there's no reason to remove support for 2.6 if it doesn't break something else.

There is additional maintenance overhead to ensure compatibility with a 2013 EOL'ed version of Python. It is also much easier to maintain source compatibility with 2.7+ and Python 3. It does break pip 10.0, as shown.

avylove:
If you look at 73404cb, where 2.6 was dropped. All it involved was getting rid of some pinned versions of wheel and setuptools as well as automated tests for 2.6. @sergeykolosov provide the pull request for pinning pip. None of the commits since then look like they would be impacted. So what exactly is the burden here? I maintain a lot of code that has to run on 2.6 through 3.7 and while there is occasionally special attention needed for 2.6, 95% of the time all it requires is an extra import like unittest2. If you're that worried, make 2.6 testing non-critical and depend on the community to provide 2.6 pull requests.

gaborbernat:
Code added must be considered by maintainers all time, otherwise they would break every other commit. Furthermore would prefer a cleaner implementation that's easier to reason about if possible rather than random magical edge case checks for python 2.6. Finally, would prefer to drop support in order to encourage people to migrate away from old, unsupported and unsecure versions.

dstufft:
Supporting that pin then means that we have to support multiple versions of pip inside of the virtual environment. And virtualenv's tests are woefully inadequate, and in many ways the best way we test virtualenv is through pip's test suite, which exercises the virtualenv code base much more rigorously than virtualenv's own test suite does.

You're free to continue to use the older version of virtualenv, or forking virtualenv to maintain compatibility with 2.6.

avylove:
> Code added must be considered by maintainers all time, otherwise they would break every other commit.

This is why there is automated testing

> Furthermore would prefer a cleaner implementation that's easier to reason about if possible rather than random magical edge case checks for python 2.6.

Sounds like a design flaw in virtualenv. I'd be very interested to hear a better solution

> Finally, would prefer to drop support in order to encourage people to migrate away from old, unsupported and unsecure versions.

This has zero impact on whether people run 2.6. It's determined by the OS and usually there are other downstream blockers preventing upgrading the OS. If you really want people to stop using 2.6, then I'd assume you make all your packages available as RPMs for Python 3.4 in EPEL 6. All you're accomplishing by dropping a Python version people still have to support is making things harder for the community.

avylove:
@gaborbernat This is not for running on RHEL 6. I'm not sure where you would get that idea. You seem to not understand the workflow of enterprise software in the community. So let me try to summarize it here. A lot of people write software for enterprise customers who are slow to change. There are various reasons for that, some external, some internal, but it's the reality. RHEL 7 has been out for 4 years and most enterprise customers I've dealt with are still 50% RHEL 6. So if you want to support these users, not only do you have to support RHEL 7 (Python 2.7) which is old by Linux standards, you have to support RHEL 6 (Python 2.6), and newer, non-enterprise distros like Fedora and Ubuntu.

While both EL6 and EL7 have Python 3.4 available (and 3.6 in EL7) through Software Collections and EPEL, package support for them is more limited. By that I mean RPM packages because that's how you install software on an enterprise system, so it is easily verifiable and auditable.

I personally prefer to do development on Fedora, and upgrade often. People use all kind of different distros. I currently use Python 3.6 as my primary interpreter when I do development. Lots of people use different distros and interpretors, but tools like tox (using virtualenv) make it easy to test against different Python interpretors. That is until they don't.

So again, what you're saying is either
A.) I shouldn't upgrade virtualenv on my Fedora machine until RHEL 6 is EOL
or
B.) I should find an alternative to virtualenv
If either of those is the case, then what mission is this project trying to fulfil?

gaborbernat:
@avylove The goal of this project is to allow creating virtual environments for all **supported** Python versions, perhaps with a bit more flexibility and configuration than the built-in Python 3 allows. The goal of tox is to allow testing your codebase against various**supported** Python version. Neither projects commit to work/support old Python versions that are no longer supported and indefinitely. That being said I do see a use case for this in a marginal edge case  so my proposal therefore is let's find a way you can work by not upgrading virtualenv, and I'm open to collaborate in that if we need to change things to achieve such. Therefore, let's progress in this direction please.

@Ekultek, your communication style are offensive, please adhere to our  code of conduct. My personal experience is that I did not have to write Python 2.6 code for the last 4 years, and did not see much request for it in general either. Most enterprise software value safety and security and an outdated Python 2.6 definitely does not serve those goals.

avylove:
@gaborbernat **supported** is a loaded term. It seems like you mean supported by The PSF, but that would exclude jython, pypy, tauthon, etc. 2.6 is under support from Red Hat, and still heavily used. Regardless, **supported** is not mentioned once in the PyPA [goals](https://www.pypa.io/en/latest/future/). But these are:

> - To provide a relatively easy to use software distribution infrastructure that is also fast, reliable and reasonably secure. “Reasonably secure”, due to backwards compatibility constraints preventing turning off some insecure legacy features.
> - To be progressive, but also be very mindful to not break things that are currently working, due to haste.
> 

I appreciate your willingness to work with me. I think we should figure this out now, because it's going to be an even bigger problem when 2.7 retirement comes along. Red Hat will support it on RHEL7 for 4.5 years beyond the 2020 date given by the PSF.

I think a lot of people don't realize that virtualenv primarily runs with the target Python interpreter. I'd argue this is not really necessary as long as the target Python can provide small amounts of information in a standard way like the location of the site file. By decoupling these it also might be possible to break up virtualenv into a package instead of one monolithic file, so it would be easier to maintain.

gaborbernat:
@avylove what I'm still missing here is why it's a problem if the latest version does not support deprecated versions (by PSF), but you have an older version you can pin and use.

avylove:
@gaborbernat I'm not sure how that would work. Pinning versions is easy for dependencies, but, via tox, I use virtualenv as a testing utility. As far as I know, there's no way to tell tox to only use a specific version of virtualenv to setup the tox enviornments. So I'd have to configure dnf (or pip in some cases) to prevent updating virtualenv on all of my test and development systems. Am I missing something?

ssbarnea:
I vote for closing this PR and not reinstanting support for py26. I have multiple reasons for proposing this, i know for sure few language feature that would make almost impossible to wrote code what works with both py26 and py37.
py26 become obsolete long time ago, let it rot. 

virtualenv depends on few other sensitive packages which already dropped support for py26 making any tentative to support it futile. 

These being sa(i)d, I hope that virtualenv will not drop support for py27 as soon PSF stops supporting it next year. There are going to be a considerable number of users still running distos based on py27 (like RHEL/CentOS7). I just want to be sure that we should not see a simiar discussion about py27 too soon ;)

Mainly I would say that the right moment to drop support for an old version is when it becomes extreamly hard to keep the producted tested on CI. If testing with py26 on Travis and AppVeyor no longer works, that is a good enough reason to close the door and move on.

dstufft:
We historically support the same versions of Python in pip and virtualenv, and the discussion about what versions to support basically revolve around usage of that version on PyPI. The more downloads we see hitting PyPI for a specific version, the less likely we are to stop supporting it. We also traditionally continue supporting any version of Python that has not yet reached its end of life date, regardless of the number of users using it.

gaborbernat:
@avylove so you're asking a tox feature not a virtualenv one then. that being said I think tox 2.7 or so would support python 2.7. So you just need to use that one too work with python 2.7.

ssbarnea:
I have two to confess that if py26 is made non voting (ignore testing result on CI), it may be ok to reinstate it. At least the py26 “support team” could spend time dealing with it without slowing down normal product development.  To me this sounds like a reasonable compromise.

This approach would also avoid the need to fork it, which is something really undesirable.

avylove:
@gaborbernat I'm writing software that gets tested on CPython 2.6 to 3.7 as well as pypy. tox, despite all it's features, is a wrapper for virtualenv. I primarily work on Fedora using Python 3.6. I use tox (currently version 2.9.1 since that's what ships with Fedora 29) to test my software. I'm launching it with Python 3.6, though I'll probably move to 3.7 soon. While tox 3.0.0 dropped 2.6 support, that should only affect the executing Python and not the target Python environments, correct? The support for target environments is dependent on virtualenv, which gets launched with 3.6, but relaunches itself with the python version specified.

@dstufft There are still millions of downloads a month for 2.6 packages (ex: six:378,746, botocore:464,950). This metric is skewed too, since a lot of 2.6 usage happens in enterprise environments without direct access to the internet, where the the majority of python packages are installed through system packages.

@ssbarnea That seems like a reasonable approach. 2.6 on travis works fine, appveyor lacks SNI support for pip on 2.6, so I've been skipping 2.6 testing on Windows which hasn't worried me since the demand seems to be primarily on Linux anyway. But yes, if it's non-voting, then leave it to those of us who want 2.6 support to figure it out.

gaborbernat:
But imagine a world where you can tell tox what virtualenv version you want to use to create the test environment with. In this world you can easily pin an old version of virtualenv for a py26. This is something on tox roadmap and probably will be implemented at some point. And while true this does not yet get shipped with Fedora, as far as tox goes this is a developer tool. you don't have to really on the system provided version, you can install the latest with --user, or create a virtualenv in which to install it and link that to use/bin one.

pfmoore:
The key "cost" of 2.6 support is the requirement that virtualenv core code is written using only Python 2.6 compatible code. That's a non-trivial impact for the maintainers.

I suggest that we don't support 2.6 (i.e. maintainers can ignore 2.6 compatibility when writing code) but we accept patches that change the code to be 2.6-compatible if it's submitted (presumably by people interested in keeping 2.6 compatibility). We don't have 2.6 in CI (or we have it, but mark the results as non-mandatory) so that we're not blocked on being 2.6-compatible. That allows the maintainers to work with 2.7+, while allowing interested parties to take on the burden of maintaining 2.6 compatibility.

We should also note that the maintainers won't consider bugs that can't be reproduced against 2.7 or later. If people want, we can allow them in the tracker, with a "2.6 only" label applied, as a help for people interested in supporting 2.6 (but the virtualenv maintainers will ignore them).

All of this should be clearly explained in the docs, so that people wanting 2.6 support know the situation.

dstufft:
I am -1 on that, if it involves bundling a second copy of pip again, which has a very real cost to our donated services via Fastly. The pip 9.0.3 wheel (the last version that supported 2.6) is 1.4MB large, in the last 30 days virtualenv was downloaded 22.6 million times, 22.5M * 1.4MB, is ~31TB of bandwidth added to PyPI, or roughly $2500/month of additional bandwidth charges to PyPI on a donated service.

It was done temporarily, but it was not a viable long term solution. If virtualenv is to continue to support, in any fashion, 2.6, then pip must also support in at least that same fashion, 2.6. Anything else is not only expecting volunteers to put in more effort to continue to maintain it but also expecting our donated services to bear the brunt of extra cost. Forking and uploading a new version of virtualenv to another name, or pinning to an old version doesn't have those same issues.

I'd note, that I'm pretty sure that this PR doesn't actually work anyways, since it only sets a version pin in the version of pip that is to be installed... but that won't work if the bundled version of pip doesn't execute on Python 2.6 either. It will need to be modified to also bundle a version of pip that supports 2.6. Of course setuptools no longer supports 2.6 nor does wheel so it'll also be required to bundle older versions of wheel and setuptools as well (because these enterprise customers don't have internet access, so we can't rely on the internet to download these supported versions).

Given all of these issues, I really don't think it's workable even on a best effort basis. At best the likely outcome is that we break it in pip X, someone notices and fixes it for pip X+1... which we won't release for another 3 months, and there's a likely chance that other changes that get merged in those 3 months re-break 2.6 (and since we don't have CI, we won't notice until pip X+1 is released, which pushes the fix time out to at a minimum of pip X+2). In the mean time, we're going to have those users putting pressure on the pip/virtualenv developers to release a hotfix patch to restore 2.6 support, which again adds additional overhead and effort for the pip/virtualenv developers to support 2.6. If we don't release a hot fix patch for 2.6 support... then people who need 2.6 are going to have to pin virtualenv in those intervening 3 (6, 9, whatever) months... and if they can pin virtualenv for those months, then surely they can just pin it for as long as they need to support 2.6.

gaborbernat:
The plan I want to achieve is drop Python 2.6 support on the virtualenv code base level, but do allow testing with Python 2.6 libraries via some kind of pinning and isolation.

avylove:
@dstufft You raise some good points, and I think they highlight some areas the designs of pip and virtualenv can be improved (for example: pypa/pip#5472). That said, pinning doesn't make sense and forking would could be confusing since it would require forking tox too or forking with the same name. Whatever happens, it's good we're discussing this now so we can avoid the same thing with 2.7.

I'm curious if those bandwidth costs are that bad with CDN caching. Is there a way those dependencies could be externalized? For example, downloaded on demand or provided as extras.

@gaborbernat Do you have a link for the tox proposal?

> The plan I want to achieve is drop Python 2.6 support on the virtualenv code base level, but do allow testing with Python 2.6 libraries via some kind of pinning and isolation.

I think this is where separating the execution and target functionality of virtualenv would make sense.

pfmoore:
@dstufft yeah, I'd not thought of that. The fact that we ship a specific version of pip pretty much kills the idea. Virtualenv can't support any versions of Python not supported by the bundled pip. And agreed we shouldn't bundle multiple versions.

There's no reason we couldn't take PRs that made `virtualenv --no-pip` work with 2.6, but that's an extremely niche case, so likely to be irrelevant.

dstufft:
The other wrinkle that occurred to me with that, our vendored dependencies in pip are *also* dropping support for Python 2.6... so that plan would also require us to either allow people to patch our dependencies inside of pip for 2.6 support (which is against our current policy) or would require them to get those dependencies to agree to allow 2.6 support in their code bases as well... which would then have to wait for those projects to release a 2.6 supporting version (and if they don't have CI, the same problem where any changes between that patch and the release risk rebreaking the next release silently), get that bundled in pip, etc.

My honest belief is that there are only two real options for supporting a version of Python, either we do support it and we expect that of our dependencies all the way up the chain and we test it in CI and the such... or we don't. Everything else is either fragile, expensive, or too niche to be very useful.

Ultimately, pretty much wherever you draw the cut off for support at, *SOMEONE* is going to want you to draw the line on the other side of them. It's a balancing act trying to get as many people on the supported side of the line, while being cognizant of the accumulated costs to the community for maintaining support for those versions all the way through the entire dependency chain. I remain firmly of the belief that 2.6 is niche enough that the cost isn't worth continuing to maintain support for it. The primary source of 2.6 users are enterprise users-- who can/should engage with their vendor who provided them with an ancient Python, and press them to support the other tools they require on their platform.

Beyond that, I think looking at reasons why pinning isn't acceptable, and figuring out what changes we can make to make it more acceptable is the best path forward.

avylove:
@dstufft While the primary users of 2.6 are in enterprise environments, the primary users of virtualenv with 2.6 support are not likely in these environments, but are developing for them (and everyone else) using more modern tools. As far as pinning, no one has been able to say how that would work without blocking system upgrades of virtualenv for the next year or two.

Tools like virtualenv occupy a special case. Why? Because when support is an issue, the answer is typically to use virtualenv and pin your dependencies. Support is not (or should not) be a single list. There is support for executing environments and target environments. virtualenv (and pip) have combined these which makes these support choices have a larger impact than necessary. So why not think about how to make the surface of support for target environments so small that it is not a barrier. There's no technical reason that can't be achieved, those were design decisions.

dstufft:
That's a possibility sure, one that involves a lot more effort but possibly workable yes. I'm not likely to be the one to undertake that work, but if you feel strongly about it, feel free to do that work to contribute that support.

gaborbernat:
I'll close this now as won't fix as no significant effort has been made on this on the last few months, leaving me to conclude there's not enough need for this.

PR #1171, opened 2018-05-01
Symlink the python interpreter unless --always-copy is specified

yrro:
Support for symlinking the python interpreter alongside the standard library, albeit only for Python 3.3 and higher.

I'm not totally sure that using `sys.executable` and `sysconfig.get_path('stdlib')` will cover all cases for testing, but hopefully the general approach makes sense.

This works for me if I directly invoke `python{2,7,3.6} setup.py test`, but when I run `tox` all versions fail with variations on the following:

```
ERROR: The executable /tmp/tmp8_vprfxd/venv/bin/python3 is not functioning
ERROR: It thinks sys.prefix is '/home/sam/src/virtualenv/.tox/python3.6' (should be '/tmp/tmp8_vprfxd/venv')
ERROR: virtualenv is not compatible with this system or executable
```

I'm not sure why yet, if you have any idea feel free to suggest...

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -3,7 +3,6 @@ import optparse
 import os
 import shutil
 import sys
-import sysconfig
 import tempfile
 import pytest
 import platform  # noqa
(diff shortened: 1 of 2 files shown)


Ashesh3:
Really useful +1

PR #1178, opened 2018-05-17
Include /path/to/ENV for activate

ryanlovett:
Be explicit about where `activate` is and maintain the convention used throughout.


Hendis:
[AUTHORS.txt](https://github.com/pypa/virtualenv/files/2065777/AUTHORS.txt)

PR #1189, opened 2018-07-03
Fixed failure to copy on platforms that use lib64

AWhetter:
The current behaviour attempts to copy a "lib" folder from the current directory, instead of copying the "lib" folder inside the virtualenv. Therefore the copying would fail.
This fixes that by always copying the "lib" folder from the virtualenv.


kxcc:
my application is failure, , when i update 16.0.0 ro 16.1.0!!!! because of you this issue

PR #1190, opened 2018-07-06
Export environment specific LD_LIBRARY_PATH

AuspeXeu:


0cjs:
If `LD_LIBRARY_PATH` was not set in the environment (i.e., not exported) before activation, it will be set but empty after activation followed by deactivation. I wonder, is this the behaviour we wish?

Also, related to this but not for this particular PR, is virtualenv now correctly copying the Python shared lib for Linux Python versions built with `--shared`? (This PR is half of what I need, and that's the other half.)

PR #1199, opened 2018-08-27
Add check if exec_dir exists before listing it's content

vstinner:
install_python() no longer fails if the exec_dir directory doesn't
exist.

Patch source:
https://src.fedoraproject.org/rpms/python-virtualenv/blame/check-exec_dir.patch?identifier=master

Patch added in this Fedora change:
https://src.fedoraproject.org/rpms/python-virtualenv/c/6b9ab1f32bfd159a7b78a34e2c80687b4bc7447c

Co-Authored-By: Michal Cyprian <michal@platform.sh>

### DIFF
--- a/src/virtualenv.py
+++ b/src/virtualenv.py
@@ -1262,8 +1262,9 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
             exec_dir = join(sys.exec_prefix, "Lib")
         else:
             exec_dir = join(sys.exec_prefix, "lib", py_version)
-        for fn in os.listdir(exec_dir):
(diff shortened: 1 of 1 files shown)


vstinner:
This patch comes from the Fedora package of virtualenv: https://src.fedoraproject.org/rpms/python-virtualenv

cc @mcyprian

gaborbernat:
Will be part of ``16.1`` to be released this week.

vstinner:
> Will be part of 16.1 to be released this week.

Cool, thank you!

gaborbernat:
@vstinner ``virtualenv 16.1.0`` is out

vstinner:
Thanks for keeping me in touch!

PR #1200, opened 2018-08-31
activate.csh handle newlines & don't use basename

gregory-nisbet:
Current `activate.csh` script is not robust against newlines in the prompt and uses `basename` unnecessarily.

For example, this `.tcshrc` file contains a prompt that breaks the script

```
onintr -

set tcshrc_nl = '\
'

 if ($?prompt) then
    set ellipsis = "..."
    set prompt = "$tcshrc_nl:q""%B[%.03]%b %# "
 endif

onintr
```

I modified the script to detect the presence of newlines and avoid changing the prompt at all in those cases.

The correct thing to do would probably be to add '[...]' to the beginning of the last line of the prompt, but I don't know of any easy way to accomplish that with csh. I briefly looked into `s///`-type substitution and using a temporary array variable to split the string into lines, but couldn't get it to work.

`tcsh` and `csh` also have the `:t` modifier for returning the last `/`-delimited component from a path, e.g.

```
[~] > set a = /a/b/c/d
[~] > echo $a:t
d
```

Also, I replaced added `:q` to every variable expansion (except for the `$?varname` ones) to guard against characters like newlines.


gregory-nisbet:
@gaborbernat The allow edits box on the sidebar is checked and I believe it was checked originally. I’m not sure why you’re having trouble editing the PR. What are you seeing when you try to edit it?

If you’d prefer, I can make the changes you suggest as a workaround.

gaborbernat:
I get a forbidden when I try to push changes 😃

gregory-nisbet:
I'm not sure what's going on. 

I unchecked the `allow edits` and waited for a green check mark to appear. Then I rechecked the `allow edits` and waited for a green check mark to appear. Can you push now?

If you cannot push, are you able to edit the PR through the website?

If neither thing works, I'll be happy to make changes if you post a patch file as a comment or something similar.

gaborbernat:
Never mind, seems just had the wrong remote :+1:

PR #1206, opened 2018-09-28
Xonsh Support

scopatz:
This aims to add xonsh support to virtualenv.  

I am not sure how to generate the top-level `ACTIVATE_XSH` variable in `virtualenv.py`, and this probably needs to be tested, but otherwise, should be good to go! 

Feedback welcome!


scopatz:
CC @pganssle

scopatz:
And feel free to push directly to this branch.

gaborbernat:
Tests?

scopatz:
@gaborbernat - any recommendations on how to add tests?

scopatz:
@gaborbernat - also I know this will fail because of the `ACTIVATE_XSH` issue.

pganssle:
This seems like the kind of thing where `xonsh` needs to be added to the matrix in Travis (i.e. all tests should be run using `xonsh` as the shell), though I don't know if `tox` will handle that as expected..

gaborbernat:
Let me have a better look at things and I'll get back to you in a few weeks with some proper advice.

pfmoore:
I'm not sure we should add activate scripts for yet more shells. It's already hard enough to find people with sufficient expertise to maintain feature parity across all the shells we do support (sh, csh, fish, cmd, powershell).

Rather than this, would anyone be interested in a PR to allow users to locally customise the activation scripts installed in a new virtualenv? I'd be willing to write such a PR if it meant that we focused on only supporting a limited number of shells (and basic functionality for them) in the supplied scripts, leaving users with different needs to use the customisation option to handle them.

pganssle:
@pfmoore I think there's a pretty strong case for including xonsh in the list of supported shells considering the nature of xonsh as a "shell but with Python". Not to volunteer him or anything, but I think that given that @scopatz is a xonsh maintainer and an active participant in the Python community, the question of finding a domain expert is somewhat moot in this case.

That said, I  do agree that it doesn't necessarily scale to have to maintain support for *every* shell in the upstream package, so providing some clear hooks for "shell support plugins" (and possibly breaking out *all* current shell support into individual plugin packages maintained by domain experts) seems like a net win.

If the timeline for "provide hooks to allow a 'xonsh support' plugin" is on the order of a few months, then I think it probably makes sense to wait. If the timeline is on the order of a year or more, I think it makes sense in the short term to add xonsh as a supported shell and then possibly use it as a test case for moving support from within virtualenv to a separate package.

gaborbernat:
I think let's add this now. In parallel we can then afterwards come up with an extensible API people can use to add their own, and then we can deprecate all this.

scopatz:
Hi All, thanks for your feedback.  I am very happy to keep maintaining this (and not likely to go anywhere, as @pganssle points out :wink:)  I just need to know what to do here to get this to work.

virtualenv.py
```diff
@@ -1452,6 +1452,7 @@ def install_activate(home_dir, bin_dir, prompt=None):
             'activate.bat': ACTIVATE_BAT,
             'deactivate.bat': DEACTIVATE_BAT,
             'activate.ps1': ACTIVATE_PS,
+            'activate.xsh': ACTIVATE_XSH,
```
pganssle: I think this needs to actually be defined, thought apparently all the others are defined [here](https://github.com/pypa/virtualenv/blob/ed553318227ed1ec6415b6785eac9add084c7c21/virtualenv.py#L1981), which is some crazy nonsense I don't understand.

pfmoore: The `ACTIVATE_xxx` variables hold base64-encoded copies of the actual activate scripts. They are embedded in the `virtualenv.py` script so that we don't need to ship them as separate files (the `virtualenv_embedded` directory isn't shipped as part of the wheel).

`bin/rebuild-script.py` rebuilds `virtualenv.py` from the files in the `virtualenv_embedded` directory. Some fiddling will be needed to add `ACTIVATE_XSH` to that mechanism (but I don't know precisely what).

scopatz: @pganssle - you mentioned you had this working somewhere?

pganssle: @scopatz I made a PR against your branch [here](https://github.com/scopatz/virtualenv/pull/1)

pganssle:
I'm not seeing any shell-specific tests in the test suite. How are these intended to be tested?

scopatz:
Yeah, I was just looking at that.  I can certainly translate `test_activate.sh` to xonsh if you would like.

scopatz:
I have updated this, resolving conflicts from master

gaborbernat:
thanks can you add a test similar to what we have for powerscript for activation?

src/virtualenv.py
```diff
@@ -2130,6 +2134,18 @@ def convert(s):
 """
 )
 
+##file activate.xsh
```
gaborbernat: this now is under just ``# file``

virtualenv_embedded/activate.xsh
```diff
@@ -0,0 +1,39 @@
...
+$_OLD_VIRTUAL_PATH = $PATH
+$PATH = $PATH[:]
+$PATH.add($VIRTUAL_ENV + _get_sep() + "__BIN_NAME__", front=True, replace=True)
+
+if ${...}.get("PYTHONHOME", ""):
+    # unset PYTHONHOME if set
+    $_OLD_VIRTUAL_PYTHONHOME = $PYTHONHOME
+    del $PYTHONHOME
+
+aliases['pydoc'] = ['python', '-m', 'pydoc']
```
gaborbernat: needs an empty end of file newline

src/virtualenv.py
scopatz: done!

virtualenv_embedded/activate.xsh
scopatz: Thanks for catching this.

scopatz:
I think this is ready for another round of review

virtualenv_embedded/activate.xsh
```diff
@@ -0,0 +1,39 @@
+"""Xonsh activate script for virtualenv"""
+from xonsh.tools import get_sep as _get_sep
```
gaborbernat: you'll need to install/package xonsh on ci for this to work

tests/activation/test_xonsh_activation.py
```diff
@@ -0,0 +1,49 @@
...
+    return pytest.mark.powershell(
+        need_executable("xonsh", (XONSH_COMMAND, "--version"))(fn)
+    )
+
+
+def print_python_exe_path():
+    return "{} -c 'import sys; print(sys.executable)'".format(virtualenv.expected_exe)
+
+
+@need_powershell
```
gaborbernat: powershell? I think you want xonsh

tests/activation/test_xonsh_activation.py
```diff
@@ -0,0 +1,49 @@
...
+def test_activate_with_powershell(tmpdir, monkeypatch):
+    monkeypatch.chdir(tmpdir)
```
gaborbernat: yeah, drop all powershell references

scopatz: Uggh Thanks for catching this.  I think I have gotten them all now.

gaborbernat:
xonsh is not available in CI so I'm expecting this to fail. we'll need to make it available as part of this ci.

gaborbernat:
also formatting is off, see https://dev.azure.com/pypa/virtualenv/_build/results?buildId=1860&view=logs

scopatz:
Yeah, sorry, it took me awhile to figure out how the test dependencies were installed.

scopatz:
@gaborbernat - Any idea why xonsh doesn't show up on the PYTHONPATH on just Windows?

gaborbernat:
Not sure, do you need the .bat? Maybe try to see what the path is, by printing it and seeing it.

gaborbernat:
Could be that on Windows the Path is not set, maybe try to make it first pass by having absolute path by re-using sys.executable.

scopatz:
I think that this is finally all green and ready to go!

gaborbernat:
@scopatz just one more thing, let's add a changelog entry announcing it 🎉

scopatz:
@gaborbernat, sure thing!  Where should I make those edits at? `docs/changes.rst`?

gaborbernat:
@schopatz yes

setup.py
```diff
@@ -125,6 +125,7 @@ def get_version():
             "pytest >= 3.0.0, <4",
             "coverage >= 4.5.0, <5",
             'pytest-timeout >= 1.3.0, <2; platform_python_implementation!="Jython"',
+            'xonsh;python_version>"2.7"',
```
gaborbernat: Shouldn't this be just ``>3``? What's xonosh version support?

scopatz: Xonsh only supports Python 3 (v3.4+)

virtualenv_embedded/activate.xsh
```diff
@@ -0,0 +1,39 @@
...
+
+    if ${...}.get("_OLD_VIRTUAL_PATH", ""):
+        $PATH = $_OLD_VIRTUAL_PATH
+        del $_OLD_VIRTUAL_PATH
+
+    if ${...}.get("_OLD_VIRTUAL_PYTHONHOME", ""):
+        $PYTHONHOME = $_OLD_VIRTUAL_PYTHONHOME
+        del $_OLD_VIRTUAL_PYTHONHOME
+
+    if 'VIRTUAL_ENV' in ${...}:
```
gaborbernat: we should be consistent and use ``"`` inside this script too.

setup.py
scopatz: Happy to change this if you want.

gaborbernat: If that's the case yes please change it so. Also we need to test that on Python 2.7 we don't generate the activation script.

pganssle: Just to be clear - you should still be able to create a Python 2.7 virtualenv *from* Python 3 in xonsh, right?

setup.py
```diff
@@ -125,7 +125,7 @@ def get_version():
             "pytest >= 3.0.0, <4",
             "coverage >= 4.5.0, <5",
             'pytest-timeout >= 1.3.0, <2; platform_python_implementation!="Jython"',
-            'xonsh;python_version>"2.7"',
+            'xonsh;python_version>"3"',
```
gaborbernat: Let's be more explicit here; aka to say ``3.4`` is what I initially meant.

setup.py
scopatz: @pganssle - yes that should still be possible.  You just can't create a vitrualenv from Python 2.7 because you wouldn't be able to run xonsh in the first place.

src/virtualenv.py
```diff
@@ -1542,6 +1542,10 @@ def install_activate(home_dir, bin_dir, prompt=None):
 
     files["activate_this.py"] = ACTIVATE_THIS
 
+    if majver >= 3:
```
gaborbernat: you should minor version qualify here the same as you do with the install before anyone with Python 3.3 turns up that it's not working

tests/activation/test_xonsh_activation.py
```diff
@@ -0,0 +1,57 @@
...
+
+def print_python_exe_path():
+    return "{} -c 'import sys; print(sys.executable)'".format(virtualenv.expected_exe)
+
+
+@need_xonsh
+def test_activate_with_xonsh(tmpdir, monkeypatch):
+    monkeypatch.chdir(tmpdir)
+    home_dir, _, __, bin_dir = virtualenv.path_locations(str(tmpdir.join("env")))
+    virtualenv.create_environment(home_dir, no_pip=True, no_setuptools=True, no_wheel=True)
```
gaborbernat: how is this test passing under python 2 🤔 if xonosh cannot run with it

src/virtualenv.py
scopatz: done!

tests/activation/test_xonsh_activation.py
scopatz: The test seems to be appropriately skipped on linux, windows, and mac on Python 2.7.  Are you seeing something different?

scopatz:
The azure failure is an `report_coverage Job` failure, which is unrelated to anything I have done here, I believe.

tests/activation/test_xonsh_activation.py
gaborbernat: Oh I understand, it's not installed on python 2.7 so it's skipped. This works by chance as https://github.com/pypa/virtualenv/blob/master/tests/lib/__init__.py#L14 should mean we never skip inside the CI 😮 please add some logic inside the needs_xonosh to always skip if xonosh is not installed.

scopatz: Ok! Does it look good now?

scopatz:
Woohoo! Thanks for all the help here!

PR #1210, opened 2018-10-04
Fix instructions for download of virtualenv.

zvezdan:
Replaced pypi.python.org with pypi.org in the download URL.
Added `-L` option to `curl` to follow redirects from randomized hash
URLS.

Also, replaced pypi.python.org/pypi with pypi.org/project throughout the
docs.

This fix addresses the most important part of #1042.


gaborbernat:
Thanks!

PR #1221, opened 2018-10-25
Add linter and code formatter

### LINKED ISSUE #1218: paint it black :-)
In part of making the code easier to maintain I plan to install pre-commit hooks (@asottile), and paint the code black via https://github.com/ambv/black (cc @ambv) - flake8 check too. 

This will probably follow shortly after figuring out the new CI. This project is hard enough without needing to think about formatting and style.

cc @dstufft

gaborbernat:
Adds pre-commit and black. Resolves #1218.

---
This change is [Reviewable](https://reviewable.io/reviews/pypa/virtualenv/1221)


gaborbernat:
The CI fails for now because of https://github.com/asottile/aspy.refactor_imports/issues/28

src/virtualenv.py
```diff
@@ -1893,13 +1890,11 @@ def after_install(options, home_dir):
...
     py_exe = "python%s" % python_version
-    content = ("#!/usr/bin/env %s\n" % py_exe) + "## WARNING: This file is generated\n" + content
+    content = ("#!/usr/bin/env %s\n" % py_exe) + "# WARNING: This file is generated\n" + content
     return content.replace("##EXT" "END##", extra_text)
 
 
-##EXTEND##
-
-
+# EXTEND
```
rinrinne: Due to this change, created bootstrap script does not have any extra_text.
https://virtualenv.pypa.io/en/latest/reference/#creating-your-own-bootstrap-scripts

With virtualenv 16.0.0
```
$ python -c "import virtualenv; print(virtualenv.create_bootstrap_script('HELLOWORLD'))"| grep HELLOWORLD
HELLOWORLD
$
```

With virtualenv 16.1.0 (same as master)
```
$ python -c "import virtualenv; print(virtualenv.create_bootstrap_script('HELLOWORLD'))"| grep HELLOWORLD
$
```
Could you fix it?

gaborbernat: yeah let's open an issue for it 😎

rinrinne: Sorry too late. I created issue #1244

Issue #1209, opened 2018-10-04
Powershell activate script is not generated on *NIX


bbhoss:
I'm using Powershell Core 6.1 and virtualenv 16. When generating an environment with Powershell on macOS, there is no scripts directory and no activate.ps1 that appears to be generated on Windows. Now that PS is cross-platform, virtualenv should generate a ps1 file for usage on those other platforms.

gaborbernat:
Fixed via #1223.

PR #1223, opened 2018-10-26
Generate activate.ps1 on all systems

karazijal:


karazijal:
Issue #1209

gaborbernat:
@karazijal thanks for this; there's a slight concern here now: we add this to all environments but will it work if we don't use the platform dependent line ending?  maybe before copy, we should normalize to the platform line separator 🤔 can you check and test this?

tests/test_cmdline.py
```diff
@@ -9,8 +10,19 @@
 
 
 def test_commandline_basic(tmpdir):
-    """Simple command line usage should work"""
-    subprocess.check_call([sys.executable, VIRTUALENV_SCRIPT, str(tmpdir.join("venv"))])
+    """Simple command line usage should work and files should be generated"""
+    venv_dir = str(tmpdir.join("venv"))
+    venv_bin_dir = os.path.join(venv_dir, "bin")
```
gaborbernat: note this is not always bin, but is platform dependent :+1:

ferdinandvanwyk:
Helping out on this as part Bloomberg PyPA sprint

gaborbernat:
closing in favor if #1229

PR #1229, opened 2018-10-27
Generate activate.ps1 for all platforms

ferdinandvanwyk:
Continuing work from #1223


tests/test_cmdline.py
```diff
@@ -9,8 +10,19 @@
 
 
 def test_commandline_basic(tmpdir):
-    """Simple command line usage should work"""
-    subprocess.check_call([sys.executable, VIRTUALENV_SCRIPT, str(tmpdir.join("venv"))])
+    """Simple command line usage should work and files should be generated"""
+    venv_dir = str(tmpdir.join("venv"))
+    venv_bin_dir = os.path.join(venv_dir, "Scripts" if sys.platform == "win32" else "bin")
```
gaborbernat: this seems a bit brittle (easy to break), maybe instead re-use the path locations logic from https://github.com/pypa/virtualenv/blob/0571b2ef5a612fa1c6a61c205b631de47aa1fc7f/src/virtualenv.py#L1025 by calling to that function to find out the bin directory

ferdinandvanwyk: That function will also make the directory, but I will add a flag to skip this `mkdir` line when running the tests.

https://github.com/pypa/virtualenv/blob/0571b2ef5a612fa1c6a61c205b631de47aa1fc7f/src/virtualenv.py#L1035

gaborbernat:
``powershell.exe`` it is, but seems our tests fail :thinking: will need to investigate and fix

ferdinandvanwyk:
Nice simplifications, thanks. I'm in the process of setting up a Windows VM to test this more easily, but it looks like most of the issues have been solved. We're left with a weird python path issue:

```bash
2018-10-30T14:07:02.0012981Z >       assert after_activate == os.path.join(bin_dir, "python.exe" if virtualenv.is_win else "python")
2018-10-30T14:07:02.0013275Z E       AssertionError: assert 'C:\\Users\\V...s\\python.exe' == 'c:\\users\\vs...s\\python.exe'
2018-10-30T14:07:02.0014890Z E         - C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_activate_with_powershell0\\env\\Scripts\\python.exe
2018-10-30T14:07:02.0015084Z E         ? ^   ^      ^  ^  ^^^^^^^^^^  ^  ^     ^      ^
2018-10-30T14:07:02.0016045Z E         + c:\\users\\vssadm~1\\appdata\\local\\temp\\pytest-of-unknown\\pytest-0\\test_activate_with_powershell0\\env\\Scripts\\python.exe
```

gaborbernat:
🐿️ time to ship this, thanks @karazijal @ferdinandvanwyk!

PR #1231, opened 2018-10-27
Refactor sys.stdout.write to print

minggli:
there were a number of print statements in `virtualenv.py`. Converted them to sys.stdout.write.
This PR refactor print statements into sys.stdout.write to be consistent to remaining code base.

**reverse**
This PR refactors all sys.stdout.write to print for brevity and readability.

### DIFF
--- a/src/virtualenv.py
+++ b/src/virtualenv.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 """Create a "virtual" Python installation"""
-from __future__ import print_function
 
 # fmt: off
 import os  # isort:skip
@@ -276,7 +275,8 @@ class Logger(object):
             if self.level_matches(level, consumer_level):
                 if self.in_progress_hanging and consumer in (sys.stdout, sys.stderr):
                     self.in_progress_hanging = False
-                    print("", flush=True)
+                    print("\n", end="")
+                    sys.stdout.flush()
                 if rendered is None:
                     if args:
                         rendered = msg % args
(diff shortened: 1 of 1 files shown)


minggli:
Hi @gaborbernat thanks for reviewing. I'm fine with either ways as long as there is consistency. can you share why favour print over sys.stdout.write?

gaborbernat:
Does not need import and shorter. Plus it's how most people write to the standard out so is fewer surprises for someone new to the project.

minggli:
cool, makes sense.

src/virtualenv.py
```diff
@@ -305,20 +303,17 @@ def end_progress(self, msg="done."):
         if self.stdout_level_matches(self.NOTIFY):
             if not self.in_progress_hanging:
                 # Some message has been printed out since start_progress
-                sys.stdout.write("..." + self.in_progress + msg + "\n")
-                sys.stdout.flush()
+                print("..." + self.in_progress + msg, flush=True)
```
gaborbernat: please make this a format, instead of the pluses :+1: thanks

gaborbernat:
the end was good, just add ``from __future__ import print_function``

gaborbernat:
please run ``tox -e fix-lint`` before you commit :-)

minggli:
@gaborbernat fixed. Review welcome!

gaborbernat:
you're welcome

PR #1238, opened 2018-10-31
As an interim solution, suppress the warning that the imp module is d…

pfmoore:
…eprecated


pfmoore:
@gaborbernat What's the failure here? It seems to be reporting lint problems on a load of files I didn't change :-(

pfmoore:
Looks like master is also failing the same check. Am I therefore good to merge this?

pfmoore:
@gaborbernat Can this go into 16.1? I know it's a bit last minute, but I hadn't realised no-one had addressed it till just now :-(

gaborbernat:
@pfmoore sure.

pfmoore:
Cool, thanks :-)

gaborbernat:
@pfmoore ``virtualenv 16.1.0`` is out

PR #1245, opened 2018-11-13
Fix output for python -m site

merwok:


gaborbernat:
You don't need to download you can just read.

merwok:
Do you have the link for that?  Didn’t find it

gaborbernat:
just click on run test.

merwok:
Ready!

PR #1249, opened 2018-11-29
made lib64 symlink relative again

### LINKED ISSUE #1248: lib64 symlink can break subtly since 16.1
## tldr
When creating a virtualenv on a path that contains a symlink, the virtualenv break if this symlink is removed due to the changes made in https://github.com/pypa/virtualenv/commit/5d560144432642a91f6e1661bed22e637226c8fe

 ## context
When running tests, we create a workspace folder for each test, which has a symlink to a shared virtual env. The first test case populates the virtualenv. Then the first testcase's workspace is removed. 

Prior to 16.1 this worked as expected. However, in 16.1 the lib64 symlink is absolute (but not canonical). So when the workspace of the first test is removed, the shared virtual env breaks. The next test run produces the error "module pip not found". 

## fix 
revert the change on line 1622 (https://github.com/pypa/virtualenv/commit/5d560144432642a91f6e1661bed22e637226c8fe#diff-51642b24ac9529f4baa0f483fa257d2dL1622)

wouterdb:
fixes #1248


wouterdb:
can I trigger a rebuild in some way?

wouterdb:
I skipped the new test case on windows, even if it only failed on windows on python 2.7, how do I make the skipping more fine-grained?

gaborbernat:
Why would you want to skip it on Windows?

tests/test_virtualenv.py
```diff
@@ -304,3 +304,31 @@ def test_always_copy_option():
                 )
     finally:
         shutil.rmtree(tmp_virtualenv)
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="requires working symlink implementation")
```
gaborbernat: instead of platform hardcoding check if gettattr os symlink is None or not 👍

src/virtualenv.py
```diff
@@ -1635,9 +1635,9 @@ def fix_lib64(lib_dir, symlink=True):
     if os.path.lexists(lib64_link):
         return
     if symlink:
-        os.symlink(lib_dir, lib64_link)
+        os.symlink("lib", lib64_link)
```
gaborbernat: Should this be ``lib64`` on some systems? Please add a changelog entry too.

src/virtualenv.py
```diff
@@ -1478,8 +1478,8 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
             raise e
 
     proc_stdout = proc_stdout.strip().decode("utf-8")
-    proc_stdout = os.path.normcase(os.path.abspath(proc_stdout))
-    norm_home_dir = os.path.normcase(os.path.abspath(home_dir))
+    proc_stdout = os.path.normcase(os.path.realpath(proc_stdout))
+    norm_home_dir = os.path.normcase(os.path.realpath(home_dir))
```
gaborbernat: should have some comment here why this need to be realpath :thinking:

src/virtualenv.py
wouterdb: it has always been `lib` (also `lib_dir` is an absolute path to `lib`), as I understand it, virtualenv always populates `lib` and creates `lib64` as a symlink to `lib` on platforms that search in `lib64`

gaborbernat:
thanks

PR #1252, opened 2018-12-05
Don't fail on missing pip/certifi's cacert.pem

hroncok:
See https://github.com/pypa/pip/blob/master/src/pip/_vendor/README.rst

It is acceptable to remove the pip/_vendor/requests/cacert.pem file from pip

Virtualenv should not error on that

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -3,15 +3,18 @@ import os
 import shutil
 import sys
 import tempfile
+import zipfile
 
 import pytest
 
 import virtualenv
 
 try:
+    from pathlib import Path
     from unittest.mock import NonCallableMock, call, patch
 except ImportError:
     from mock import NonCallableMock, call, patch
+    from pathlib2 import Path
 
 
 def test_version():
@@ -332,3 +335,44 @@ def test_relative_symlink(tmpdir):
     shutil.rmtree(workdir)
 
     assert os.path.exists(lib64)
+
+
+def test_missing_certifi_pem(tmp_path):
+    """Make sure that we can still create virtual environment if pip is
+    patched to not use certifi's cacert.pem and the file is removed.
+    This can happen if pip is packaged by Linux distributions."""
+    proj_dir = Path(__file__).parent.parent
+    support_original = proj_dir / "src" / "virtualenv_support"
+    pip_wheel = sorted(support_original.glob("pip*whl"))[0]
+    whl_name = pip_wheel.name
+
+    wheeldir = tmp_path / "wheels"
+    wheeldir.mkdir()
+    tmpcert = tmp_path / "tmpcert.pem"
+    cacert = "pip/_vendor/certifi/cacert.pem"
+    certifi = "pip/_vendor/certifi/core.py"
+    oldpath = b"os.path.join(f, 'cacert.pem')"
+    newpath = "r'{}'".format(tmpcert).encode()
+    removed = False
+    replaced = False
+
+    with zipfile.ZipFile(str(pip_wheel), "r") as whlin:
+        with zipfile.ZipFile(str(wheeldir / whl_name), "w") as whlout:
+            for item in whlin.infolist():
+                buff = whlin.read(item.filename)
+                if item.filename == cacert:
+                    tmpcert.write_bytes(buff)
+                    removed = True
+                    continue
+                if item.filename == certifi:
+                    nbuff = buff.replace(oldpath, newpath)
+                    assert nbuff != buff
+                    buff = nbuff
+                    replaced = True
+                whlout.writestr(item, buff)
+
+    assert removed and replaced
+
+    venvdir = tmp_path / "venv"
+    search_dirs = [str(wheeldir), str(support_original)]
+    virtualenv.create_environment(str(venvdir), search_dirs=search_dirs)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,7 @@ Release History
 -------------------
 * The xonsh shell is now supported :pull:`1206`
 * bug fix: make lib64 symlink relative (as in <``16.1.0``) > :issue:`1248`
+* pip wheels with removed certifi's cacert.pem are now supported :pull:`1252`
 
 16.1.0 (2018-10-31)
 -------------------
--- a/src/virtualenv.py
+++ b/src/virtualenv.py
@@ -924,6 +924,8 @@ def install_wheel(project_names, py_executable, search_dirs=None, download=False
         except ImportError:
             from pip import main as _main
             cert_data = pkgutil.get_data("pip._vendor.requests", "cacert.pem")
+        except IOError:
+            cert_data = None
 
         if cert_data is not None:
             cert_file = tempfile.NamedTemporaryFile(delete=False)


hroncok:
Sure, tests are great.

What needs to be done in tests:

 * create a new virtual environment
 * install virtualenv package into it
 * patch the pip whl from the virtualenv_support package not to have and use certifi's cacert.pem but instead use `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem'
 * create a new virtual environment using that virtualenv

this currentl blows with:

```
(__venv__) [virtualenvtest]$ python -m virtualenv test_virtualenv
Using base prefix '/usr'
New python executable in /home/churchyard/tmp/virtualenvtest/test_virtualenv/bin/python

  Complete output from command /home/churchyard/tmp...irtualenv/bin/python - setuptools pip wheel:
  File "<stdin>", line 9, in <module>
  File "/usr/lib64/python3.7/pkgutil.py", line 637, in get_data
    return loader.get_data(resource_name)
OSError: [Errno 0] Error: 'pip/_vendor/certifi/cacert.pem'
  File "/usr/lib64/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/churchyard/tmp/virtualenvtest/__venv__/lib64/python3.7/site-packages/virtualenv.py", line 2462, in <module>
    main()
  File "/home/churchyard/tmp/virtualenvtest/__venv__/lib64/python3.7/site-packages/virtualenv.py", line 762, in main
    symlink=options.symlink,
  File "/home/churchyard/tmp/virtualenvtest/__venv__/lib64/python3.7/site-packages/virtualenv.py", line 1015, in create_environment
    install_wheel(to_install, py_executable, search_dirs, download=download)
  File "/home/churchyard/tmp/virtualenvtest/__venv__/lib64/python3.7/site-packages/virtualenv.py", line 968, in install_wheel
  File "/home/churchyard/tmp/virtualenvtest/__venv__/lib64/python3.7/site-packages/virtualenv.py", line 854, in call_subprocess
    raise OSError("Command {} failed with error code {}".format(cmd_desc, proc.returncode))
OSError: Command /home/churchyard/tmp...irtualenv/bin/python - setuptools pip wheel failed with error code 1
```

With my patch, it should not.

Now if I provide Python code that does and verifies the above, would you be able to guide me on where to stick that into the current test suite?

gaborbernat:
Put it at the end of https://github.com/pypa/virtualenv/blob/master/tests/test_virtualenv.py#L334

hroncok:
Added a test, verified that it fails without that fix.

tests/test_virtualenv.py
```diff
@@ -332,3 +334,38 @@ def test_relative_symlink(tmpdir):
...
 
     assert os.path.exists(lib64)
+
+
+def test_missing_certifi_pem(tmpdir):
+    """Make sure that we can still create virtual environment if pip is
+    patched to not use certifi's cacert.pem and the file is removed.
+    This can happen if pip is packaged by Linux distributions."""
+    tests_dir = os.path.dirname(__file__)
+    support_original = os.path.join(tests_dir, "../src/virtualenv_support/")
```
gaborbernat: this most certainly will fail on Windows; and indeed it has https://dev.azure.com/pypa/virtualenv/_build/results?buildId=3008 (not sure why the build does not show up here as Github check, although the job did start, you can see new jobs at https://dev.azure.com/pypa/virtualenv/_build?definitionId=11&_a=summary)

gaborbernat: Please also add a changelog entry about the change while you are at it.

hroncok:
OK, I have no idea what's the problem on Windows now or how to see any details :cry: 

Help me please?

tests/test_virtualenv.py
```diff
@@ -332,3 +334,40 @@ def test_relative_symlink(tmpdir):
     shutil.rmtree(workdir)
 
     assert os.path.exists(lib64)
+
+
+def test_missing_certifi_pem(tmpdir):
```
gaborbernat: use the new ``tmp_path`` please instead of ``tmpdir``. 👍

hroncok:
oh I forgot that it goes through subprocess and is not working with forward slashes.

tests/test_virtualenv.py
hroncok: the test above uses tmpdir. what is tmp_path?

gaborbernat: we'll migrate the old tests sometime in the future, but no point in not already use it for new tests - https://docs.pytest.org/en/latest/tmpdir.html?highlight=tmp_path#the-tmp-path-fixture

hroncok: oh, pathlib! may i use pathlib also instead of the os.path functions?

tests/test_virtualenv.py
```diff
@@ -332,3 +334,39 @@ def test_relative_symlink(tmpdir):
...
+    tests_dir = os.path.dirname(__file__)
+    proj_dir = os.path.dirname(tests_dir)
+    support_original = os.path.join(proj_dir, "src", "virtualenv_support")
+    pip_wheel = sorted(glob.glob(os.path.join(support_original, "pip*whl")))[0]
+    whl_name = os.path.basename(pip_wheel)
+
+    wheeldir = tmpdir / "wheels"
+    wheeldir.mkdir()
+    tmpcert = tmpdir / "tmpcert.pem"
+    cacert = "pip/_vendor/certifi/cacert.pem"
```
gaborbernat: does not the same path separator counts on this and below (hence why still failing on Windows)?

hroncok:
Windows still failing even with `os.path.join(proj_dir, "src", "virtualenv_support")`, no idea why. Also only on Python 3.

https://dev.azure.com/pypa/virtualenv/_build/results?buildId=3012&view=logs&jobId=debe4b1e-a6de-59f5-07ab-92771d5f9a9b&taskId=3601f2ed-d83f-5185-4e8b-01cc25e9c135&lineStart=42&lineEnd=42&colStart=38&colEnd=108

tests/test_virtualenv.py
hroncok: this is inside the zip but let me assert we found the file.

gaborbernat:
Once we get the tests passing I would like to ask you to break up into a failing test only commit, and then another on top that does the source change and fixes it.

tests/test_virtualenv.py
hroncok: asserts work, different problem

hroncok:
Azure now says Git fetch failed.

tests/test_virtualenv.py
hroncok: Done.

gaborbernat: yeah 👍

hroncok:
> Once we get the tests passing I would like to ask you to break up into a failing test only commit, and then another on top that does the source change and fixes it.

Shall I mark the test xfail in the first commit or not?

----

I have figured out that if the script I've changed fails, there is no error output anywhere in the test log, so I've temporarily enabled that output for now to see what's wrong on the Windows Python 3 code (maybe it's juts a different exception type). However the CI won't start.

gaborbernat:
No need for xfail just let it fail. if you ammend your commit it my trigger a redirect.

gaborbernat:
https://dev.azure.com/pypa/virtualenv/_build/results?buildId=3021

tests/test_virtualenv.py
```diff
@@ -332,3 +339,44 @@ def test_relative_symlink(tmpdir):
...
+    pip_wheel = sorted(support_original.glob("pip*whl"))[0]
+    whl_name = pip_wheel.name
+
+    wheeldir = tmp_path / "wheels"
+    wheeldir.mkdir()
+    tmpcert = tmp_path / "tmpcert.pem"
+    cacert = "pip/_vendor/certifi/cacert.pem"
+    certifi = "pip/_vendor/certifi/core.py"
+    oldpath = b"os.path.join(f, 'cacert.pem')"
+    newpath = "r'{}'".format(tmpcert).encode("ascii")
```
hroncok: So the Windows path has backslashes and if you put `\u` into str, it blows up :grinning:

gaborbernat:
thanks

hroncok:
Thanks to you.

tests/test_virtualenv.py
gaborbernat: easy to achieve by running in c:\users folder

PR #1267, opened 2018-12-26
make assertions on packaging, recommend tox for all testings

gaborbernat:
Resolves #909.


tox.ini
```diff
@@ -91,16 +91,16 @@ commands = pip wheel -w {envtmpdir}/build --no-deps .
...
-changedir = {toxinidir}/bin
+changedir = {toxinidir}/tasks
 extras =
-commands = python rebuild-script.py
+commands = python update_embedded.py
 
 [testenv:upgrade]
 description = upgrade pip/wheels/setuptools to latest
 skip_install = true
 extras = pip >= 18.1.0
```
webknjaz: @gaborbernat this is what I was talking about on Twitter

gaborbernat: that's indeed a bug 😅

webknjaz: PR incoming

gaborbernat: should be deps instead

webknjaz: Sure :)

Issue #1260, opened 2018-12-19
Virtualenv should not override cert option from pip.conf


daa:
Consider following setup: in our organization we have internal pypi server serving https with certificate signed with our internal root certificate authority, in ``/etc/pip.conf`` I have following settings:
```
[global]
cert = /etc/ssl/certs/ca-certificates.crt
```
to make ``pip`` use system certificates. Also I set ``PIP_INDEX_URL`` environment variable pointing to our internal pypi. ``pip`` works as expected with such settings.

The problem arises when ``virtualenv`` tries to install ``pip``, ``setuptools``, ``wheel`` and looks at our pypi for packages but does not use certificates from ``/etc/pip.conf`` because it's overridden by ``--cert`` option in ``install_wheel`` script:
```
            if cert_file is not None:
                args += ["--cert", cert_file.name{}]
```
It results in failure to install those requirements from our internal pypi. However ability to install latest ``pip`` and ``setuptools`` from our internal pypi on virtualenv setup is very desirable for me.

I know of workarounds like setting ``PIP_TRUSTED_HOST`` environment variable or setting ``trusted-host`` in ``/etc/pip.conf`` but I'd very like to not use them because they are not required for ``pip`` to work. Or if checking for ``cert`` in ``pip.conf`` is complicated I'd like to be able to specify option for ``virtualenv`` to use given certificates.

gaborbernat:
I'll try to think of a good solution for this. Thanks for reporting this, I agree we should allow some way to make this work.

daa:
What do you think about approach in #1273?

gaborbernat:
officially pip has no public interface other than the CLI, if you get the value via that I'm on board with this change.

pfmoore:
There's no means that I know of to get that information via pip's CLI. (A case could be made for exposing it via the CLI, but that would be a new feature). So at the moment, you have to use the internals if you need that data.

The code in question already uses pip's internals (the `main` function, and the location of the cert files in the vendored modules) so the proposed change doesn't add *that* much more of a problem. But you should carefully review this code with every pip release to ensure it still works. I think the code is on;y run against the embedded copy of pip that you ship with virtualenv - so the exposure is limited. If you run that code against an arbitrary copy of pip you'll have to be far more careful.

Ultimately, the internals you are using can break arbitrarily, with no warning, in any release of pip. As long as you're managing that risk, though, this is probably the only reasonable way of getting that data.

daa:
[...]

PR #1273, opened 2018-12-29
Do not override certificate bundle if it is defined in pip config

daa:
## Keep certificate option if it is set in pip config file or environment variable

This is one possible approach to solve #1260: script wrapping ``pip`` invocation first loads configuration and checks if ``cert`` option is defined there, and if it is defined does not override it with ``--cert`` command line argument.

I'd like to know your opinion on it. Personally I'm not very fond of depending on ``pip``'s internal modules but it should not be a big problem as ``virtualenv`` bundles and uses ``pip`` wheel of known version.

- [x] added/updated test(s)


spikerobinsonDLG:
This may have created a wider problem that --cert is never honoured, if the config file sets a value for cert? Was that intended / expected? I think I'm seeing that behaviour, that --cert is ignored in general, not just in the intended context (a new installation within virtualenv).

spikerobinsonDLG:
Specifically what is happening in my case is that the curl environment variable CURL_CA_BUNDLE is overriding the pip --cert setting. I can prove this by changing the CURL_CA_BUNDLE value and the behaviour of pip (with or without --cert) immediately changes. It suggests there is an inheritance where PIP_CERT inherits from CURL_CA_BUNDLE if that is set? In turn, the effect of this patch (#1273) is that the environment variable overrides the command line switch (--cert). I don't think this is desirable behaviour in the general case, outside of the special case of installing inside virtualenv.

gaborbernat:
This PR is for an old no longer maintained version of virtualenv.

daa:
Most likely it's an effect of ``requests`` using ``CURL_CA_BUNDLE`` to find the list of trusted CAs: https://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification. More specifically such behaviour is defined at https://github.com/psf/requests/blob/master/requests/sessions.py#L50-L67 and https://github.com/psf/requests/blob/master/requests/sessions.py#L717-L724

spikerobinsonDLG:
> This PR is for an old no longer maintained version of virtualenv.

Ah sorry I had misunderstood that. If this PR was only scoped to virtualenv and did not affect pip in general then probably my issue lies elsewhere.

gaborbernat:
This PR was scoped at an old version of virtualenv <16, version 20+ is a total rewrite so anything you see here is not relevant there.

spikerobinsonDLG:
> Most likely it's an effect of `requests` using `CURL_CA_BUNDLE` to find the list of trusted CAs: https://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification. More specifically such behaviour is defined at https://github.com/psf/requests/blob/master/requests/sessions.py#L50-L67 and https://github.com/psf/requests/blob/master/requests/sessions.py#L717-L724

Thanks for those references, much appreciated. Your refs make it clear that CURL_CA_BUNDLE is used if REQUESTS_CA_BUNDLE is empty.  I am still left wondering why --cert commandline switch in pip seems to be overriden by these environment variables.

Is the simple answer that I am using conda activate, and conda activate uses virtualenv under the hood?

spikerobinsonDLG:
Thanks for the help. It looks like I am commenting in the wrong place - I will try to get myself to the correct repo and comment there. I appreciate your patience!

daa:
``pip`` uses ``requests`` for HTTP communication.

PR #1282, opened 2019-01-14
Copy only required files from exec_prefix and only if it is really different from prefix path

### LINKED ISSUE #1270: 16.1.0 with --python creates symlink to distutils directory
PR #1227 seems to have created a regression in 16.1.0 as compared to 16.0.0 and earlier, at least on my dev systems where I use `make altinstall` to keep a bunch of different builds of Python for compatibility testing purposes. Now, ``lib/pythonX.Y/distutils`` in the virtualenv is created as a symlink to the distutils directory for my Python installation if `--python=pythonX.Y` is supplied, and then the subsequent ``install_distutils()`` call overwrites ``distutils/__init__.py`` in that Python's stdlib (following the symlink) with virtualenv's modified version. This of course results in all manner of later havoc as other callers into distutils end up getting virtualenv's init for it rather than the stdlib implementation.

Prior to the PR (or with it reverted), files in the virtualenv libdir are still symlinked as expected but the distutils directory is (correctly) not a symlink. I apologize for not having a better suggestion at this point than a revert, though hopefully someone with a deeper knowledge of the linkfarm bits will see a simple solution. As it stands, this basically renders tox unusable for me in conjunction with non-system-packaged Python interpreters.

Python built as follows:

    export TAG=3.7.2
    git clean -dfx
    git checkout v$TAG
    ./configure --enable-optimizations --prefix /home/fungi/lib/cpython/$TAG
    make profile-opt
    make altinstall
    ln -sf ../lib/cpython/$TAG/bin/python3.7 ~/bin/

Virtualenv installed with:

    python3.7 -m venv ~/lib/virtualenv
    ~/lib/virtualenv/bin/pip install virtualenv
    ln -sf ../lib/virtualenv/bin/virtualenv ~/bin/

Behavior reproduced with:

    virtualenv --python=python3.7 foo

With this, ``foo/lib/python3.7/distutils`` is created as a symlink to ``~/lib/cpython/3.7.2/lib/python3.7/distutils`` so then virtualenv overwrites the stdlib-supplied ``~/lib/cpython/3.7.2/lib/python3.7/distutils/__init__.py`` with the contents from the https://github.com/pypa/virtualenv/blob/master/virtualenv_embedded/distutils-init.py file.

daa:
When ``sys.exec_prefix`` contains relative parts but actually points to the same directory as ``sys.base_prefix`` or ``sys.prefix`` ``virtualenv`` unnecessarily copies files from it and can't set up virtual environment. This happens for example when distribution uses ``python-exec`` such as Gentoo, another manifestation is described in #1270.

Closes #1270.

- [x] added/updated test(s)


gaborbernat:
@daa should a symlink be enough to test, not?

fungi:
A symlink should be enough to test as long as it's to a relative rather than absolute path.

daa:
@gaborbernat Symlink will suffice, thank you for suggestion.

tests/test_virtualenv.py
```diff
@@ -471,3 +472,26 @@ def test_create_environment_with_local_https_pypi(tmpdir):
...
+        os.environ["PATH"] = os.pathsep.join(
+            p for p in os.environ["PATH"].split(os.pathsep) if not p.startswith(sys.prefix)
+        )
+    python = virtualenv.resolve_interpreter(os.path.basename(sys.executable))
+    try:
+        subprocess.check_call([sys.executable, "-m", "virtualenv", "-p", python, venvdir])
+        home_dir, lib_dir, inc_dir, bin_dir = virtualenv.path_locations(venvdir)
+        assert not os.path.islink(os.path.join(lib_dir, "distutils"))
+    finally:
+        os.environ["PATH"] = old_path
```
gaborbernat: note literals are unicode and windows requires str for os.environ (python 2)

daa: Thank you, seen that recently but forgot.

daa:
While working on the test for this case I've noticed that creation of virtual environment using Python from another virtual environment results in ``distutils`` in new environment being symlink to ``distuils`` from source environment. Not sure if it should be fixed because it makes no harm besides new virtual environment being dependent on the older and the fix may be complicated.
Reason is the same - copying files when ``exec_prefix`` does not match ``prefix``.

daa:
As mentioned in the related ticket linking ``distutils`` when creating virtual environment from virtual environment actually makes harm, to cure it I've limited copying from Python library directory inside ``sys.exec_prefix`` to required files only.

tests/test_virtualenv.py
```diff
@@ -518,3 +518,39 @@ def test_pypy_pre_import(tmp_path):
...
+    """Create virtual environment for Python with ``sys.exec_prefix`` pointing
+    to ``sys.prefix`` or ``sys.base_prefix`` or ``sys.real_prefix`` under a
+    different name
+    """
+    venvdir = str(tmpdir / "venv")
+    python_dir = tmpdir / "python"
+    python_dir.mkdir()
+    path_key = "PATH"
+    if six.PY2:
+        path_key = path_key.encode()
```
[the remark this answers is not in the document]
daa: Thank you.

daa:
Thank you!

andreineculau:
:clap:

any chance of a patch release soon to get this out?

gaborbernat:
Released as ``16.3.0``.
thanks

PR #1283, opened 2019-01-14
Discourage installation as root, such as with sudo

### LINKED ISSUE #1061: Stop recommending usage of sudo pip in installation documentation
Please suggest users to use pip --user instead of sudo pip to install virtualenv in https://virtualenv.pypa.io/en/stable/installation/ ; latter can break the system completely and runs setup.py from PyPI as root when pip doesn't support wheels

altendky:
Resolves #1061.


pfmoore:
Nice - very comprehensive and effective update. Thanks!

gaborbernat:
thanks

Issue #1284, opened 2019-01-14
ADMINISTRATIVE: CONTRIBUTING.rst & docstring of tasks/update_embedded.py appear out of date


bskinn:
[CONTRIBUTING.rst currently says](https://github.com/pypa/virtualenv/blob/59a5ab8ce460de950f97ffba8e1af6d79d530aa9/CONTRIBUTING.rst):

> * All changes to files inside `virtualenv_embedded` should be integrated to `virtualenv.py` with `bin/rebuild-script.py`.

If I read the project tree correctly, though, that update script is now [`tasks\update_embedded.py`](https://github.com/pypa/virtualenv/blob/59a5ab8ce460de950f97ffba8e1af6d79d530aa9/tasks/update_embedded.py).  As well, the script [docstring](https://github.com/pypa/virtualenv/blob/59a5ab8ce460de950f97ffba8e1af6d79d530aa9/tasks/update_embedded.py#L3) appears to be incorrect, as `virtualenv.py` is now updated from `virtualenv_embedded`, not `virtualenv_support`.

Small fixes, I'm happy to make the PR if I'm on the right track here.

gaborbernat:
Indeed we need to upgrade those. Though we no longer recommend calling those scripts directly, but instead via tox (see embed, fix_lint targets). We would love you to make a PR fixing this, thanks!

bskinn:
Ok -- so, contributors should:

 1.  Run `tox -e embed` after making any changes to the scripts in `virtualenv_embedded`; and
 2.  Run `tox -e fix_lint` before committing to feature branches

?

gaborbernat:
yes

bskinn:
Note to self .. temporarily setting `skip_install = False` in the `[testenv:fix_lint]` environment settings within `tox.ini` resolved a `.../.tox/fix_lint/bin/python: No module named virtualenv` error initially encountered.

bskinn:
This was fixed as part of #1330.

PR #1286, opened 2019-01-15
Update outdated docs referring to bin/rebuild-script.py

asottile:


docs/development.rst
```diff
@@ -15,9 +15,9 @@ pip.
 Files in the ``virtualenv_embedded/`` subdirectory are embedded into
 ``virtualenv.py`` itself as base64-encoded strings (in order to support
 single-file use of ``virtualenv.py`` without installing it). If your patch
-changes any file in ``virtualenv_embedded/``, run ``bin/rebuild-script.py`` to
-update the embedded version of that file in ``virtualenv.py``; commit that and
-submit it as part of your patch / pull request.
+changes any file in ``virtualenv_embedded/``, run ``tasks/update_embedded.py``
```
gaborbernat: Note in https://github.com/pypa/virtualenv/issues/1284 we agreed that this script shouldn't be used on its own but via tox.

asottile:
done!

PR #1287, opened 2019-01-20
produce a virtualenv zipapp

asottile:
this "works" -- I would like some feedback on the approach however:

```console
$ python3 tasks/make_zipapp.py 
zipapp created at /tmp/x/virtualenv/virtualenv.pyz
$ python virtualenv.pyz venv
New python executable in /tmp/x/virtualenv/venv/bin/python
done.
$ python3 virtualenv.pyz venv3
Using base prefix '/usr'
New python executable in /tmp/x/virtualenv/venv3/bin/python3
Also creating executable in /tmp/x/virtualenv/venv3/bin/python
done.
$ du -hs virtualenv.pyz 
2.0M	virtualenv.pyz
```

### DIFF
--- a/docs/changelog/1092.feature.rst
+++ b/docs/changelog/1092.feature.rst
@@ -1,0 +1,3 @@
+Enable virtualenv to be distributed as a ``zipapp`` or to be run as a
+wheel with ``PYTHONPATH=virtualenv...any.whl python -mvirtualenv`` - by
+Anthony Sottile
--- a/tasks/make_zipapp.py
+++ b/tasks/make_zipapp.py
@@ -1,0 +1,37 @@
+"""https://docs.python.org/3/library/zipapp.html"""
+import argparse
+import io
+import os.path
+import zipapp
+import zipfile
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--root", default=".")
+    parser.add_argument("--dest")
+    args = parser.parse_args()
+
+    if args.dest is not None:
+        dest = args.dest
+    else:
+        dest = os.path.join(args.root, "virtualenv.pyz")
+
+    bio = io.BytesIO()
+    with zipfile.ZipFile(bio, "w") as zipf:
+        filenames = ["LICENSE.txt", "virtualenv.py"]
+        for whl in os.listdir(os.path.join(args.root, "virtualenv_support")):
+            filenames.append(os.path.join("virtualenv_support", whl))
+
+        for filename in filenames:
+            zipf.write(os.path.join(args.root, filename), filename)
+
+        zipf.writestr("__main__.py", "import virtualenv; virtualenv.main()")
+
+    bio.seek(0)
+    zipapp.create_archive(bio, dest)
+    print("zipapp created at {}".format(dest))
+
+
+if __name__ == "__main__":
+    exit(main())
--- a/tests/test_zipapp.py
+++ b/tests/test_zipapp.py
@@ -1,0 +1,87 @@
+import os.path
+import subprocess
+import sys
+
+import pytest
+
+import virtualenv
+
+HERE = os.path.dirname(os.path.dirname(__file__))
+
+
+def _python(v):
+    return virtualenv.get_installed_pythons().get(v, "python{}".format(v))
+
+
+def _major_version(venv):
+    _, _, _, bin_dir = virtualenv.path_locations(venv)
+    exe = os.path.join(bin_dir, "python")
+    cmd = (exe, "-c", "import sys; print(sys.version_info[0])")
+    out = subprocess.check_output(cmd)
+    return int(out)
+
+
+@pytest.fixture(scope="session")
+def call_zipapp(tmp_path_factory):
+    if sys.version_info[:2] == (3, 4):
+        pytest.skip("zipapp was introduced in python3.5")
+    pyz = str(tmp_path_factory.mktemp(basename="zipapp") / "virtualenv.pyz")
+    subprocess.check_call(
+        (_python("3"), os.path.join(HERE, "tasks/make_zipapp.py"), "--root", virtualenv.HERE, "--dest", pyz)
+    )
+
+    def zipapp_make_env(path, python=None):
+        cmd = (sys.executable, pyz, "--no-download", path)
+        if python:
+            cmd += ("-p", python)
+        subprocess.check_call(cmd)
+
+    return zipapp_make_env
+
+
+@pytest.fixture(scope="session")
+def call_wheel(tmp_path_factory):
+    wheels = tmp_path_factory.mktemp(basename="wheel")
+    subprocess.check_call((sys.executable, "-mpip", "wheel", "--no-deps", "-w", str(wheels), HERE))
+    wheel, = wheels.iterdir()
+
+    def wheel_make_env(path, python=None):
+        cmd = (sys.executable, "-mvirtualenv", "--no-download", path)
+        if python:
+            cmd += ("-p", python)
+        env = dict(os.environ, PYTHONPATH=str(wheel))
+        subprocess.check_call(cmd, env=env)
+
+    return wheel_make_env
+
+
+def _test_basic_invocation(make_env, tmp_path):
+    venv = tmp_path / "venv"
+    make_env(str(venv))
+    assert venv.exists()
+    assert _major_version(str(venv)) == sys.version_info[0]
+
+
+def test_zipapp_basic_invocation(call_zipapp, tmp_path):
+    _test_basic_invocation(call_zipapp, tmp_path)
+
+
+def test_wheel_basic_invocation(call_wheel, tmp_path):
+    _test_basic_invocation(call_wheel, tmp_path)
+
+
+def _test_invocation_dash_p(make_env, tmp_path):
+    venv = tmp_path / "venv"
+    python = {2: _python("3"), 3: _python("2.7")}[sys.version_info[0]]
+    make_env(str(venv), python)
+    assert venv.exists()
+    expected = {3: 2, 2: 3}[sys.version_info[0]]
+    assert _major_version(str(venv)) == expected
+
+
+def test_zipapp_invocation_dash_p(call_zipapp, tmp_path):
+    _test_invocation_dash_p(call_zipapp, tmp_path)
+
+
+def test_wheel_invocation_dash_p(call_wheel, tmp_path):
+    _test_invocation_dash_p(call_wheel, tmp_path)
--- a/tests/test_zipapp.py
+++ b/tests/test_zipapp.py
@@ -1,3 +1,6 @@
+from __future__ import unicode_literals
+
+import json
 import os.path
 import subprocess
 import sys
@@ -11,14 +14,6 @@ HERE = os.path.dirname(os.path.dirname(__file__))
 
 def _python(v):
     return virtualenv.get_installed_pythons().get(v, "python{}".format(v))
-
-
-def _major_version(venv):
-    _, _, _, bin_dir = virtualenv.path_locations(venv)
-    exe = os.path.join(bin_dir, "python")
-    cmd = (exe, "-c", "import sys; print(sys.version_info[0])")
-    out = subprocess.check_output(cmd)
-    return int(out)
 
 
 @pytest.fixture(scope="session")
@@ -42,24 +37,17 @@ def call_zipapp(tmp_path_factory):
 @pytest.fixture(scope="session")
 def call_wheel(tmp_path_factory):
     wheels = tmp_path_factory.mktemp(basename="wheel")
-    subprocess.check_call((sys.executable, "-mpip", "wheel", "--no-deps", "-w", str(wheels), HERE))
+    subprocess.check_call((sys.executable, "-m", "pip", "wheel", "--no-deps", "-w", str(wheels), HERE))
     wheel, = wheels.iterdir()
 
     def wheel_make_env(path, python=None):
-        cmd = (sys.executable, "-mvirtualenv", "--no-download", path)
+        cmd = (sys.executable, "-m", "virtualenv", "--no-download", path)
         if python:
             cmd += ("-p", python)
         env = dict(os.environ, PYTHONPATH=str(wheel))
         subprocess.check_call(cmd, env=env)
 
     return wheel_make_env
-
-
-def _test_basic_invocation(make_env, tmp_path):
-    venv = tmp_path / "venv"
-    make_env(str(venv))
-    assert venv.exists()
-    assert _major_version(str(venv)) == sys.version_info[0]
 
 
 def test_zipapp_basic_invocation(call_zipapp, tmp_path):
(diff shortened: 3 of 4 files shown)


pfmoore:
I like this approach a lot. Distributing virtualenv as a standalone pyz file that can be run with any Python interpreter sounds like a much better approach. From what I can tell, the pyz file will still support the `-p` option, even though it's not particularly needed for a standalone pyz.

There are a lot of projects that depend on virtualenv that likely expect it to be installed (tox, pew, pipenv, ...) - otherwise, I'd be in favour of making this the official distribution method for virtualenv. It may be worth reaching out to those projects to see how hard it would be for them to switch from needing an installed virtualenv, to simply needing a virtualenv command available.

gaborbernat:
The main impediment Paul for tox e.g. is that while one can guarantee virtualenv exist via install requires, one cannot do the same for pyz. This method would only work if downstream tools themselves would be pyz, however one downside of doing so is that no direct command to the tool is available, but rather someone needs to pass the pyz to the interpreter. Therefore I would say not viable. asottile please add some tests for this so we can validate it works for all flavours of python.

pfmoore:
@gaborbernat Thanks for the clarification on tox's position. I thought something like that would be the issue. I've done some work bundling various tools, including tox and virtualenv, as zipapps using [shiv](https://pypi.org/project/shiv/). It needs a little fiddling to ensure that the unbundled zipapp is on `sys.path` for subprocess calls (by setting `PYTHONPATH`, which causes tox to issue a warning). I'd wondered whether there was a way (probably involving changes to tox and other such tools) to work around that, but I've not really had the time to look into it properly. It's very much a niche use case, but it's a pretty useful one (having Python dev tools available standalone as pyz apps is *really* convenient for certain types of workflow).

I'd still like to see the pyz form as a supported approach, even if only as an additional option.

asottile:
I've written some tests, re-enabled `install_wheel` to run with `search_dirs=None`, and fixed `-p` -- the diff is probably best viewed with [`?w=1`](https://github.com/pypa/virtualenv/pull/1287/files?w=1)

tox.ini
```diff
@@ -78,6 +78,14 @@ changedir = {toxinidir}/tasks
 extras =
 commands = python update_embedded.py
 
+[testenv:zipapp]
```
gaborbernat: definitely add it to the default targets too 👍 just to ensure the tox interface does not break

virtualenv.py
```diff
@@ -459,19 +465,35 @@ def _find_file(filename, folders):
...
 
 
-def file_search_dirs():
-    here = os.path.dirname(os.path.abspath(__file__))
-    dirs = [here, join(here, "virtualenv_support")]
-    if os.path.splitext(os.path.dirname(__file__))[0] != "virtualenv":
-        # Probably some boot script; just in case virtualenv is installed...
+@contextlib.contextmanager
+def virtualenv_support_dirs():
+    """Context manager yielding either [virtualenv_support_dir] or []"""
```
gaborbernat: maybe more accurate to say: ``locate virtualenv support files directory``

virtualenv.py
```diff
@@ -459,19 +465,35 @@ def _find_file(filename, folders):
...
+
+    # normal filesystem installation
+    if os.path.isdir(join(HERE, "virtualenv_support")):
+        yield [join(HERE, "virtualenv_support")]
+    elif IS_ZIPAPP:
+        tmpdir = tempfile.mkdtemp()
```
gaborbernat: worth writing a comment here that we provide these files on the fly, so we'll extract it and clean it up afterwards 👍

That being said seems fairly expensive to keep extracting this every time, worth investigating if:
1. Can't we provide it as a zip path?
2. Can we cache it, into e.g. userdir?

virtualenv.py
```diff
@@ -459,19 +465,35 @@ def _find_file(filename, folders):
...
-    dirs = [here, join(here, "virtualenv_support")]
-    if os.path.splitext(os.path.dirname(__file__))[0] != "virtualenv":
-        # Probably some boot script; just in case virtualenv is installed...
+@contextlib.contextmanager
+def virtualenv_support_dirs():
+    """Context manager yielding either [virtualenv_support_dir] or []"""
+
+    # normal filesystem installation
+    if os.path.isdir(join(HERE, "virtualenv_support")):
+        yield [join(HERE, "virtualenv_support")]
```
gaborbernat: note this will still provide it if virtualenv support is a file instead of a folder (change of behavior compared to beforehand), we probably should warn in such cases and break instead

virtualenv.py
```diff
@@ -459,19 +465,35 @@ def _find_file(filename, folders):
...
         try:
             # noinspection PyUnresolvedReferences
             import virtualenv
         except ImportError:
-            pass
+            yield []
         else:
-            dirs.append(os.path.join(os.path.dirname(virtualenv.__file__), "virtualenv_support"))
-    return [d for d in dirs if os.path.isdir(d)]
+            yield [join(os.path.dirname(virtualenv.__file__), "virtualenv_support")]
```
gaborbernat: again virtualenv_support is file case is not handled

virtualenv.py
```diff
@@ -900,98 +924,105 @@ def find_wheels(projects, search_dirs):
...
 def install_wheel(project_names, py_executable, search_dirs=None, download=False):
     if search_dirs is None:
-        search_dirs = file_search_dirs()
+        ctx = virtualenv_support_dirs
+    else:
 
-    wheels = find_wheels(["setuptools", "pip"], search_dirs)
-    python_path = os.pathsep.join(wheels)
+        @contextlib.contextmanager
+        def ctx():
```
gaborbernat: I prefer spelled out as ``support_files_context``

tasks/make_zipapp.py
```diff
@@ -0,0 +1,36 @@
+import argparse
```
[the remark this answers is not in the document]
asottile: further than someone would run `python3 -m pydoc zipapp`?

docs/changelog/1092.feature.rst
```diff
@@ -0,0 +1 @@
+Enable virtualenv to be distributed as a ``zipapp`` - by Anthony Sottile
```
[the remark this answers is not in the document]
asottile: yeah I was on the fence with this changelog, I didn't actually add the packaged zipapp to the repo but only enabled one to create a zipapp so I think this current changelog entry is accurate.  I can link the zipapp docs here though I really suck at rst

tasks/make_zipapp.py
```diff
@@ -0,0 +1,36 @@
...
+import io
+import os.path
+import zipapp
+import zipfile
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--root", default=".")
```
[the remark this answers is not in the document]
asottile: they're exactly the same here 🤷‍♂️

tasks/make_zipapp.py
```diff
@@ -0,0 +1,36 @@
...
+    parser.add_argument("--dest")
+    args = parser.parse_args()
+
+    if args.dest is not None:
```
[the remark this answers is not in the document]
asottile: I think it's fine as is

tasks/make_zipapp.py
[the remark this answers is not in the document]
gaborbernat: more importantly, the zipapp library is python 3 only (should we backport to 2, given that 2.6+ supports zip files)?

tasks/make_zipapp.py
```diff
@@ -0,0 +1,36 @@
...
+
+        zipf.writestr("__main__.py", "import virtualenv; virtualenv.main()")
+
+    bio.seek(0)
+    zipapp.create_archive(bio, dest)
+    print("zipapp created at {}".format(dest))
+
+
+if __name__ == "__main__":
+    exit(main())
```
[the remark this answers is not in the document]
gaborbernat: It probably would seem more same if the function would return some alternative values, just felt a bit off given it only ever returns None 🤔

tests/test_zipapp.py
```diff
@@ -0,0 +1,35 @@
...
+    return virtualenv.get_installed_pythons().get(v, "python{}".format(v))
+
+
+@pytest.fixture(scope="session")
+def zipapp(tmp_path_factory):
+    if sys.version_info[:2] == (3, 4):
+        pytest.skip("zipapp was introduced in python3.5")
+    pyz = str(tmp_path_factory.mktemp(basename="zipapp") / "virtualenv.pyz")
+    subprocess.check_call(
+        (_python("3"), os.path.join(virtualenv.HERE, "tasks/make_zipapp.py"), "--root", virtualenv.HERE, "--dest", pyz)
```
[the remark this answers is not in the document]
asottile: it was either tie the test to `virtualenv.py` location, the process `cwd`, or the test location -- I picked what I thought was the least bad -- do you have a better suggestion?

`FILE_PATH` is inaccurate, it's not actually the path to a file, but the path to the directory -- conventionally I've only seen this path referred to as `HERE`

tests/test_zipapp.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    return pyz
+
+
+def test_basic_invocation(zipapp, tmp_path):
+    venv = tmp_path / "venv"
+    subprocess.check_call((sys.executable, zipapp, str(venv)))
+    assert venv.exists()
```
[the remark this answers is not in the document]
asottile: then I have to deal with the windows-wart `Scripts` directory :(

tests/test_zipapp.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_invocation_dash_p(zipapp, tmp_path):
+    venv = tmp_path / "venv"
```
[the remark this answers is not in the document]
asottile: it just needs to ensure that the `re-exec` logic succeeds, so `-p` needs to be different than `sys.executable`

virtualenv.py
asottile: premature optimization (~650μs per call on my VM)

also think it's fairly obvious what's happening based on the code (not warranting a comment) -- let me know!

virtualenv.py
asottile: it would always crash before if that were the case so it's not worth mucking this further:

```
$ rm -rf virtualenv_support/ && touch virtualenv_support && python virtualenv.py venv
New python executable in /tmp/virtualenv/venv/bin/python
Cannot find a wheel for setuptools
Cannot find a wheel for pip
  Complete output from command /tmp/virtualenv/venv/bin/python - setuptools pip wheel:
  File "<stdin>", line 11, in <module>
ImportError: No module named pip
  File "virtualenv.py", line 2344, in <module>
    main()
  File "virtualenv.py", line 712, in main
    symlink=options.symlink)
  File "virtualenv.py", line 948, in create_environment
    download=download,
  File "virtualenv.py", line 905, in install_wheel
  File "virtualenv.py", line 796, in call_subprocess
    % (cmd_desc, proc.returncode))
OSError: Command /tmp/virtualenv/venv/bin/python - setuptools pip wheel failed with error code 1
```

virtualenv.py
asottile: it was unhandled before, it's even highlighted in this diff 😆 

```
return [d for d in dirs if os.path.isdir(d)]
```

virtualenv.py
asottile: :java:

tasks/make_zipapp.py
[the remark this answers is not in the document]
asottile: no, like the other scripts in `tasks` this is python3+.  there is no runtime support library of `zipapp` to run the resulting `.pyz`

`zipapp` is only needed to _create_ the `.pyz`, after that it works with py26+ without dependencies

tasks/make_zipapp.py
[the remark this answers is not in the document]
asottile: that's the beauty of it, `exit(None)` is the same as `exit(0)`

tests/test_zipapp.py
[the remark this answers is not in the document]
asottile: also can't use `sys.executable` in python2 because of `zipapp` -- that's why it's `_python("3")`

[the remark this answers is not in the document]
asottile: this call is `tasks/make_zipapp.py`, not the call of the zipapp itself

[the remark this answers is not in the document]
pfmoore: Oh, sorry. What you mean is if you're running virtualenv with Python 2, you need to use Python 3 to run make_zipapp? That makes sense, I guess. So if the user doesn't have Python 3 installed, this will fail. Not a real problem, as this is just a test.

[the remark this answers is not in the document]
asottile: yep!

it's also the reason I had to skip python3.4 since `python3` in that case is too old to run `zipapp`

that end-of-life is coming up soon I think 🎉

asottile:
coincidentally, this branch also fixes:

```bash
PYTHONPATH=virtualenv-16.2.0-py2.py3-none-any.whl python -m virtualenv venv
```

docs/changelog/1092.feature.rst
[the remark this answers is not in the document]
gaborbernat: It's alright, let's just add the wheel fix 👍 and while at it add a test for it 👍

tasks/make_zipapp.py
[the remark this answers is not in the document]
gaborbernat: As far as functionality yes, but I think the help message would be clearer with the os.getcwd. To be fair for that we would need to add the default formatter, can you do that? Thanks!

tasks/make_zipapp.py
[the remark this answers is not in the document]
gaborbernat: Without custom codes though (which here lack this format feels a overcomplication)

tests/test_zipapp.py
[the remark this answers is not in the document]
gaborbernat: you can just call the virtualenv path locations to find that out 👍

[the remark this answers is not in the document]
gaborbernat: just use the virtualenv path locations 👍

tests/test_zipapp.py
```diff
@@ -0,0 +1,48 @@
...
+def test_basic_invocation(zipapp, tmp_path):
+    venv = tmp_path / "venv"
+    subprocess.check_call((sys.executable, zipapp, str(venv)))
+    assert venv.exists()
+    assert _major_version(str(venv)) == sys.version_info[0]
+
+
+def test_invocation_dash_p(zipapp, tmp_path):
+    venv = tmp_path / "venv"
+    python = {2: _python("3"), 3: _python("2.7")}[sys.version_info[0]]
```
gaborbernat: Again the exists I find a too shallow test 😭

tasks/make_zipapp.py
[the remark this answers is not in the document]
asottile: this isn't a user facing script at all, I don't think this is worth doing.  None of the other scripts in `tasks` have useful defaults or help messages either

tests/test_zipapp.py
```diff
@@ -0,0 +1,48 @@
...
+    subprocess.check_call((sys.executable, zipapp, str(venv), "-p", python))
+    assert venv.exists()
```
gaborbernat: This would pass if all virtualenv would do is create this folder and then exit 🤔

tox.ini
```diff
@@ -78,6 +78,14 @@ changedir = {toxinidir}/tasks
...
+basepython = python3
+description = make a redistributable virtualenv zipapp
+skip_install = true
+changedir = {toxinidir}/tasks
+extras =
+commands = python make_zipapp.py --root ..
```
gaborbernat: we probably generate this into toxs tmp folder to not pollute the working tree

tests/test_zipapp.py
asottile: scroll down, bud <3

tox.ini
asottile: I assume the person that would be manually running this would actually want to find the file so they could upload it somewhere -- I don't expect this to be run very often

virtualenv.py
```diff
@@ -900,98 +924,105 @@ def find_wheels(projects, search_dirs):
...
-    # PIP_FIND_LINKS uses space as the path separator and thus cannot have paths
-    # with spaces in them. Convert any of those to local file:// URL form.
-    try:
-        from urlparse import urljoin
-        from urllib import pathname2url
-    except ImportError:
-        from urllib.parse import urljoin
-        from urllib.request import pathname2url
+    with search_dirs_context() as search_dirs:
+        wheels = find_wheels(["setuptools", "pip"], search_dirs)
```
gaborbernat: this generates a long diff because of the indent, what if we would just extract the indented part into its own function?

tests/test_zipapp.py
asottile: fortunately the rest of the test suite makes sure that isn't the case

virtualenv.py
asottile: This is why I suggested viewing the diff with [`?w=1`](https://github.com/pypa/virtualenv/pull/1287/files?w=1) in [this comment](https://github.com/pypa/virtualenv/pull/1287/files?w=1)

tox.ini
gaborbernat: This gets run at every commit (it's part of the default target). It's same as documentation if you want to find it just print the path after generating it (or part of the generation).

tests/test_zipapp.py
gaborbernat: it makes sure in case of the source invocation though, not the zipapp interface, not? would we zipapp a single mkdir command I think think this would still pass.

tox.ini
asottile: yeah you told me to add it to the default :P

I don't think it should be in the default, it's similar to `release` imo

virtualenv.py
gaborbernat: Not if you move it to a private function, right after this function. The private function call would be an insert and everything below would be all is same.

tests/test_zipapp.py
gaborbernat: eh, talk about a fail 😆 , probably I should fix it :-) I would like to validate a simple call with the created virtualenv ```python -c 'import sys; print(sys.executable)'``` should return the same thing we calculated with the virtualenv path.

virtualenv.py
asottile: I'd rather not add lines for no reason, what's the problem with the indent?

tests/test_zipapp.py
asottile: since I'm already better than the status quo I'm going to drop this, feel free to follow up with your more strict requirements but I don't reasonably think you should hold me to higher standards than the status quo which I'm already exceeding

gaborbernat:
Released as ``16.3.0``.
thanks

PR #1291, opened 2019-01-25
bump vendored pip to 19.0.1

gaborbernat:
Following https://github.com/pypa/pip/issues/6106 👍


tox.ini
```diff
@@ -22,13 +22,14 @@ commands = coverage run --source=virtualenv \
            !jython: coverage report --show-missing
 
 [testenv:jython]
-# pip 18.1 cache does not work with jython
-install_command = python -m pip install {opts} {packages} --no-cache
+install_command = python -m pip install {opts} {packages} --disable-pip-version-check --no-cache
+deps = pip == 18.1.0
```
gaborbernat: pip 19 uses the cache folder for to build pep-517 😢 but the pip cache msgpack is still not compatible with jython

PR #1293, opened 2019-01-27
Use importlib in python3

asottile:


virtualenv.py
```diff
@@ -1185,26 +1185,55 @@ def change_prefix(filename, dst_prefix):
...
 
 
-def copy_required_modules(dst_prefix, symlink):
-    import warnings
+if sys.version_info < (3, 4):
+    import imp
+
+    def find_module_filename(modname):
+        try:
+            f, filename, _ = imp.find_module(modname)
```
gaborbernat: I would prefer a more verbose name than f, maybe ``file_handler``?

virtualenv.py
```diff
@@ -1185,26 +1185,55 @@ def change_prefix(filename, dst_prefix):
     assert False, "Filename {} does not start with any of these prefixes: {}".format(filename, prefixes)
 
 
-def copy_required_modules(dst_prefix, symlink):
-    import warnings
+if sys.version_info < (3, 4):
```
gaborbernat: I'm not a big fan of this kind of global switch, think to put the if within the function makes it easier to read.

virtualenv.py
```diff
@@ -1185,26 +1185,55 @@ def change_prefix(filename, dst_prefix):
...
+                return None
+            else:
+                return importlib.util.spec_from_loader(modname, loader)
+
+    else:
+        find_spec = importlib.util.find_spec
+
+    def find_module_filename(modname):
+        spec = importlib.util.find_spec(modname)
+        # on pypy3, some builtin modules havae a bogus build-time file path
```
gaborbernat: a typo here

virtualenv.py
```diff
@@ -1185,26 +1185,55 @@ def change_prefix(filename, dst_prefix):
...
+        if spec is None or not os.path.exists(spec.origin):
+            return None
+        elif os.path.basename(spec.origin) == "__init__.py":
```
gaborbernat: I would prefer here to set filename as spec.origin and have an if to fix up when it's just ``__init__.py`` 🤔

virtualenv.py
gaborbernat: would prefer to have some pypy issue linked to these bogus claims 🤔

asottile:
in pypy3 this ~slightly changes which modules are symlinked -- for instance, `abc.py` gets symlinked from `lib-python`, however as far as I can tell it s unused.  `imp.find_module(...)` returns that path, but the importlib machinery returns a file path that appears to be from build-time of pypy (I think this implies, or at least it does to me, that the code for `abc.py` is baked into the interpreter?)

virtualenv.py
gaborbernat: can it be non py ending? (pyc, .class (jython), etc?)

gaborbernat:
> in pypy3 this ~slightly changes which modules are symlinked -- for instance, abc.py gets symlinked from lib-python, however as far as I can tell it s unused. imp.find_module(...) returns that path, but the importlib machinery returns a file path that appears to be from build-time of pypy (I think this implies, or at least it does to me, that the code for abc.py is baked into the interpreter?)

I agree with your intuition here. Let's open a pypy issue to clarify 100%.

virtualenv.py
asottile: this is existing code -- I just moved it

gaborbernat: Indeed, :-D if you don't mind changing it though that would be great.

virtualenv.py
asottile: unclear how jython will implement this since jython is stuck in python2.x -- should always be `.py` as far as I know as well (it's a SourceFileLoader)

virtualenv.py
asottile: I can if you want, but then it won't preserve blame:

```
$ git blame -w -M3 virtualenv.py | grep 'f, filename'
d777eb63 virtualenv.py     (Antonio Cuni           2010-06-30 14:57:52 +0200 1193)             f, filename, _ = imp.find_module(modname)
```

virtualenv.py
gaborbernat: Might be so with servers and other high-performance code that calls the function multiple times, but here we're talking a process that 99% of the time loads module, executes and finishes. I think through such global ifs make the code hard to read and make you redefine the function header multiple time, which is error-prone and extra maintenance needed. We're talking either way about an import (a dictionary lookup, once already loaded), and an if check. Worrying about this here feels very much pre-mature optimization.

virtualenv.py
gaborbernat: Does the source file loader make this guarantee in the documentation, then?

virtualenv.py
gaborbernat: I'm alright here with as we're talking about a low number of trivial lines.

virtualenv.py
asottile: I disagree, I think it's actually _less_ maintenance.  Right now when dropping an old version you simply delete the block of code -- if you intersperse it inside the function you have to perform surgery.  Also I think it makes more sense to always follow a practice instead of cherry picking on personal preference

virtualenv.py
asottile: I made https://bitbucket.org/pypy/pypy/issues/2944/origin-for-several-builtin-modules

virtualenv.py
asottile: specified here: https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec as equivalent to `__file__`

PEP 3147 says that `__file__` points to the _source file_: https://www.python.org/dev/peps/pep-3147/#file (and `__cached__` will refer to the `.pyc` file)

virtualenv.py
asottile: idk about you but I'd rather not have the deprecated lines blame to me ;)

gaborbernat: fair enough

virtualenv.py
gaborbernat: You haven't swayed me personally on this yet. Do you have any material that recommends this best practice with reasons why? Any other maintainers for virtualenv feel strongly enough to churn in on this? Otherwise I'm tempted to go with what feels easiest to me. I don't believe it's any easier, no surgery needed, you just remove an inner if branch rather than an outside one.

gaborbernat:
thank you

virtualenv.py
asottile: lol rip, they ended up blaming to me anyway :'(

gaborbernat: I shelter the blame, feel free to blame me extra. 👍

PR #1300, opened 2019-01-29
virtualenv does not *always* have a runtime dependency on setuptools

asottile:
confused

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,9 +37,6 @@ include_package_data = True
 zip_safe = True
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
 
-# setuptools needed for the entry points below to work
-install_requires = setuptools >= 18.0.0
-
 [options.extras_require]
 testing = mock;python_version<"3.3"
           pytest >= 4.0.0, <5


gaborbernat:
Eh, now seems to not be needed, but I remember vividly getting burned via the console_script generated file using some pkg_resources/packaging from setuptools.

gaborbernat:
See http://locallyoptimal.com/blog/2014/03/14/executable-python-scripts-via-entry-points/, also https://github.com/pypa/pip/issues/4702; so when does easy install script get generated over the distutils one? Because pkg_resources is what makes here setuptools a install dependency.

pfmoore:
In theory, pip should never generate easy_install style scripts. In practice however, if pip does the old-style direct `setup.py` install of a source distribution, then it's not pip that generates the wrappers but setuptools, and you get the easy_install versions of the scripts.

Roughly, this will only happen if someone installs virtualenv from source, with either a version of pip prior to 19.0, or pip >=19.0 using the `--no-use-pep517` flag or an older version of the virtualenv sources that don't have a `pyproject.toml` file. Or more succinctly, "not very often" ;-)

gaborbernat:
@pfmoore makes sense. So if I understand correctly setuptools is an install requires dependency from ``setup.py`` point of view:

- we are invoked as ``python setup.py install``,
- via the PEP-517 mechanism with a ``-no-use-pip517`` flag.

Any way to detect this from within the ``setup.py`` 🤔 For now I'm inclined to make setuptools an always dependency to cover those edge cases, but would be nice to opt out when it's not needed. What I want to avoid is someone running into the case of:

```bash
pip install setuptools
pip install virtualenv
pip uninstall setuptools # pip is fine to uninstall this, as at no point we state we depend on it post install
virtualenv # this breaks now
```

pfmoore:
> Any way to detect this from within the setup.py

Not that I know of.

> What I want to avoid is someone running into the case of:

Well, pip won't stop you uninstalling setuptools no matter what dependency information virtualenv includes, so I'd class this as simple "user error" and not worry about it. Your call, though.

gaborbernat:
I was thinking more along the lines if we don't state it the user has little chance to know about this latent dependency 🤔

gaborbernat:
Seems https://github.com/ninjaaron/fast-entry_points allows side-stepping the issue by adding it as setup requires inside the setup script.

asottile:
This seems exceedingly rare,  the only case when there'd be a runtime dependency on setuptools is when it isn't being installed from a wheel (which `virtualenv` provides and `pip>=1.5` will use by default).

Getting past that exceedingly rare situation, we come to another exceedingly rare situation:
- the user doesn't have setuptools installed => can't install from source (setuptools is invariably imported in `setup.py`)
- **OR** the user does have setuptools installed so it installs using `setup.py install` making `pkg_resources`-style entrypoints

But up until this point that's fine, the user has to have `setuptools` installed at that point because they installed the package.

The only way they can get into an error state is by another exceedingly rare interaction: uninstalling setuptools.

gaborbernat:
Maybe it's okay to drop this install requires argument, but do have some mention of this interaction in our documentation to cover our bases.

asottile:
I've never seen this happen and I use entrypoints in almost all of my projects :man_shrugging: -- I don't think it's even worth documenting

obestwalter:
I'll add myself as another random data point. I am **always** using entry points and never stumbled about this.

PR #1302, opened 2019-01-30
Link or copy PyPy headers instead of include directory as a whole

### LINKED ISSUE #1198: virtualenv_directory/include is read-only on Pypy
If I run
```
$ pypy ./virtualenv-16.0.0/virtualenv.py zzz
```
then `zzz/include` is a symlink to `/usr/lib/pypy/include` and hence read-only. This makes
```
$ pip install pybind11
```
fail, which (seemingly, legitimately, by way of the `setup(headers=....)` mechanism) wants to write to that location. I should note that on CPython, the `include` directory is not a symlink, allowing modification.

Prior bug: https://github.com/pybind/pybind11/issues/1491

daa:
This pull request changes method of virtualenv's include directory population - instead of linking Python include directory (for example ``/usr/include/python3.7m``) it links or copies distinct header files from that directory. The reason for this change is to make virtualenv'` ``include`` directory not to be symlink in case of PyPy.

Closes #1198 

- [x] added/updated test(s)

### DIFF
--- a/docs/changelog/1302.bugfix.rst
+++ b/docs/changelog/1302.bugfix.rst
@@ -1,1 +1,1 @@
-Copy or link Python header files instead of include directory itself
+Copy or link PyPy header files instead of include directory itself
--- a/docs/changelog/1302.bugfix.rst
+++ b/docs/changelog/1302.bugfix.rst
@@ -1,0 +1,1 @@
+Copy or link Python header files instead of include directory itself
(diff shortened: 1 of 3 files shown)


gaborbernat:
@daa is there any reason why this should not be a PyPy only feature? What are the performance penalties we take for this? Is it needed for CPython and Jython?

gaborbernat:
Given this, I would prefer to make it a pypy only feature. We already have tons of extra edge cases we handle for various platforms, just add an explanation of why this issue arises for pypy only. @asottile any opinions on this?

asottile:
only enabling this for pypy makes sense to me (probably in the same place where pypy creates the `include` symlink?)

Since `include` isn't created for the other platforms, things which install to `venv/include` already work

gaborbernat:
thanks

PR #1308, opened 2019-02-05
Add an option not to resolve source when symlinking.

David-5-1:
When the links are willingly made, the option allow forcing copyfile to create a symlink to the source without resolving its path.

In our setup we have a link for each major-minor python version which points to a specific build.
With current copyfile, this results in links to the sepcific build in created virtualenvs so we cannot remove any build unless when ensure all virtualenvs are no longer using it.
This MR adds a way to create symlinks to those symlinks, while keeping the default of resolving the symlinks (which was done to address issues with certain setups).

For now, my goal is mainly to get feedback on the proposed change, so I did not attempt to add documentation or edit changelog.

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -351,7 +351,7 @@ def test_relative_symlink(tmpdir):
 
 
 @pytest.mark.skipif(not hasattr(os, "symlink"), reason="requires working symlink implementation")
-def test_copyfile_from_symlink(tmp_path):
+def test_copyfile_content_with_symlink_chain_resolving_source(tmp_path):
     """Test that copyfile works correctly when the source is a symlink with a
     relative target, and a symlink to a symlink. (This can occur when creating
     an environment if Python was installed using stow or homebrew.)"""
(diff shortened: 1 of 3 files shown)


David-5-1:
In our setup we have a link for each major-minor python version which points to a specific build.
With current copyfile, this results in links to the sepcific build in created virtualenvs so we cannot remove any build unless when ensure all virtualenvs are no longer using it.
This MR adds a way to create symlinks to those symlinks, while keeping the default of resolving the symlinks (which was done to address issues with certain setups).

For now, my goal is mainly to get feedback on the proposed change, so I did not attempt to add documentation or edit changelog.

David-5-1:
My changes to tests were breaking on Windows.
As `copyfile` never creates symlinks on Windows, I disabled the added parts of the tests on Windows…

tests/test_virtualenv.py
```diff
@@ -370,9 +371,20 @@ def test_copyfile_from_symlink(tmp_path):
...
-    virtualenv.copyfile(str(src_dir / "link2"), str(copy_path))
+
+    # Check with symlink_resolve True then False
+    virtualenv.copyfile(str(src_dir / "link2"), str(copy_path), symlink_resolve=True)
+    with open(str(copy_path), "r") as f:
+        assert f.read() == "contents"
+    if not virtualenv.IS_WIN:
+        assert os.readlink(str(copy_path)) == str(src_dir / "file")
+
+    os.remove(str(copy_path))
```
gaborbernat: I think this test should be separate from this 👍 and we should leave untouched this test 👍

David-5-1: With theses changes, we get 4 tests for copyfile.
To be sure I understand, do you suggest having two test functions, one using `copyfile` with `symlink_resolve` set to `True` and one where it is `False` ?

David-5-1:
As you guessed, we are using the symlinks to builds to upgrade versions. We are providing a few different “major” releases (2.7, 3.6, 3.7) and upgrading them as new minor releases come.
From our setup, the new python minor version will be only used when we restart uWSGI service so it is as if the virtualenv was liked to a system python install which would be upgraded.
Major python upgrades would likely however break things when there are incompatible API changes (I can think of `async` and `await` that were added as keywords recently).

tests/test_virtualenv.py
gaborbernat: I would like to have these four test case as separate test functions, and please add a changelog that explains the need and addition.

gaborbernat:
I see your use case, in practice I've seen it applied more along the lines that all server packages are rebuilt on minor interpreter release. Feels a bit safer to do so.

asottile:
in my experience, patch upgrades usually break virtualenvs (for example, try making a 3.6.2 virtualenv against symlinks and then upgrading to 3.6.3) -- I'm surprised you've had success with this approach

gaborbernat:

Issue #490, opened 2013-10-30
virtualenv goes wild when python install is symlinked


rofl0r:
in my setup, python is in /opt/python/{include, lib}
and symlinked into /
like this: /bin/python -> ../opt/python/bin/python

```
Symlinking Python bootstrap modules
  Symlinking /lib/python2.7/config -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/config
  Symlinking /lib/python2.7/lib-dynload -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/lib-dynload
  Symlinking /lib/python2.7/os.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/os.py
  Ignoring built-in bootstrap module: posix
  Symlinking /lib/python2.7/posixpath.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/posixpath.py
  Cannot import bootstrap module: nt
  Symlinking /lib/python2.7/ntpath.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/ntpath.py
  Symlinking /lib/python2.7/genericpath.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/genericpath.py
  Symlinking /lib/python2.7/fnmatch.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/fnmatch.py
  Symlinking /lib/python2.7/locale.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/locale.py
  Symlinking /lib/python2.7/encodings -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/encodings
  Symlinking /lib/python2.7/codecs.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/codecs.py
  Symlinking /lib/python2.7/stat.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/stat.py
  Symlinking /lib/python2.7/UserDict.py -> /src/build/firefox/mozilla-release/firefox-build-dir/_virtualenv/lib/python2.7/UserDict.py
[...]

PR #1309, opened 2019-02-06
fix handling of relative symlinks

rofl0r:
addressing #490

this is a forward port of my patch against firefox 52esr.
i'd appreciate if someone could run tests to see if everything works as intended for other platforms (but it *should* be good)

### DIFF
--- a/docs/changelog/1309.bugfix.rst
+++ b/docs/changelog/1309.bugfix.rst
@@ -1,1 +1,0 @@
-Fix handling of relative symlinks
--- a/docs/changelog/1309.bugfix.rst
+++ b/docs/changelog/1309.bugfix.rst
@@ -1,0 +1,1 @@
+Fix handling of relative symlinks
--- a/docs/changelog/490.bugfix.rst
+++ b/docs/changelog/490.bugfix.rst
@@ -1,0 +1,1 @@
+fixes the scenario where the python base install is symlinked with relative symlinks
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -548,6 +548,33 @@ def test_create_environment_with_exec_prefix_pointing_to_prefix(tmpdir):
         os.environ[path_key] = old_path
 
 
+@pytest.mark.skipif(not hasattr(os, "symlink"), reason="requires working symlink implementation")
+def test_relative_symlink(tmp_path):
+    """Test that copyfile works correctly when the source is a symlink with a
+    relative target, and a symlink to a symlink."""
+
+    # Set up src/link2 -> ../src/link1 -> file.
+    # We will copy to a different directory, so misinterpreting either symlink
+    # will be detected.
+    src_dir = tmp_path / "src"
+    src_dir.mkdir()
+    with open(str(src_dir / "file"), "w") as f:
+        f.write("contents")
+    os.symlink("file", str(src_dir / "link1"))
+    os.symlink(str(Path("..") / "src" / "link1"), str(src_dir / "link2"))
+
+    # Check that copyfile works on link2.
+    # This may produce a symlink or a regular file depending on the platform --
+    # which doesn't matter as long as it has the right contents.
+    copy_path = tmp_path / "copy"
+    virtualenv.copyfile(str(src_dir / "link2"), str(copy_path))
+    with open(str(copy_path), "r") as f:
+        assert f.read() == "contents"
+
+    shutil.rmtree(str(src_dir))
+    os.remove(str(copy_path))
+
+
 @pytest.mark.skipif(not hasattr(sys, "real_prefix"), reason="requires running from inside virtualenv")
 def test_create_environment_from_virtual_environment(tmpdir):
     """Create virtual environment using Python from another virtual environment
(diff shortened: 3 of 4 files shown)


rofl0r:
```
An unexpected error has occurred within this region of the page.
You can try reloading this component or refreshing the entire page.
Show more info
```
could someone with a azure-pipelines-compatible browser please paste the error message from the CI check ?

rofl0r:
thank you. i force-pushed the requested changes. could you toggle another CI run ?

rofl0r:
what's wrong now ?

rofl0r:
thank you, much appreciated

gaborbernat:
it has also been published today with version 16.4.0 😀

hogepodge:
This appears to have broken building wheels for a packaging project I maintain.

http://lists.openstack.org/pipermail/openstack-discuss/2019-February/002592.html

Pinning to 16.3.0 restores previous behavior.

https://review.openstack.org/#/c/636252/

rofl0r:
from reading your report, this looks as if the change uncovered a bug in your build.
but without further details it's hard to tell what's wrong.

hogepodge:
At this point I can't say either way.

gaborbernat:
I'm tempted to agree with @rofl0r for now, due let us know once you find the root cause.

foobacca:
Here is an example of a key change in behaviour between 16.3.0 and 16.4.0.

- 16.3.0
    - create bare virtualenv
    - the files in `<venv>/lib/python2.7/encodings/` will all be real files
    - and the same for `<venv>/lib/python2.7/lib-dynload`
- 16.4.0
    - create bare virtualenv
    - the files in `<venv>/lib/python2.7/encodings/` will all be sym links - and sym links to the environment that the virtualenv executable was part of.
    - and the same for `<venv>/lib/python2.7/lib-dynload`
    
So if you:

- create virtualenv1, activate it and install the virtualenv package in it, then
- create virtualenv2

You will have sym links inside virtualenv2 pointing at files inside virtualenv1.  

For example:

```
% virtualenv /tmp/outer
% source /tmp/outer/bin/activate
[outer] % pip install virtualenv==16.3.0 
[outer] % virtualenv /tmp/inner1630
[outer] % ls -l /tmp/inner1630/lib/python2.7/encodings/aliases.py
-rw-r--r-- 1 root root 14848 Nov 12 14:31 /tmp/inner1630/lib/python2.7/encodings/aliases.py                                                                                                                    

[outer] % pip install virtualenv==16.4.0
[outer] % virtualenv /tmp/inner1640
[outer] % ls -l /tmp/inner1640/lib/python2.7/encodings/aliases.py
lrwxrwxrwx 1 user user 45 Feb 12 16:22 /tmp/inner1640/lib/python2.7/encodings/aliases.py -> /tmp/outer/lib/python2.7/encodings/aliases.py
```

- If `outer` is deleted, then `inner1640` contains broken sym links.  
- If (as in our case) `inner1640` is exported over NFS, while `outer` is unavailable over NFS, then for the NFS clients, `inner1640` contains broken sym links.
    - (I know this use case is odd - it was an accident that `outer` was still activated when `inner` was created - but it's worked for a good number of years)

If the sym link behaviour is considered desirable, would it be possible to add a command line option of `--no-sym-links` so that the virtualenv created contains **no** sym links?

hogepodge:
I verified that is indeed the key change in behavior. I believe the workaround for this is to use --always-copy when creating the virtualenv, but in my own environment we start with a vendor packaged version of virtualenv with an --always-copy bug.

In Loci, we're going to workaround the behavior change by just keeping the outer virtualenv (which we use to bootstrap pip and virtualenv to the latest versions) around to preserve the symlinks. It's a bit of a hack and not ideal, but works fine and restores our functionality.

foobacca:
> If the sym link behaviour is considered desirable, would it be possible to add a command line option of `--no-sym-links` so that the virtualenv created contains **no** sym links?

Ah, so `--always-copy` does what we want in our case.  I'll use that.

nailor:
Hello,

This change is now causing dh-virtualenv to accidentally destroy Python installation. We are still digging into what we can do on our side, but currently the behaviour looks pretty bad: https://github.com/spotify/dh-virtualenv/issues/272

gaborbernat:
We'll now revert until we figure out how to have a solution that does not break debian, an important platform.

rofl0r:
well, imo the onus is on debian's side. obviously their toolset does something in a less-than-optimal way. by simply reverting this they won't move a finger.

nailor:
@rofl0r There are actually good and perfectly valid reasons why Debian does what it does. As there are good and perfectly valid reasons for this change. 

Now, in reality, one can do all the finger pointing one wants, but one thing is very much guaranteed: Debian won't be changing that behaviour. Even if in the wildest dreams if Debian would change the way they handle symlinks in packages, that would not be rolled backwards to existing distributions. To put that into perspective, Ubuntu Bionic, i.e. 18.04, which is a long-term-support has guaranteed support until April 2023. That is a five year window for collateral breakage from this package to nuke people's system Pythons.

This unfortunately means that the only way to fix the original issue is to redesign this change so that it won't affect Debian packaging. I'm happy to help testing that revisited change, as we have already a very clean, minimal test setup to verify it!

Issue #1294, opened 2019-01-28
[1168/1285] Implement tests for prompt behavior of activate.sh


bskinn:
Per discussion at #1168 and #1285, as the first step of harmonizing the prompt handling behavior of the various supported shells, implement tests to confirm the following behaviors of `activate.sh`:

 * Obeys `VIRTUAL_ENV_DISABLE_PROMPT`
 * Implements default `__VIRTUAL_PROMPT__`, prompt prefix of "`($VIRTUAL_ENV) `", when `--prompt` is not supplied
 * Creates `__VIRTUAL_PROMPT__` prompt prefix from argument to `--prompt` *without* inserting a space

Issue #1299, opened 2019-01-28
[1168/1285] (TENTATIVE) Implement changes to prompt behavior of activate.xsh


bskinn:
Per discussion at #1168 and #1285, as part of harmonizing the prompt handling behavior of the various supported shells, changes to `activate.xsh` _may_ be warranted.

`xonsh` [has its own machinery](https://xon.sh/tutorial.html#customizing-the-prompt) for reporting the current virtual environment in the prompt (the `{env_name}` field), which is included in the default `$PROMPT` of the shell. **Of note, as currently implemented (v0.8.8) `xonsh` *always* surrounds `$VIRTUAL_ENV` with parentheses when one uses the built-in `{env_name}` prompt-variable.** Due to this machinery and its default configuration, implementation of `--prompt` and `VIRTUAL_ENV_DISABLE_PROMPT` behaviors require a rather different approach with `xonsh` than with the other shells.

By default, `xonsh`:

 * *Does* implement default `__VIRTUAL_PROMPT__` prefix of "`($VIRTUAL_ENV) `" (at least, if one uses the `$PROMPT` that ships by default with the shell).
    * Given that the formatting of the environment name is freely customizable by the user, I see no reason to forcibly re-implement this feature to *coerce* to "`($VIRTUAL_ENV) `".
 * ***DOES NOT*** do *anything at all* with the argument to `--prompt`.
    * **I favor implementation here**, as I frequently invoke as `virtualenv env --prompt="(descriptive name) "`.
 * ***DOES NOT*** obey `VIRTUAL_ENV_DISABLE_PROMPT`.
    * This feature *could* be left un-implemented, as the user can just remove `{env_name}` from their `$PROMPT` if they don't want the venv name displayed. However, I can hypothesize situations where someone would want to have `{env_name}` in `$PROMPT` in general, but be able to suppress it in certain situations. **Thus, I favor implementation**.

-----

For the `VIRTUAL_ENV_DISABLE_PROMPT` behavior, the simplest approach seems to be removing the `{env_name}` field (if present) from the current `$PROMPT`:

```
$_OLD_XONSH_PROMPT = $PROMPT
$PROMPT = re.sub(r'\{env_name(:([^{}]*\{[^}]*\})*[^{}]*)?\}', '', $PROMPT)
```

(Logic of the regex to be laid out in a following comment.)

-----

[...]

PR #1316, opened 2019-02-18
Add tests to confirm activate.sh prompt modification behaviors

bskinn:
**THIS PR WAS OBSOLETED BY #1330.**

-----

Adds the tests called for in #1294. No changes to API/features. Ping @gaborbernat for review.

No changes yet to documentation; no news item added yet -- I would think neither is necessary for this, though?

Let me know if I should squash-rebase.

- [x] added/updated test(s)
- [ ] updated/extended the documentation (???)
- [ ] added news fragment in ``docs/changelog`` folder (???)

### DIFF
--- a/tests/test_prompts.py
+++ b/tests/test_prompts.py
@@ -1,60 +1,78 @@
 """test that prompt behavior is correct in supported shells"""
 from __future__ import absolute_import, unicode_literals
 
+import os
 import subprocess
 
 import pytest
 
 import virtualenv
 
-try:
-    from pathlib import Path
-except ImportError:
-    from pathlib2 import Path
 
-CUSTOM_PROMPT = "!!!ENV!!!"
 ENV_DEFAULT = "env"
 ENV_CUSTOM = "env_custom"
 
+PREFIX_DEFAULT = "({}) ".format(ENV_DEFAULT)
+PREFIX_CUSTOM = "!!!ENV!!!"
+
+OUTPUT_FILE = "outfile"
+
+VIRTUAL_ENV_DISABLE_PROMPT = "VIRTUAL_ENV_DISABLE_PROMPT"
+VIRTUAL_ENV = "VIRTUAL_ENV"
+
 
 @pytest.fixture(scope="module")
-def env_root(tmp_path_factory):
+def tmp_root(tmp_path_factory):
     """Provide Path to root with default and custom venvs created."""
     root = tmp_path_factory.mktemp("env_root")
+    virtualenv.create_environment(str(root / ENV_DEFAULT), no_setuptools=True, no_pip=True, no_wheel=True)
     virtualenv.create_environment(
-        str(root / ENV_DEFAULT),
-        no_setuptools=True,
-        no_pip=True,
-        no_wheel=True,
-    )
-    virtualenv.create_environment(
-        str(root / ENV_CUSTOM),
-        prompt=CUSTOM_PROMPT,
-        no_setuptools=True,
-        no_pip=True,
-        no_wheel=True,
+        str(root / ENV_CUSTOM), prompt=PREFIX_CUSTOM, no_setuptools=True, no_pip=True, no_wheel=True
     )
     return root
 
 
-def test_functional_env_root(env_root):
-    assert subprocess.call(
-        'ls',
-        cwd=str(env_root),
-        shell=True,
-    ) == 0
+@pytest.fixture(scope="function")
+def clean_env():
+    """Provide a fresh copy of the shell environment."""
+    return os.environ.copy()
 
 
-def test_nonzero_exit(env_root):
-    assert subprocess.call(
-        'exit 1',
-        cwd=str(env_root),
-        shell=True,
-    ) == 1
+def fix_PS1_return(result, trim_newline=True):
+    """Replace select byte values with escape sequences."""
+    result = result.replace(b"\x1b", b"\\e")
+    result = result.replace(b"\x07", b"\\a")
+
+    # For some reason, indexing a single value into a bytes string
+    # gives the underlying integer value. b'\n' --> 10
+    if trim_newline and result[-1] == 10:
+        return result[:-1]
+    else:
+        return result
+
+
+def test_functional_env_root(tmp_root):
+    assert subprocess.call("ls", cwd=str(tmp_root), shell=True) == 0
+
+
+def test_nonzero_exit(tmp_root):
+    assert subprocess.call("exit 1", cwd=str(tmp_root), shell=True) == 1
 
 
 class TestBashPrompts:
+    @staticmethod
+    @pytest.fixture(scope="class")
+    def PS1():
+        return os.environ.get("PS1")
 
     @staticmethod
-    def test_dummy(env_root):
-        assert 1 == 1
+    def test_suppressed_prompt_default_env(tmp_root, PS1, clean_env):
+        clean_env.update({VIRTUAL_ENV_DISABLE_PROMPT: "1"})
+        command = '. {0}/bin/activate && echo "$PS1"'.format(ENV_DEFAULT)
+        result = subprocess.check_output(command, cwd=str(tmp_root), shell=True, env=clean_env)
+        result = fix_PS1_return(result)
+
+        # This assert MUST be made by encoding PS1, **NOT** by decoding
+        # result. Python's decoding machinery mangles the content of the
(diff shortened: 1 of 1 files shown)


gaborbernat:
@bskinn activator script tests are here https://github.com/pypa/virtualenv/blob/master/tests/activation/test_activation.py, any reason why this needs to be separate?

bskinn:
Mm, no, you're right - that's a much better place to put these. Should enable me to take advantage of the tooling for working with the other shells, too.  

(I had started them out in a new module so that I'd have less chance of colliding with other PRs to `master`, and so that I could smash around in the file w/o having to worry about keeping track of what was already there.)

bskinn:
I figure I should cluster the new tests for each shell together with the pre-existing tests for that shell?

I.e., bash tests [here](https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L164-L165), csh tests [here](https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L178-L179), etc.?

gaborbernat:
It's here https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L90, would prefer per feature. Once you add test for one feature, you'll need to fix issues for all interpreters. This ensures our feature parity is uniform.

bskinn:
Hm... can you restate/clarify? I'm not sure what you're asking for. 

Right now my [testing strategy](https://github.com/bskinn/virtualenv/blob/bfee6b733e6894adb24628b39ca5b7e8d580b541/tests/test_prompts.py#L66-L78) is to activate/deactivate the environment in a `subprocess` subshell, and, e.g., `echo $PS1 > out_file` at various points, then after the `.call()` is done inspect the contents of `out_file`.

If there's a different way you'd prefer me to go about it, I can reimplement... if I understand you correctly, though, you're suggesting inspecting the activate scripts themselves, rather than their effects on the prompt... and the syntaces of the various shells are different enough that it seems like that introspection would be difficult to implement in a single (parametrized?) test.

gaborbernat:
The way we test the activation scripts is already to create an env and then generate various commands: - see https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L92, your prompt tests should be part of this, e.g. by printing PS1, probably through python 👍 

https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L94

```
      self.print_os_env_var("PS1"),
```

And then make asserts on that line 👍

bskinn:
Ok, gotcha -- I'll give it a go, trying to integrate them there.

Few things:

1) One hassle that came up with `$PS1` was that some of the escape sequences got parsed oddly... this probably won't be a problem as long as they're odd-parsed the same way before and after env activation... which they should be..?
2) In the tests as they stand, are there environments created both with **and** without `--prompt` specified? When I looked around before it seemed like there was only an env created without `--prompt`
3) The different shells use at minimum different env variables to hold the prompt... bash uses `$PS1`, csh uses (IIRC) `$prompt`... and IIRC fish uses some funny functional thing, `fish_prompt()` or whatever. (And that's just the *Linux* shells...!)  I'm a bit worried that trying to implement these would be really ugly.  *Maybe* all it'll take is some expansion of the `__init__` of the `Activation` class, though?
4) We'll need to discuss further about xonsh (#1299), because it has special prompt handling stuff that I *think* will need to be tested in a fundamentally different way than the other shells.

gaborbernat:
> One hassle that came up with `$PS1` was that some of the escape sequences got parsed oddly... this probably won't be a problem as long as they're odd-parsed the same way before and after env activation... which they should be..?

you'll need to be specific, what oddities we need to pick from exactly?

> In the tests as they stand, are there environments created both with **and** without `--prompt` specified? When I looked around before it seemed like there was only an env created without `--prompt`

we could though add a test parameter to be with or without though 👍 

>  The different shells use at minimum different env variables to hold the prompt... bash uses `$PS1`, csh uses (IIRC) `$prompt`... and IIRC fish uses some funny functional thing, `fish_prompt()` or whatever. (And that's just the _Linux_ shells...!)  I'm a bit worried that trying to implement these would be really ugly.  _Maybe_ all it'll take is some expansion of the `__init__` of the `Activation` class, though?

Each shells have their own activations scripts, etc too. The environment variable would be an argument on these classes, see https://github.com/bskinn/virtualenv/blob/d5fed2304992a85b378e31b5cf26ce957b324976/tests/activation/test_activation.py#L181

> We'll need to discuss further about xonsh (#1299), because it has special prompt handling stuff that I _think_ will need to be tested in a fundamentally different way than the other shells.

Let's talk concretes so we can plan accordingly.

bskinn:
* Re escape sequence oddities, I'll just give a go at implementing the tests, and hopefully they'll just be a non-issue.

* Re `--prompt`, I think I'll have to augment the fixturing around the temporary dir, to create a new venv with `create_environment(..., prompt="foo")`... then, yeah, it should hopefully be ~as simple as an extra `parametrize`-ation on the tests.

* :+1: re env prompt variable as an init argument to the shell class

Do you want to discuss xonsh here, or under #1299?

bskinn:
Have an idea for handling xonsh. No need to discuss for now.

bskinn:
@gaborbernat I don't think I'm going to be able to get these tests to work by integrating them into the existing `test_activation.py` workflow.

The biggest problem is that the `python -c '{script}'` approach for querying environment variables &c. is not relevant to fish and posh, as those shells use shell-special functions/aliases to generate the prompt, not environment variables, and so [`print_os_env_var`](https://github.com/pypa/virtualenv/blob/751f8f92abbaa6c11ee0dbabafe8d3d64b352232/tests/activation/test_activation.py#L85-L87) won't work.  Inserting shell commands explicitly into [`commands`](https://github.com/pypa/virtualenv/blob/751f8f92abbaa6c11ee0dbabafe8d3d64b352232/tests/activation/test_activation.py#L92-L104) also didn't work, b/c for mysterious reasons the output from those commands didn't show up as lines/elements in [`out`](https://github.com/pypa/virtualenv/blob/751f8f92abbaa6c11ee0dbabafe8d3d64b352232/tests/activation/test_activation.py#L123).

FWIW, I'm pretty sure it's an additional problem that the `test_activation` tests are [run under `subprocess`'s `shell=False` default](https://github.com/pypa/virtualenv/blob/751f8f92abbaa6c11ee0dbabafe8d3d64b352232/tests/activation/test_activation.py#L120); in order for the prompts to be interrogated correctly, I *think* `shell=True` is needed.

TBH, in exploring the behavior of `test_activation.py`, I'm not at all sure that those tests are testing what they're intended to -- it seems like there's a LOT of spillover of the calling environment into what's passed to  `.check_output(...)`, that might be influencing things in funny ways.  I didn't really explore/diagnose in that direction, though.

Anyways, I now have a unified, cross-shell framework put together for these prompt behavior tests, [here](https://github.com/bskinn/virtualenv/blob/all-prompts/tests/test_prompts.py). I believe it accurately reflects the state of the activation scripts as laid out in [the tracking project on my fork](https://github.com/bskinn/virtualenv/projects/1).  `test_prompts.py` could be easily relocated into `activation/`, but I think they should be kept separate from `test_activation.py`.

What do you think?

gaborbernat:
I think the first paragraph issues you enlist problems are solvable. That being said having it separate for now is still an improvement, let's put it under activation/test_prompts.py. My biggest issue with having it separate is overhead on running the CI time wise. It already takes 13 minutes to run Jython, creating new venvs will only increase that even more. That being said we can improve on time later. Feel free to make that PR.

bskinn:
Wouldn't surprise me if there is a way to fully integrate them -- I just don't think I'm going to be able to figure it out.

I'll relocate the test module, and I should be able to piggyback the non-`--prompt` env onto `clean_python` without much trouble. I'll take care of the first and look into the second before creating the PR. 

But, I think creating a second venv for the `--prompt`-specified test cases will be unavoidable, because the [`__VIRTUAL_PROMPT__` field in the `activate` templates](https://github.com/pypa/virtualenv/blob/c37e89bb63a162e6e6d30651448a68c799d75c18/virtualenv_embedded/activate.sh#L58-L60) is replaced at venv creation-time.

PR #1285, opened 2019-01-14
Modify activate.bat to remove forced space after --prompt arg

bskinn:
Fixes #1168, per the approach suggested there (**BUT SEE BELOW**).

~~The `tasks/update_embedded.py` script has been run, updating
`virtualenv.py`.~~

~~A local test virtual environment created after implementing the change
does show the desired space between the virtualenv `--prompt`
argument and the old `$PS1` portion of the prompt.~~

Per discussion under #1168, this PR will now **REMOVE** forced spaces added after custom prompt prefixes as specified via `--prompt`. Only `activate.bat` required changes; these changes were embedded into `virtualenv.py` via `tox -e embed`.

## Thanks for contributing a pull request, see checklist all is good!

- [x] wrote descriptive pull request text
- [ ] added/updated test(s) 
- [x] updated/extended the documentation **[N/A]**
- [ ] added news fragment in ``docs/changelog`` folder

PR #1330, opened 2019-03-01
Implement prompt consistency tests and harmonize all activate scripts

### LINKED ISSUE #1168: COSMETIC: On Windows, --prompt adds a space to the resulting prompt after its argument; doesn't add space on Linux
When creating a new virtualenv with the `--prompt` option, on Windows a space is automatically added between the `--prompt=` value and the tail of the prompt:

```
C:\temp>virtualenv env --prompt="(myenv)"
...

C:\temp>env\scripts\activate

(myenv) C:\temp> 
```

On Debian stretch, a space is not added:

```
me@mycomp:~$ virtualenv env --prompt="(myenv)"
...

me@mycomp:~$ source env/bin/activate

(myenv)me@mycomp:~$ 
```

If auto-insertion of a space is desired behavior, it should be a simple fix: adding a space [here in `activate.sh`](https://github.com/pypa/virtualenv/blob/2f5c75bd4762463e904aa1c06ccba29dd817e940/virtualenv_embedded/activate.sh#L59), before the `$PS1`.

If this is desired, I can create a PR.

-----

**EDIT:** Per discussion below, the decision has been made to ***remove*** the mandatory space after the custom prompt string, for all shells.

### LINKED ISSUE #1294: [1168/1285] Implement tests for prompt behavior of activate.sh

### LINKED ISSUE #1295: [1168/1285] Implement tests for prompt behavior of activate.fish
Per discussion at #1168 and #1285, as part of harmonizing the prompt handling behavior of the various supported shells, implement tests to confirm the following behaviors of `activate.fish`:

 * Obeys `VIRTUAL_ENV_DISABLE_PROMPT`
 * Implements default `__VIRTUAL_PROMPT__`, prompt prefix of "`($VIRTUAL_ENV) `", when `--prompt` is not supplied
 * Creates `__VIRTUAL_PROMPT__` prompt prefix from argument to `--prompt` *without* inserting a space

### LINKED ISSUE #1296: [1168/1285] Implement changes to prompt behavior of activate.csh
Per discussion at #1168 and #1285, as part of harmonizing the prompt handling behavior of the various supported shells, implement changes to `activate.csh` to correct the following behaviors, with confirming tests:

 * ***Does NOT*** obey `VIRTUAL_ENV_DISABLE_PROMPT`
 * *Does* implement default `__VIRTUAL_PROMPT__` when `--prompt` is not supplied, ***BUT*** with a prefix of "`[$VIRTUAL_ENV] `"
 * *Does* create `__VIRTUAL_PROMPT__` from `--prompt`, ***BUT*** just inserts the argument to `--prompt` inside the standard prefix (becomes "`[__VIRTUAL_PROMPT__] `" instead of completely replacing the prefix):

```
$ virtualenv envtest --prompt="envtest---"
...
$ source envtest/bin/activate.csh
[envtest---] $
```

### LINKED ISSUE #1297: [1168/1285] Implement changes to prompt behavior of activate.bat
Per discussion at #1168 and #1285, as part of harmonizing the prompt handling behavior of the various supported shells, implement changes to `activate.bat` to correct the following behaviors, with confirming tests:

 * ***DOES NOT*** obey `VIRTUAL_ENV_DISABLE_PROMPT`
 * ***DOES NOT*** create `__VIRTUAL_PROMPT__` from `--prompt` *without* inserting a space (i.e., *always* inserts a space between the `--prompt` argument and the previous prompt contents)

Further, add confirming test(s) for the current correct behavior:

 * *Does* implement default `__VIRTUAL_PROMPT__` prefix of "`($VIRTUAL_ENV) `" when `--prompt` is not supplied

### LINKED ISSUE #1298: [1168/1285] Implement changes to prompt behavior of activate.ps1
Per discussion at #1168 and #1285, as part of harmonizing the prompt handling behavior of the various supported shells, implement changes to `activate.ps1` to correct the following behavior, with confirming test(s):

 * ***DOES NOT*** do *anything at all* with the argument to `--prompt`

Further, add confirming tests for the current correct behavior:

 * *Does* obey `VIRTUAL_ENV_DISABLE_PROMPT`
 * *Does* implement default `__VIRTUAL_PROMPT__`, prefix of "`($VIRTUAL_ENV) `", when `--prompt` is not supplied

### LINKED ISSUE #1299: [1168/1285] (TENTATIVE) Implement changes to prompt behavior of activate.xsh

### LINKED ISSUE #1304: [1168/1285] Fix description of VIRTUAL_ENV_DISABLE_PROMPT behavior in reference guide
The [documentation for `$VIRTUAL_ENV_DISABLE_PROMPT`](https://virtualenv.pypa.io/en/latest/reference/#envvar-VIRTUAL_ENV_DISABLE_PROMPT) currently states that:

> Any virtualenv created when this is set to a non-empty value will not have it’s activate script modify the shell prompt.

This conflicts with both the current behavior [implemented in `activate.sh`](https://github.com/pypa/virtualenv/blob/23e7f7a628b6a3801ee4764ed6ffc7e71ff038bd/virtualenv_embedded/activate.sh#L56-L64) and that described in the [user's guide](https://virtualenv.pypa.io/en/latest/userguide/#activate), where a truthy value of `$VIRTUAL_ENV_DISABLE_PROMPT` has the effect of suppressing modification of the prompt **at the time of environment *activation***.

As part of #1285, the docs should be revised here to reflect the actual behavior.

bskinn:
See history of [bskinn/virtualenv:all-prompts branch](https://github.com/bskinn/virtualenv/commits/all-prompts), tip at
bskinn/virtualenv@67cda5f, for the actual sequence of
development of the prompt tests, leading to the first commit of the PR (e6a2b99)

This PR will NOT be ready to merge until the activate scripts for csh, xonsh, cmd and posh have been fixed. PR created now per discussion at #1316, downstream of #1168 and #1285.

At time PR was opened, closes #1294 and closes #1295. By the time PR is complete, will close #1296, close #1297, close #1298, and close #1299. Also resolves #1168, the initial discussion that prompted all this.

Planning also to fix the minor documentation error to close #1304 before merging.

- [x] added/updated test(s)


setup.cfg
```diff
@@ -47,6 +47,7 @@ testing = mock;python_version<"3.3"
           pytest-xdist; platform_python_implementation!="Jython"
           pytest-localserver
           pypiserver
+          black; python_version>="3.6"
```
gaborbernat: eh, black has nothing to do with testing, use the fix_lint environment 👍

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@pytest.fixture(scope="module")
+def shell_info():
+    class ShellInfo:
+
+        execute_cmds = EchoDict({"cmd": "", "powershell": "powershell -File "})
+
+        # {shell}.script.(normal|suppress).(default|custom)[extension]
+        script_template = "{0}.script.{1}.{2}{3}"
```
gaborbernat: just use auto number?

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,261 @@
...
+            "fish": "",
+            "csh": "set prompt=%",
+            "xonsh": "$VIRTUAL_ENV = ''; $PROMPT = '{env_name}$ '",
+            "cmd": "@echo off & set PROMPT=$P$G",
+            "powershell": "",
+        }
+
+        prompt_cmds = {
+            "bash": 'echo "$PS1"',
+            "fish": "fish_prompt; echo ' '",
```
gaborbernat: not the biggest fan of this dictionaries, maybe group by shell instead and use a namedtuple 🤔

setup.cfg
bskinn: Does black auto-install as a downstream dependency somewhere? I had to install it manually.

For that matter, I had to install *tox* manually.

bskinn: Fixed in e0f7394

tests/activation/test_prompts.py
bskinn: Fixed in e0f7394

tests/activation/test_prompts.py
bskinn: Fixed in e0f7394

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,280 @@
...
+import sys
+from collections import namedtuple
+from textwrap import dedent
+
+import pytest
+
+import virtualenv
+
+try:
+    from pathlib import Path
```
gaborbernat: always use pathlib2 👍

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,280 @@
...
+SCRIPT_TEMPLATE = "{}.script.{}.{}{}"
+
+# Filename template: {shell}.out.(normal|suppress).(default|custom)
+OUTPUT_TEMPLATE = "{}.out.{}.{}"
+
+SHELL_LIST = ["bash", "fish", "csh", "xonsh", "cmd", "powershell"]
+
+
+# Py2 doesn't like unicode in the environment
+def env_compat(s):
```
gaborbernat: one character variables only within list/dict comprehension 👍

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+@pytest.fixture(scope="module")
+def platform_check_skip(tmp_path_factory):
+    """Return function triggering skip based on platform & shell."""
+    platform_incompat = "No sane provision for {} on {} yet"
+
+    def check(platform, shell):
+
+        if (sys.platform.startswith("win") and shell in ["bash", "csh", "fish"]) or (
```
gaborbernat: use the constants defined within virtualenv

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,280 @@
...
+        ):
+            pytest.skip(platform_incompat.format(shell, platform))
+
+        if sys.platform.startswith("win") and shell == "xonsh":
+            pytest.skip("Provisioning xonsh on windows is unreliable")
+
+        if shell == "xonsh" and sys.version_info < (3, 4):
+            pytest.skip("xonsh requires Python 3.4 at least")
+
+        if shell == "powershell":
```
gaborbernat: this feels like hard to maintain, branches for various shells, can we somehow group testing logic together for per shell 🤔 to avoid the need for ifs that is 👍

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,280 @@
...
+            "testscript_extension",
+            "preamble_cmd",
+            "activate_cmd",
+            "deactivate_cmd",
+        ],
+    )
+
+    # execute_cmd, prompt_cmd, activate_script are required.
+    # Defaults here are for testscript_extension, preamble_cmd, activate_cmd, and deactivate_cmd.
+    ShellInfo.__new__.__defaults__ = ("", "", "source ", "deactivate")
```
gaborbernat: is this the official way? 😄 haven't seen it yet 👍

virtualenv_embedded/activate.ps1
```diff
@@ -51,10 +51,22 @@ if (!$env:VIRTUAL_ENV_DISABLE_PROMPT)
...
+            write-host "__VIRTUAL_PROMPT__" -nonewline
+            & $function:_old_virtual_prompt
+        }
+    }
+    else
+    {
+        function global:prompt
+        {
+            # Add a prefix to the current prompt, but don't discard it.
+            write-host "($( split-path $env:VIRTUAL_ENV -leaf )) " -nonewline
```
gaborbernat: maybe we just define a variable within the if to contain the text, and have a single prompt function, not? I assume there's some way to bind in the string as an argument to the function 🤔

tests/activation/test_prompts.py
bskinn: Based on an SO post, I'm not sure there *is* an official way for <3.7.... 3.7 has a `defaults=` argument to the constructor, which AFAIK is canonical.

I can re-implement without this if desired, it'll just be more verbose with all of the replicated default values.

tests/activation/test_prompts.py
bskinn: Not sure what you mean -- this fixture is just to handle the logic of skipping certain test combinations based on platform and shell.  I think it's about as centralized as it can get, and it's not like any of the other shells care about, e.g., whether powershell has scripting permissions enabled.

It could be split up, I guess, where there's a separate function for each shell, defining its own skip conditions... folded into the `shell_info` namedtuple as `.platform_skip`, maybe?  And then the tests could call `shell_info[shell].platform_skip()`?

tests/activation/test_prompts.py
bskinn: :+1: I've been Python 3-only for all of my side projects for ~4yrs now, so I'm totally rusty on 2/3 compat!

virtualenv_embedded/activate.ps1
bskinn: Mm, make it a default value of an argument or something?

This could maybe be done for most of the shells, actually -- e.g., a new alias in the `activate.sh` that gets configured to return the appropriate prompt string, and then the `$PS1` makes use of it somehow?

That seems out of scope for this PR, though.

bskinn: Thinking about it, I think it would make more sense to do this by changing the logic in virtualenv.py itself: basically, to change all of the activate scripts to do [what __VIRTUAL_WINPROMPT__ does](https://github.com/pypa/virtualenv/blob/94c6bcb4a3f8d7ff63e2215b18d76514d147a762/virtualenv.py#L1696). Currently I think only Win cmd uses this `WINPROMPT` template substitution; the others use the [`__VIRTUAL_PROMPT__`](https://github.com/pypa/virtualenv/blob/94c6bcb4a3f8d7ff63e2215b18d76514d147a762/virtualenv.py#L1695) template substitution.

I mean, the virtualenv name/folder/etc. are hardcoded and set at the time of environment creation, so there should be little lost by also hardcoding the prompt substitution into the activate scripts.

tests/activation/test_prompts.py
bskinn: Is [this](https://github.com/bskinn/virtualenv/blob/5e5b5c17da7674f7d1e55d835baca48f2b6d3dad/tests/activation/test_prompts.py#L67-L97) more like what you had in mind, @gaborbernat?

Else, I could add a new field to [`ShellInfo`](https://github.com/bskinn/virtualenv/blob/5e5b5c17da7674f7d1e55d835baca48f2b6d3dad/tests/activation/test_prompts.py#L138-L148), say `skip_check`, and populate with functions encapsulating the skip logic per-shell.

virtualenv_embedded/activate.ps1
bskinn: This might munge the `--relocatable` capability, though..?

tests/activation/test_prompts.py
bskinn: Fixed in 5e5b5c1

tests/activation/test_prompts.py
bskinn: Believe fixed in 5e5b5c1, with switch to `virtualenv.IS_WIN`

tests/activation/test_prompts.py
bskinn: Restructured implementation in 5e5b5c1

virtualenv_embedded/activate.ps1
bskinn: I strongly prefer this be considered as out of scope for this PR. My goal with the PR is only to change and put under test the cosmetics of the prompts, and this seems like it has the potential to dig into subtle behavioral aspects of the code.

tests/activation/test_prompts.py
bskinn: Converted to fully explicit instantiation of all `ShellInfo` instances in 5e5b5c1.

conftest.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import absolute_import, unicode_literals
+
+
+def pytest_addoption(parser):
+    parser.addoption("--xonsh-prompt", action="store_true", help="Run (do not skip) xonsh prompt behavior tests")
```
gaborbernat: why shouldn't this be a flag?

tests/activation/test_prompts.py
```diff
@@ -0,0 +1,349 @@
...
+    """Check whether to skip based on platform & shell.
+
+    Returns a string if test should be skipped, or None if test should proceed.
+
+    """
+    platform_incompat = "No sane provision for {} on {} yet"
+
+    def check(platform, shell):
+
+        if shell == "bash":
```
gaborbernat: feels to me like a lot of code repetition here? Can't we make it shorter by some inheritance/class attributes as we did in the other activation test suite?

virtualenv_embedded/activate.ps1
gaborbernat: ok

conftest.py
bskinn: Clarity/explicitness, mainly.

I'm not tied to it, though -- what would you prefer?

gaborbernat: I meant can't we always run this when xonosh is installed? Why have a flag to enable/disable it?

bskinn: The prompt handling in xonsh is mostly sited *within `xonsh`*, sniffing `$VIRTUAL_ENV`, `$VIRTUAL_ENV_DISABLE_PROMPT`, and `$VIRTUAL_ENV_PROMPT`. It would be very easy for a change to be made in xonsh that starts breaking virtualenv's CI builds. 

Thus, I disabled the xonsh tests by default, to keep the CI clean and to try to minimize the codebase cross-coupling, while still making it easy to test xonsh for consistency in a local dev environment.

tests/activation/test_prompts.py
bskinn: Hmm... so, say -- folding `platform_check_skip` together with `shell_info` and refactoring into classes?

Oh, and, probably [this per-shell customization](https://github.com/pypa/virtualenv/blob/a46be0f90ccfd65aeefc8cb500851b4d6960d106/tests/activation/test_prompts.py#L338-L349) of `test_activated_prompt` could go in there too?

conftest.py
gaborbernat: I'm not comfortable with this. We either support or don't support something. If we support xonosh we should always run it to make sure it works. The solution could be to raise an issue in xonosh but we should always fun it.

tests/activation/test_prompts.py
gaborbernat: Yeah sounds good to me  👌

conftest.py
bskinn: :+1:

Totally cool by me to strip this out, and just always run the xonsh tests. I actually figured that's how it would end up. But, I wanted to make sure you were specifically aware of the potential for this kind of cross-coupled breakage.

gaborbernat: Yeah let's rip it out, 👌 I'm aware, all shells potentially breaks us (fish did recently with version 3), but we'll know as soon as the CI runs.

tests/activation/test_prompts.py
gaborbernat: Given the above two changes I'm comfortable with merging this in, thanks for the contribution! Let me know once you finish them.

bskinn: Ok, @gaborbernat -- first attempt at the refactor into the shell-info class hierarchy is implemented in 677bbc1. Thoughts please!

conftest.py
bskinn: Done in 2dcd1e2

gaborbernat:
Thanks, @bskinn, however once merged pypy3 tests started failing - https://dev.azure.com/pypa/virtualenv/_build/results?buildId=6588 can you investigate?

bskinn:
@gaborbernat I checked it...the failing test isn't related to the new prompt test machinery. Was something changed/fixed relating to pypy3 recently? I realize now, I forgot to merge/rebase master, so the merged PR may have obliterated a recent change?

bskinn:
Hm, no -- none of the recent commits on the history seem at all related to this... if anything, it looks like a behavior change in pypy3... though it's really odd timing that it would've been changed between my committing 677bbc1 and your merging the PR.

It looks like the test was expecting pypy3 to error on an attempt to import `distutils`, due to some pypy3 oddity, but it did **not** error.

```
        def _check_no_warnings(module):
            subprocess.check_call((exe, "-Werror", "-c", "import {}".format(module)))
    
        # pypy3's `distutils.sysconfig_pypy` imports `imp`
        # https://bitbucket.org/pypy/pypy/pull-requests/634/remove-unused-and-deprecated-import-of-imp/diff
        if virtualenv.IS_PYPY and sys.version_info > (3,):
            with pytest.raises(subprocess.CalledProcessError):
>               _check_no_warnings("distutils")
E               Failed: DID NOT RAISE <class 'subprocess.CalledProcessError'>
```

I don't think this is something that I'm equipped to fix -- I don't know anything about the internals of pypy3, and I have no idea if/how this `distutils` behavior affects virtualenv.

bskinn:
I looked at it some more and... Huh. That [pypy PR mentioned in the comment](https://bitbucket.org/pypy/pypy/pull-requests/634/remove-unused-and-deprecated-import-of-imp/diff) in the test function got merged.

Maybe this fixed the error in pypy3, and there's no reason any more for the [pypy3 special-case `pytest.raises(...)` check](https://github.com/pypa/virtualenv/blob/c787671ec455e3212fd78126cea558b318f48806/tests/test_cmdline.py#L40-L44)?

gaborbernat:
I believe so too 😃

PR #1333, opened 2019-03-10
Fix execsitecustomize/execusersitecustomize in site.py

zhangyangyu:
Don't exit the process when exceptions other than ImportError are
raised while trying to import sitecustomize/usercustomize.

Fixes::#1128

This is another version of the closed PR #1145.  Sorry for the delay :-(.  Update some code and add the requested test.


zhangyangyu:
Oops, pypy checks failed.  Not hard to fix but curious why for pypy `path_locations` returns a `lib_dir` that's same as `home_dir`, not even in `sys.path`? Because pypy has multiple lib dirs?

gaborbernat:

PR #1340, opened 2019-04-01
get python versions from 32&64-bit registry views

Tithen-Firion:
On Windows, get Python paths from both 32-bit and 64-bit registry views.

Additionally you get both major and minor versions, both with and without bitness information.

For example if you have versions 3.5 (32 bit), 3.5 (64 bit) and 3.6 (32 bit) you can choose from: `3.5-32`, `3.5-64`, `3.6-32`, `3.5`, `3.6`, `3-32`, `3-64`  and `3`, where:
* `3` == `3.6-32` (because it has the highest version)
* `3-32` == `3.6-32`
* `3-64` == `3.5-64`
* `3.5` == `3.5-64` (because 64-bit wins over 32-bit)
* `3.6` == `3.6-32`

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -145,7 +145,7 @@ else:
                         continue
                     # Remove bitness from version
                     if version.endswith(bitness):
-                        version = version[:-len(bitness)]
+                        version = version[: -len(bitness)]
                     exes[(version, bitness)] = join(at_path, "python.exe")
                 except WindowsError:
                     break
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -104,9 +104,11 @@ else:
 
         # Grab exes from 32-bit registry view
         exes = _get_installed_pythons_for_view("-32", winreg.KEY_WOW64_32KEY)
-        # For 64-bit OS grab exes from 64-bit registry view
-        if sys.maxsize > 2**32 or "PROGRAMFILES(X86)" in os.environ:
-            exes.update(_get_installed_pythons_for_view("-64", winreg.KEY_WOW64_64KEY))
+        # Grab exes from 64-bit registry view
+        exes_64 = _get_installed_pythons_for_view("-64", winreg.KEY_WOW64_64KEY)
+        # Check if exes are unique
+        if set(exes.values()) != set(exes_64.values()):
+            exes.update(exes_64)
 
         # Create dict with all versions found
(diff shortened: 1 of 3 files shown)


virtualenv.py
```diff
@@ -100,12 +100,35 @@ def get_installed_pythons():
         import _winreg as winreg
 
     def get_installed_pythons():
+        final_exes = dict()
+
+        # Grab exes from 32-bit registry view
+        exes = _get_installed_pythons_for_view("-32", winreg.KEY_WOW64_32KEY)
+        # For 64-bit OS grab exes from 64-bit registry view
+        if sys.maxsize > 2**32 or "PROGRAMFILES(X86)" in os.environ:
```
Tithen-Firion: This check is a bit hackish:
* `sys.maxsize > 2**32` checks if Python running is 64-bit. This can be replaced with [`platform.architecture`](1)`()[0] == "64bit"`, which one is prefered?
* `"PROGRAMFILES(X86)" in os.environ` will fail if someone creates that environment variable on 32-bit Windows (you get each version twice) or deletes it on 64-bit one (you get 32-bit versions only).

[1]: https://docs.python.org/3/library/platform.html#platform.architecture

gaborbernat: Do we need to check? Can't we just try to collect, and ignore if it fails?

Tithen-Firion: The thing is it won't fail. On 32-bit system when trying to open 64-bit registry view it will still open 32-bit view. Versions will be falsely labelled as 64-bit or worse, we will get something like `3.5-32-64`.

I think comparing values of dicts returned from `_get_installed_pythons_for_view` would work. If they are the same, skip the second dict. Would that be better?

gaborbernat: I would then go for architecture via platform 👌

Tithen-Firion: `platform.architecture()[0]` will be `32bit` when running 32-bit Python on 64-bit OS, so `"PROGRAMFILES(X86)" in os.environ` is still needed.

How about this:
```python
exes_64 = _get_installed_pythons_for_view("-64", winreg.KEY_WOW64_64KEY)
if set(exes.values()) != set(exes_64.values()):
    exes.update(exes_64)
```

gaborbernat: feels good

Issue #1339, opened 2019-03-26
virtualenv fails when run from a venv


pfmoore:
**Environment**:
Python 3.7.3, 64 bit, on Windows 7. Note this does *not* happen with Python 3.6.8, it seems to be specific to Python 3.7.

**Steps to reproduce**:
```
py -m venv xxxx
xxx\scripts\python -m pip install virtualenv
xxx\scripts\virtualenv.exe yyy
```

**Possible diagnosis**:
Virtualenv and venv use different means of locating the "base" Python interpreter (`sys.real_prefix` and `sys.base_prefix`). It's possible that these are clashing somehow.

It's also possible that some of the work around https://bugs.python.org/issue35872 might be relevant.

**Impact**:
This means that tools that manage virtualenvs, such as nox or tox, cannot be run from a Python 3 venv. The `pipx` tool does precisely this, and so `pipx` cannot be used to manage those tools if used in this environment.

**Test case output**:
```
>py -m venv xxxx
>xxxx\scripts\python -m pip install virtualenv
Collecting virtualenv
  Using cached https://files.pythonhosted.org/packages/33/5d/314c760d4204f64e4a968275182b7751bd5c3249094757b39ba987dcfb5a/virtualenv-16.4.3-py2.py3-none-any.whl
Installing collected packages: virtualenv
Successfully installed virtualenv-16.4.3
>.\xxxx\scripts\virtualenv.exe yyyy
Using base prefix 'C:\\Users\\pfmoore\\AppData\\Local\\Programs\\Python\\Python37'
New python executable in C:\Work\Scratch\n\yyyy\Scripts\python.exe
No pyvenv.cfg file
ERROR: The executable C:\Work\Scratch\n\yyyy\Scripts\python.exe is not functioning
ERROR: It thinks sys.prefix is 'c:\\work\\scratch\\n' (should be 'c:\\work\\scratch\\n\\yyyy')
ERROR: virtualenv is not compatible with this system or executable
Note: some Windows users have reported this error when they installed Python for "Only this user" or have multiple versions of Python installed. Copying the appropriate PythonXX.dll to the virtualenv Scripts/ directory may fix this problem.
```

pfmoore:
Further diagnosis - I'm not sure this is about base_prefix. It looks like Python changed the specific executable that's copied into a venv, so it's no longer the same as the executable in the base interpreter, and it appears to be what's producing the "No pyvenv.cfg file" message.

[...]

PR #1345, opened 2019-04-15
Use sys._base_executable

pfmoore:
This avoids problems with Python 3.7.3+ venv redirector on Windows. See #1339 for discussion of the issue.

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -752,12 +752,19 @@ def main():
             # 3. Special case for Python 3.7.2, where we have a redirector,
             #    but sys._base_executable does not exist.
             if sys.version_info[:3] == (3, 7, 2):
-                # We are in a venv if base_prefix != prefix
-                if sys.base_prefix != sys.prefix:
-                    # We have to guess where the Python executable is. In a
-                    # normal installation, it's in sys.prefix, so we assume that.
-                    # If that file doesn't exist, we give up and don't reinvoke.
-                    base_exe = os.path.join(sys.base_prefix, "python.exe")
+                # We are in a venv if the environment variable __PYVENV_LAUNCHER__ is set.
+                if "__PYVENV_LAUNCHER__" in os.environ:
+                    # The base environment is either sys.real_prefix (if
+                    # we were invoked from a venv built from a virtualenv) or
+                    # sys.base_prefix if real_prefix doesn't exist (a simple venv).
+                    base_prefix = getattr(sys, "real_prefix", sys.base_prefix)
+                    # We assume the Python executable is directly under the prefix
+                    # directory. The only known case where that won't be the case is
+                    # an in-place source build, which we don't support. We don't need
+                    # to consider virtuale environments (where python.exe is in "Scripts"
+                    # because we've just followed the links back to a non-virtual
+                    # environment - we hope!)
+                    base_exe = os.path.join(base_prefix, "python.exe")
                     if os.path.exists(base_exe):
                         return base_exe
         # We don't need to reinvoke
@@ -772,8 +779,8 @@ def main():
         else:
             logger.notify("Running virtualenv with interpreter {}".format(interpreter))
             env["VIRTUALENV_INTERPRETER_RUNNING"] = "true"
-            # Remove the variable __PYVENV_LAUNCHER__ as it causes the interpreter
-            # to redirect back to the virtual environment.
+            # Remove the variable __PYVENV_LAUNCHER__ if it's present, as it causes the
+            # interpreter to redirect back to the virtual environment.
             if "__PYVENV_LAUNCHER__" in env:
                 del env["__PYVENV_LAUNCHER__"]
             file = __file__
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -772,6 +772,9 @@ def main():
         else:
             logger.notify("Running virtualenv with interpreter {}".format(interpreter))
             env["VIRTUALENV_INTERPRETER_RUNNING"] = "true"
+            # Remove the variable __PYVENV_LAUNCHER__ as it causes the interpreter
+            # to redirect back to the virtual environment.
+            del env["__PYVENV_LAUNCHER__"]
             file = __file__
             if file.endswith(".pyc"):
                 file = file[:-1]
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -745,10 +745,7 @@ def main():
         if options.python and not os.environ.get("VIRTUALENV_INTERPRETER_RUNNING"):
             return options.python
         # 2. Are we running from a venv-style virtual environment with a redirector?
-        # TODO: Do we need VIRTUALENV_INTERPRETER_RUNNING?
-        # TODO: This won't work if we set _base_executable
-        # TODO: Python 3.7.2 on Windows (redirector but no _base_executable)
-        if hasattr(sys, '_base_executable'):
+        if hasattr(sys, "_base_executable"):
             return sys._base_executable
         # 3. Special case for Windows venv under Python 3.7.2, where we have
         #    a redirector but sys._base_executable does not exist.
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -744,20 +744,22 @@ def main():
         # 1. Did the user specify the --python option?
         if options.python and not os.environ.get("VIRTUALENV_INTERPRETER_RUNNING"):
             return options.python
-        # 2. Are we running from a venv-style virtual environment with a redirector?
-        if hasattr(sys, "_base_executable"):
-            return sys._base_executable
-        # 3. Special case for Windows venv under Python 3.7.2, where we have
-        #    a redirector but sys._base_executable does not exist.
-        if sys.platform == "win32" and sys.version_info[:3] == (3, 7, 2):
-            # We are in a venv if base_prefix != prefix
-            if sys.base_prefix != sys.prefix:
-                # We have to guess where the Python executable is. In a
-                # normal installation, it's in sys.prefix, so we assume that.
-                # If that file doesn't exist, we give up and don't reinvoke.
-                base_exe = os.path.join(sys.base_prefix, "python.exe")
-                if os.path.exists(base_exe):
-                    return base_exe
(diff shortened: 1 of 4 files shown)


pfmoore:
We may need to give up on Python 3.7.2, I suspect - if the test output is correct, it has the redirector that causes this issue, but no `sys._base_executable`. @zooba, is that correct? BTW, this illustrates the issue with semantic changes in point releases - it's an utter PITA to test this, as I need to keep installing different point releases (you can't have 3.7.1, 3.7.2 and 3.7.3 simultaneously installed).

pfmoore:
Tests seem OK except for Windows 3.7, which is 3.7.2, see above.

pfmoore:
The CI link seems to have disappeared from this PR. @gaborbernat do you know why that might have happened? I'm going to try closing and re-opening to see if that triggers something.

pfmoore:
Yeah, close and re-open sorted the checks issue. Weird.

@gaborbernat the only thing I can think of to do with the 3.7.2 issue is to skip the test if `sys.version` is 3.7.2, and accept that it's a core Python issue that affects that specific version only. Are you OK with that? Or do you have any suggestion of how I could handle that case better?

virtualenv.py
```diff
@@ -1459,6 +1459,10 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
     if sys.executable != py_executable:
         # FIXME: could I just hard link?
         executable = sys.executable
+        # If sys.executable is a redirector (for example, on Windows)
+        # get the real exe to copy using the semi-private sys._base_executable attribute)
+        if hasattr(sys, "_base_executable"):
```
gaborbernat:
```
executable = getattr(sys,  "_base_executable", executable)
```
Simpler?

pfmoore: Yeah - or even just `executable = getattr(sys, "_base_executable", sys.executable)` :-)

pfmoore: Regarding docs to the "semi-private thing", I don't think there are any :-( I'm going off what @zooba said over on #1339 - well, that plus the fact that there's no other way of getting the information, so private or not we don't have much choice other than to use this attribute.

I agree, though, that it would be nice if this attribute was documented.

gaborbernat: Seems this is Python 3.7+ only. Link it to https://github.com/python/cpython/pull/11745 as what implemented it.

gaborbernat: Seems to be a site.py feature, maybe we should backport it to our site.py to make it a more general feature.

pfmoore: Oh, sorry, I thought you'd followed the background in #1339. This is a new feature in Windows Python 3.7. In 3.7.2, the Python executable installed in a venv-style virtual environment changed from being a full `python.exe` to being a "redirector", that relies on being run from a venv. When virtualenv copies that redirector, it breaks. In 3.7.3, in response to some bug reports (that I haven't fully reviewed) the `sys._base_executable` attribute was added to point back to the actual exe from the source environment. That gives us a pointer to the file we need to copy.

**Unfortunately**, it looks like even this isn't sufficient - because the redirector in the intermediate venv just runs the base `python.exe`, it doesn't need the Python DLLs, and so they aren't present in the venv - so we don't copy them to the virtualenv, and as a result, `python.exe` fails because a DLL it relies on isn't present. I have no idea why this error didn't show up in my manual testing - I've just started getting errors now, even though I'm doing basically the same as I was before. Maybe I had some weird `PATH` setup that obscured the problem...

Anyway, I suspect now that this approach might be dead in the water. The whole situation is an utter mess, to be honest - and it's definitely not helped by the fact that the details aren't documented and hence aren't reliable.

I don't think it's viable to simply say that we don't support running virtualenv from within a venv - tools like pipx use venv and so running something like pew or tox from pipx will hit this limitation.

One option that *might* work to handle this is to effectively do what the `--python` option does. If we find we're running from a venv, re-invoke ourselves from the underlying "base" Python. I *think* that should be reliable, but I need to do some local tests first (as CI has Python 3.7.2, which is always going to be a problem because of the lack of `sys._base_executable`). I'll test it and update this PR once I know one way or the other.

On a somewhat related note, as you mentioned `site.py`, I'd suggest that we add support for *all* of the venv "standard" attributes (`sys.base_prefix`, `sys.base_exec_prefix`, and `sys._base_executable`) in our `site.py`, so that at least end users have a standard interface they can rely on (I don't know why venv didn't use the `sys.real_prefix` attribute that was already defined by virtualenv, but it's too late to change that now). I can prepare a separate PR to do that.

gaborbernat: Sounds good to me 👍 Thanks!

zooba: Just FYI, the reason it's `_base_executable` is because it was added in 3.7.3 and isn't available in earlier versions. For 3.8, we could promote it to `base_executable`.

It should behave the same on macOS and Windows, right? Both platforms have scenarios where you can't just copy python.exe around and expect it to work, which is why the redirectors exist on both.

pfmoore: It doesn't seem to be working the same in my tests (the OSX test was failing with what looked like an infinite recursion of running the same exe) but as I don't have access to an OSX machine, I can't do much other than fire speculative changes at CI. For now, I'm ignoring OSX, I'll come back to it later.

Do you know who is responsible for the redirector on OSX (I assume you are the person for the Windows one)? While it's low-level technical detail, I *really* think that this should be better documented, as at the moment it's a minefield for anyone trying to go beyond the "end user" level. I'm happy to write some docs, but I'd need to discuss the details with the people who know the reasoning behind what's going on.

zooba: I assume it's @ned-deily, or at least he can pass the blame along to someone else.

Unfortunately, I think if all this had been properly documented, it would simply be preventing us from fixing fatal bugs in released versions, so I'm not so keen to lock in the implementation as a public interface. It's an end user level tool, and so far only virtualenv has any need to go beyond this that I'm aware of.

pfmoore:
> Just FYI, the reason it's _base_executable is because it was added in 3.7.3

Sorry, I should have been clearer, I understand the reason for not exposing it in 3.7.3, I'm less comfortable with the switch to redirectors in a point release, and then *not* reverting that change when it was found to cause issues. But I saw the bpo discussions go past and didn't comment on them at the time, so that's as much on me as anyone. It's just a shame that leaves us with issues like https://github.com/pipxproject/pipx/issues/123 and #1339 here.

If I can get a working approach on this PR, at least for Windows (AFAIK, there was no change during 3.7.x for other platforms, so any issues there isn't new), then I'm happy that the immediate problem is dealt with (even if it has to be done by classing 3.7.2 as not fixable).

BTW, does anyone know how to find out the timescale for Azure updating their Python on Windows to 3.7.3? I'd rather not spend ages trying to get CI working here using 3.7.2 if it's going to be switching to 3.7.3 before the change gets merged (without CI, I don't see that we can reasonably support special casing for 3.7.2, conversely I'd rather not merge a 3.7.3-only fix if we don't have CI for it).

virtualenv.py
pfmoore: >  only virtualenv has any need to go beyond this that I'm aware of

Well, certainly if virtualenv can solve the "creating one virtual environment from within another" issue, then that fixes a big chunk of the problems. But there are a lot of tools out there that for better or worse introspect the layout of virtual environments<sup>1</sup>, and that means that either the implementation just gets locked in in an ad-hoc fashion<sup>2</sup> or we need to expose/document APIs that we *can* control.

For now, though, I'm going to concentrate on trying to get running virtualenv from a venv working on Windows again<sup>3</sup> and leave the wider picture for another day.

----

<sup>1</sup> For example, a *lot* of code relies on the Python binary being `Scripts/python.exe` on Windows and `bin/python` on Unix.
<sup>2</sup> Assuming no-one is willing to break a load of longstanding code on the basis of "it's not documented so even though we didn't give you access to information you needed, it's your fault".
<sup>3</sup> And probably slightly improved, with `sys.real_prefix` pointing at the original interpreter rather than the intermediate venv.

gaborbernat:
I'll make sure it passes on Mac if you get a Windows working.

pfmoore:
I think I might have found why Mac was breaking - there's another undocumented variable `__PYVENV_LAUNCHER__` that affects the core interpreter, it looks like it may force it to redirect *back* to the venv executable, which causes an infinite loop.

Just testing a fix now, and if it works I'll tidy everything up and add some documentation (regardless of what @zooba thinks, this needs writing up *somewhere* to save the next poor soul who has to work out what's going on, so if it's not going to be documented in the core, we should document it here, if only as "this is what we found out and are relying on").

pfmoore:
I think this is now good to go. I've just added some docs.

@gaborbernat @zooba could you review the code and see if there are any improvements that could be made? The tests seem to work for non-Windows platforms, I think that's because the existing support is fine for cases where there's no redirector.

pfmoore:
@gaborbernat Do you want to merge this? I can do it, but I wasn't sure I should as you'd assigned it to yourself.

pfmoore:
Excellent, thanks!

techalchemy:
Reading over this PR it is a shame how many of the same bugs were encountered here that we solved in pipenv already :|

gaborbernat:
@techalchemy I wonder why solve it in pipenv; pipenv should not fix virtualenv/venv bugs himself but rather fix it inside here 🤔

techalchemy:
definitely agree that is the preference, I think many of the fixes aren't actually virtualenv bugs but are similar to what you are encountering here and were before you planned the venv changes -- we had to tackle a lot of this ~2 years ago and already hit a number of these pain points

between a number of pieces of the ecosystem changing a lot, it became safer to just implement the fixes we needed at the time (and virtualenv wasn't being maintained yet I think)

if any of us ever has time we will likely review those changes and see if you've already implemented better fixes here

virtualenv.py
pradyunsg: > Assuming no-one is willing to break a load of longstanding code on the basis of "it's not documented so even though we didn't give you access to information you needed, it's your fault"

OT: I'm super glad we bit this bullet for pip 10.

PR #1353, opened 2019-05-04
allow specifying the seed packages

gaborbernat:
i have a dream

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ dist
 .vscode
 
 /docs/_draft.rst
+/pip-wheel-metadata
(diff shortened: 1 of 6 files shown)


gaborbernat:
Withdrawn.

sinoroc:
Am I understanding this correctly that the original intention of that change was to allow users to specify wheels (in addition to _pip_, _setuptools_, and _wheel_) that would be seeded into the virtual environment?

gaborbernat:
Yes. However, in the end, we decided against this, and now only allow specifying the exact version constraint for pip/setuptools/wheel. Note one could write a plugin for virtualenv that triggers extra seed packages post environment creation, that now can be installed via pip.

sinoroc:
That answers the question. Thanks!

PR #1354, opened 2019-05-05
drop Jython supportCI runs very long - can no longer maintain it and have very little usage

gaborbernat:


pradyunsg:
:(

Sorta sad to see this but I think this is right, given that the support adds maintainance overhead.

Thanks for all the effort you put into Jython support. I'd say if folks from the Jython community come in and want to help out with supporting Jython in virtualenv, we should be open to that opportunity.

PR #1357, opened 2019-05-15
release 16.6.0 and drop all jython reference

gaborbernat:


gaborbernat:
Done via https://dev.azure.com/pypa/virtualenv/_build/results?buildId=7719

PR #1360, opened 2019-05-23
Check for os.pathsep for path.  If we use a directory with os.pathsep…

rrauenza:
…, the virtualenv can't be added to the PATH.

- [x ] added/updated test(s)

Do we want to document this limitation somewhere?  Or just let it be an error?

- [x ] added news fragment in ``docs/changelog`` folder

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -38,6 +38,14 @@ def test_commandline_basic(tmpdir):
     _check_no_warnings("distutils")
 
 
+def test_commandline_ospathsep(tmpdir):
+    path = str(tmpdir.join("pathsepvenv" + os.pathsep + "0"))
+    assert not os.path.exists(path)
+    ret = subprocess.call([sys.executable, VIRTUALENV_SCRIPT, path])
+    assert ret != 0
+    assert not os.path.exists(path)
+
+
 def test_commandline_explicit_interp(tmpdir):
     """Specifying the Python interpreter should work"""
     subprocess.check_call([sys.executable, VIRTUALENV_SCRIPT, "-p", sys.executable, str(tmpdir.join("venv"))])
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -9,18 +9,10 @@ import pytest
 import virtualenv
 
 
-def get_src(path):
-    base, _ = os.path.splitext(path)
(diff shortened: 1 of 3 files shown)


rrauenza:
Hmm -- I think os.path.exists(path) on windows checks for ;?

gaborbernat:
maybe does multi path check , no idea, can you find it out?

rrauenza:
So it does actually check for the ; ... according to my interactive test.

```
>>> os.path.exists('c:\\temp')
True
>>> os.path.exists('c:\\temp;1')
False
>>> os.path.exists('c:\\temp;')
False
>>> os.path.exists('c:\\temp\;')
False
>>> os.path.exists('c:\\temp;')
False
>>> os.path.exists('c:\\temp')
True
>>> os.mkdir('c:\\temp\\x;')
>>> 
```

rrauenza:
It's as if the directory actually exists.  I've modified my test slightly.  I've rebased and force pushed ...  did that trigger a new CI?  I wouldn't think it would since it's a new SHA.

rrauenza:
ok, latest test shows that the exists fails before my test even begins -- the directory already exists.  Maybe from another test?

gaborbernat:
you should definitely changedir to a temp folder to perform the operation

tests/test_cmdline.py
```diff
@@ -38,6 +38,14 @@ def _check_no_warnings(module):
     _check_no_warnings("distutils")
 
 
+def test_commandline_ospathsep(tmpdir):
+    path = str(tmpdir.join("pathsepvenv" + os.pathsep + "0"))
+    assert not os.path.exists(path)
+    ret = subprocess.call([sys.executable, VIRTUALENV_SCRIPT, path])
```
gaborbernat: pass in the tmpdir as cwd

rrauenza:
Can we force the CI again? Looks like an infrastructure problem this time.

gaborbernat:
just force push ammend your last commit.

tests/test_cmdline.py
rrauenza: `test_commandline_basic` takes the `tmpdir`  and joins it with `venv` so I followed that same pattern, but have changed the name.

gaborbernat:
Thanks for your contribution!
thanks

PR #1361, opened 2019-05-28
move to src layout

gaborbernat:
hurray


pradyunsg:
:eyes: What happened here?

gaborbernat:
will be part of #1366

PR #1364, opened 2019-06-03
Further fix for #1339 (--python option)

pfmoore:


pfmoore:
Hmm, I'm getting test errors saying "ERROR: target path contains the operating system path separator ';'". That doesn't seem related to my change (and I think there was a recent change merged to do with that). So is there some other issue with master? @gaborbernat 

Having said that, there do seem to be some issues with my code - but I'm not clear what yet. I'm working on those :-)

pfmoore:
Thanks for the fast merge!

gaborbernat:
I thank you for the fast fix. Release will take a week or two as I'm in middle of multiple major changes. Will be inside release 17.0.0 though 😁

pfmoore:
No worries, it's been hitting me personally, so I was well motivated to find a fix :-) And I can use the version from github until the release arrives, so we're good there.

gaborbernat:
FYI will move to src layout soon, so use pep-517+518 to install from GitHub 😁

asottile:
Awesome, just did some debugging / spelunking to find this solves the issue I'm seeing as well in https://github.com/pre-commit/pre-commit/issues/1046

looking forward to the release!

fwiw my reproduction was a little more involved for whatever reason:

```
python -mvirtualenv venv2 -p 'C:\Users\Anthony\AppData\Local\Temp\t\x\vvv\Scripts\python.exe'
```

gaborbernat:
I'll release tomorrow as a patch release 👌

PR #1365, opened 2019-06-08
Remove warning about missing license file

### LINKED ISSUE #1352: Creating virtualenvs with PyPy always warns about a missing LICENSE file
```
⊙  pypy ~/Development/virtualenv/virtualenv.py --quiet foo                                                                     jberman@USNYHJBERMANMB2
  No LICENSE.txt / LICENSE found in source
```

https://github.com/pypa/virtualenv/blob/5e3a70c365f62cfb0f39dc4851d2104efc76c02f/virtualenv.py#L1321 looks like the problematic line, introduced in #1318.

The `license()` builtin in PyPy is static, and doesn't have a LICENSE to copy.

asottile:
Resolves #1352

Some platforms' packaging of pypy removes the license from the source directory

- [ ] added/updated test(s)

### DIFF
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -1345,7 +1345,6 @@ def copy_license(prefix, dst_prefix, lib_dir, symlink):
             dest = subst_path(license_path, prefix, dst_prefix)
             copyfile(license_path, dest, symlink)
             return
-    logger.warn("No LICENSE.txt / LICENSE found in source")
 
 
 def copy_include_dir(include_src, include_dest, symlink):


gaborbernat:
Not sure this should be a blind remove, maybe more like if the source python does not have it, don't warn in the target one.

asottile:
hmmm I just thought about this, an even better way to accomplish this would be to _just call `license()`_ on creation 🤦‍♂ -- that's wayyy simpler than trying to find and copy the file around

asottile:
actually that won't work, the `license()` builtin usually prints -- it doesn't return a string, I'll have to go back to the drawing board

blueyed:
`str(license)` appears to work?!  (pypy and pypy3 at least)

pfmoore:
Not on CPython on Windows:

```
>py -c "print(str(license))"
Type license() to see the full license text
```

pradyunsg:
Not on CPython basically.

blueyed:
Oh snap.
Maybe the IO redirection could be used there, or `str(licence)` only as a fallback?

PR #1370, opened 2019-06-21
Correct python option render

zhongjiajie:
Old content will render to `-python=2.5`
This patch correct it to `--python=2.5`

- [x] added/updated test(s): just docs changes and don't need test
- [x] added news fragment in ``docs/changelog`` folder: minor change and not nedd


zhongjiajie:
I'm new in virtualenv and I confuse why only docs change will cause `tests/activation/test_activation.py::test_xonosh` test error? Is there someone clarification

zhongjiajie:
Thanks for merging.

PR #1387, opened 2019-07-18
Fix python2 tests to use python3 to make zipapp

asottile:
looks like this was inadvertently broken in 1d4702a9


asottile:
ugh right the py launcher is gone on AP :rage:

PR #1443, opened 2019-11-14
Improve executable discovery of a PythonInfo inside a given folder

saytosid:
- Use architecture, version, implementation and platform extensions for candidates
- Cache PythonInfo.from_exe by path (resolving symlinks)
- Ensure what we discover has the same version, implementation and architecture.
- Improve our test suite for this functionality.

### DIFF
--- a/src/virtualenv/interpreters/discovery/py_info.py
+++ b/src/virtualenv/interpreters/discovery/py_info.py
@@ -120,46 +120,67 @@ class PythonInfo(object):
         if env_prefix:
             if self.real_prefix is None and self.base_executable is not None:
                 return self.base_executable
-            return self.find_exe(env_prefix)
+            return self.find_exe_based_of(inside_folder=env_prefix)
         else:
             return self.executable
 
-    def find_exe(self, home):
+    def find_exe_based_of(self, inside_folder):
         # we don't know explicitly here, do some guess work - our executable name should tell
-        exe_base_name = os.path.basename(self.executable)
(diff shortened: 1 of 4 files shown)


test/unit/interpreters/discovery/test_py_info.py
```diff
@@ -88,3 +88,7 @@ def test_satisfy_not_version(spec):
     parsed_spec = PythonSpec.from_string_spec("{}{}".format(CURRENT.implementation, spec))
     matches = CURRENT.satisfies(parsed_spec, True)
     assert matches is False
+
+
+def test_python_executable_version_number_discovery():
+    assert "python" in PythonInfo._find_possible_exe_names("python3")
```
gaborbernat: This test need to be rewritten so that it detects the failure on **all** machines not just your test machine.

src/virtualenv/interpreters/discovery/py_info.py
```diff
@@ -136,7 +137,7 @@ def _find_possible_folders(self, exe_base_name, home):
 
     @staticmethod
     def _find_possible_exe_names(exe_base_name):
-        exe_no_suffix = os.path.splitext(exe_base_name)[0]
+        exe_no_suffix = re.match(r"([a-zA-Z]+)([0-9]+)(.*)", exe_base_name).groups()[0]
```
gaborbernat: so what'll happen when this regex does not match

PR #1451, opened 2019-11-22
release 16.7.8

gaborbernat:
👍 

Release via https://dev.azure.com/pypa/virtualenv/_build/results?buildId=15674&view=results


gaborbernat:
cc @asottile @bskinn @nedbat https://pypi.org/project/virtualenv/16.7.8/

PR #1463, opened 2019-12-21
use coverage 5

gaborbernat:
This will answer questions like who covers what.


gaborbernat:
For now blocked on https://github.com/nedbat/coveragepy/issues/895

Issue #1366, opened 2019-06-10
[RFC] The next generation of virtualenv


gaborbernat:
Following the discussion in [1362](https://github.com/pypa/virtualenv/issues/1362) it became apparent that for virtualenv to be able to support the new world (Windows Store installs, venv interpreters) we need some major change of direction.

Here's my proposal for going ahead with this project. This will be a major refactor for the project, that plans to keep existing interface compatibility though.

# Project goal
- create **a fresh copy of the system (invoker) python that has its own custom, separately controllable package list**,
- an interface and behaviour that's as much as possible consistent across all supported Python environments (e.g. backport new venv features to older interpreters),
- extensible (both creation and shell activation script support wise),
- PyPi package (ability to upgrade out of band with interpreter).

# Planned features
- **universal pypi wheel**,
- **cross-platform** - Windows, all UNIX flavours, MacOS.
- **support for most supported pythons**, the initial pool of supported Python version: ``python2.7``, ``python3.4``, ``python3.5``, ``python3.5``, ``pypy3``, ``pypy2`` (hoping for IronPython and Jython at some point - any other ones we should support?)
- A **two year grace period** of support: our interface will keep the interpreter support for two more years past its EOL: creating virtual environments is so basic that this package is the last one that should drop compatibility. During this transitional two years period, we guarantee bug fixes and compatibility. New feature support is a best effort though.
- ability to **specify the target python** (we'll use PEP-514/PEP-394/explicit link to discover these versions) and create virtual environments even if that target does not have this package installed
- **prefer built-in venv**: if the target python has venv we'll create the environment using that (and then perform subsequent operations on that to facilitate other guarantees we offer)
- ability to provision **seed packages** (after the creation of the virtual environments this packages will be already installed):
  
  - we accept PEP-508 format,
  - ability to reach out to PyPi to get latest matching the spec, or offline only (such install packages must be offline)
  - by default ``pip``, ``setuptools`` and ``wheel`` are the seed packages (these packages are also automatically injected as offline wheel packages)
- **interfaces**:
   
  - CLI interface (``python -m virtualenv``, ``virtualenv``)
  - wheel interface (env ``PYTHONPATH=/t/path/to/virtualenv.whl python -m virtualenv``) - universal wheel,
  - zipapp interface,
  - API interface: ``import virtualenv; virtualenv.create("target")``
- **shell support** -  activation/deactivation scripts and prompt environment variables - by default we generate:

  - bash / zsh,
  - csh,
[...]

PR #1485, opened 2020-01-09
Fix for Python 4: use >= 3 instead of == 3

hugovk:
- [n/a] added/updated test(s)
- [n/a] updated/extended the documentation
- [n/a] added news fragment in ``docs/changelog`` folder

We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule).

When run on Python 4, this line would be skipped:

```python
    if sys.version_info[0] == 3:
        enablerlcompleter()	
```

Instead:

```python
    if sys.version_info[0] >= 3:
        enablerlcompleter()	
```

Found using https://github.com/asottile/flake8-2020.


gaborbernat:
We do, see https://www.python.org/dev/peps/pep-0596/#id6. Either way, we don't know if python 4 (if ever released) will be python 3 compatible, so what we have now is the correct thing to do.

hugovk:
Still a draft, but that 3.9 release PEP looks fairly definitive. Some PEPs refer to 3.10, others to 4, so it's not entirely clear.

I also found this, and the consensus is towards 3.10:

* https://mail.python.org/pipermail/python-committers/2018-September/006152.html

Anyway, thanks!

gaborbernat:
@hugovk side-note that we'll be soon dropping the master branch as a whole, and promoting the rewrite branch from #1366 (see https://github.com/pypa/virtualenv/milestone/7) - so would this need be fixed most likely would be on that branch either way

Issue #1332, opened 2019-03-05
virtualenv --always-copy fails on CentOS (lib64 problem?)


muxator:
Creating a virtualenv with `--always-copy` on a platform that uses `/lib64` directory seems to be impossible.

On CentOS 7.6, the same happens both with the system packages (python **2.7.5**, virtualenv **15.1.0**) and the ones installed via SCL.
There are some tickets that may be related (#985, #1192, #3477, https://github.com/FPGAwars/icestudio/issues/244), and a merged pull request that supposedly fixes the problem (#1189), yet I have the impression that there is still something wrong happening.

Example using SCL (Python **3.6.3**, virtualenv manually upgraded to **16.4.3**).
The first error seems to happen in `virtualenv_support/pip-19.0.3-py2.py3-none-any.whl`:

```
  File "<scl_base>/lib64/python3.6/socket.py", line 49, in <module>
    import _socket
ModuleNotFoundError: No module named '_socket'
```

Complete log to see it in context.

```
$ <scl_base>/bin/virtualenv --python <scl_base>/bin/python --always-copy --verbose TEST
Running virtualenv with interpreter <scl_base>/bin/python
Using base prefix '<scl_base>'
Creating <somedir>/TEST/lib/python3.6
Copying to <somedir>/TEST/lib64
Symlinking Python bootstrap modules
  Copying to <somedir>/TEST/lib/python3.6/config-3.6m-x86_64-linux-gnu
  Copying to <somedir>/TEST/lib/python3.6/lib-dynload
  Copying to <somedir>/TEST/lib64/python3.6/os.py
  Ignoring built-in bootstrap module: posix
  Copying to <somedir>/TEST/lib64/python3.6/posixpath.py
  Cannot import bootstrap module: nt
  Copying to <somedir>/TEST/lib64/python3.6/ntpath.py
  Copying to <somedir>/TEST/lib64/python3.6/genericpath.py
  Copying to <somedir>/TEST/lib64/python3.6/fnmatch.py
  Copying to <somedir>/TEST/lib64/python3.6/locale.py
  Copying to <somedir>/TEST/lib64/python3.6/encodings
  Copying to <somedir>/TEST/lib64/python3.6/codecs.py
  Copying to <somedir>/TEST/lib64/python3.6/stat.py
  Cannot import bootstrap module: UserDict
  Creating parent directories for <somedir>/TEST/lib64/python3.6/lib-dynload
[...]

PR #1500, opened 2020-01-17
CentOs and Fedora support

gaborbernat:
Instead of hard coding patterns and guessing let's ask the host python via the sysconfig (for stdlib) and distutils install (for virtualenv elements - we target pip installable envs here).

Resolves https://github.com/pypa/virtualenv/issues/1332

TODO:
[ ] validate again with Fedora
[ ] validate again with CentOS
[ ] add test for --copies

### DIFF
--- a/src/virtualenv/activation/python/__init__.py
+++ b/src/virtualenv/activation/python/__init__.py
@@ -14,6 +14,6 @@ class PythonActivator(ViaTemplateActivator):
 
     def replacements(self, creator, dest_folder):
         replacements = super(PythonActivator, self).replacements(creator, dest_folder)
-        site_dump = json.dumps([os.path.relpath(str(i), str(dest_folder)) for i in creator.site_packages], indent=2)
+        site_dump = json.dumps(list({os.path.relpath(str(i), str(dest_folder)) for i in creator.libs}), indent=2)
         replacements.update({"__SITE_PACKAGES__": site_dump})
         return replacements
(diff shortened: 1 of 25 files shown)


ssbarnea:
I am extreamly happy to see this ticket raised. I can provide help on achieving this as I had to do the same distro-testing with other python packages, for the same scope: to test that we don't break on these.

gaborbernat:
@ssbarnea thank you very much. I'll merge this PR, I've tested on Fedora 29 and CentOS 7 and for myself, the test suite passed with flying colours. Feel free to take the code out for a sping yourself and let me know if you run into any issues. 

There exists one slight discrepancy between what the builtin venv module does and what we do. The platform library folder (``lib64``) symlinks to the pure library folder (``lib``). I prefer to not do but instead create two folder hierarchies. For all my tests this did not seem to matter, and both of those are added to the ``sys.path``. Until someone can point me why the symlink should be preferable I'll leave with this approach. On CentOS I've only found ``python2.7`` and ``python3.6``, so I've tested only those two (also ran ``python3.8`` but I've built this from source, so I consider it not an exact match, as the upstream binary patches are missing). Any way to get the other pythons?

CC @hroncok you might be interested in this.

hroncok:
> I've tested on Fedora 29

I can test this somehow on Fedora 31 if needed. 29 is out of support. But our patches are the same regarding lib64.

Not creating the symbolic link is probably OK. cc @encukou @stratakis

hroncok:
> Any way to get the other pythons?

On CentOS 7, I only know about 2.7, 3.4 (in EPEL) and 3.6.

On recent Fedoras, there should be anything from 2.6, 2.7, 3.4-3.9 and pypy2+3.

gaborbernat:
Yeah on fedora I could get all those easily and the test suite passed 😎

Issue #1448, opened 2019-11-18
support building of C extensions - rewrite


gaborbernat:
With the rewrite branch, let's ensure https://github.com/pypa/virtualenv/pull/1023 is supported. The hunch is we **might** need the header files but we need to validate.

gaborbernat:
As the test library, we probably can use https://github.com/sumerc/yappi 👍

gaborbernat:
Where I've got is that pip will inject the virtual environment include folder onto the include path, however, can there be any content there? Standard library headers are already added as part of the standard library path; so the question feels to me goes towards do any package install headers that would as such not be available globally, but only within a virtual environment?

```
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/private/var/folders/kt/btg285ds5kx4l1k398lb7df80000gr/T/pytest-of-bgabor8/pytest-228/test_can_build_c_extensions_cp0/include -I/Users/bgabor8/.pyenv/versions/3.8.1/include/python3.8 -c yappi/_yappi.c -o build/temp.macosx-10.14-x86_64-3.8/yappi/_yappi.o
```

gaborbernat:
Now works.

PR #1503, opened 2020-01-21
support for c-extension builds within virtualenv

gaborbernat:
- add test to check if it works

Resolves https://github.com/pypa/virtualenv/issues/1448


src/virtualenv/activation/python/activate_this.py
```diff
@@ -5,51 +5,28 @@
...
-    path = os.path.realpath(os.path.join(os.path.dirname(__file__), site_package))
-    if sys.version_info[0] == 2:
-        path = path.encode(sys.getfilesystemencoding())
-    site.addsitedir(path)
+os.environ["PATH"] = os.pathsep.join([bin_dir] + os.environ.get("PATH", "").split(os.pathsep))
+os.environ["VIRTUAL_ENV"] = base  # virtual env is right above bin directory
 
+# add the virtual environments libraries to the host python import mechanism
+prev_length = len(sys.path)
+for lib in "__LIB_FOLDERS__".split(os.pathsep):
```
kunesj, line 26: This is not correct. It will always cycle through `['__LIB_FOLDERS__', ]`.

src/virtualenv/activation/python/activate_this.py
```diff
@@ -5,51 +5,28 @@
...
+    path = os.path.realpath(os.path.join(bin_dir, lib))
+    site.addsitedir(path.decode("utf-8") if "__DECODE_PATH__" else path)
```
kunesj, line 28: This is not correct. It will always try to decode.

gaborbernat: why will it always try to decode?

src/virtualenv/activation/python/activate_this.py
gaborbernat: Why?

kunesj: Because you are splitting `"__LIB_FOLDERS__"` which is a string not a variable. 
You probably wanted to use something like `os.environ["__LIB_FOLDERS__"]` instead?

src/virtualenv/activation/python/activate_this.py
kunesj: Because strings evaluate to `True`. 
You probably wanted to use something like `os.environ.get("__DECODE_PATH__", "")` instead of just `"__DECODE_PATH__"`?

I think I saw similar errors in more places, but these were just the ones that caused this script to fail when I tried to use it.

src/virtualenv/activation/python/activate_this.py
gaborbernat: You're incorrect on this; the file is a template not the actual file - https://github.com/pypa/virtualenv/blob/master/src/virtualenv/activation/python/__init__.py#L17-L25

src/virtualenv/activation/python/activate_this.py
gaborbernat: same as above - things get replaced per https://github.com/pypa/virtualenv/blob/master/src/virtualenv/activation/python/__init__.py#L17-L25

src/virtualenv/activation/python/activate_this.py
kunesj: Huh? OK, now I feel stupid.
Sorry for bothering you. :bow:

PR #1519, opened 2020-01-29
rewrite the documentation

### LINKED ISSUE #1465: rewrite documentation
With the rewrite, we need to update the documentation.

gaborbernat:
Resolves #1465.


LICENSE
```diff
@@ -1,6 +1,4 @@
-Copyright (c) 2007 Ian Bicking and Contributors
-Copyright (c) 2009 Ian Bicking, The Open Planning Project
-Copyright (c) 2011-2016 The virtualenv developers
+Copyright (c) 2020-202x The virtualenv developersÂ
```
pradyunsg: Note the unicode character at the end

pradyunsg:
There seem to be a whole bunch of not-documentation changes here -- is that intentional?

gaborbernat:
@pradyunsg moved those out now to #1535, #1536 and #1534

pradyunsg:
FYI -- the ReadTheDocs build is failing due to:

```
Problem in your project's configuration. Invalid "sphinx.builder": .readthedocs.yml: Your project is configured as "Sphinx HtmlDir" in your admin dashboard, but your "sphinx.builder" key does not match.
```

gaborbernat:
Done, documentation is now ready under https://external-builds.readthedocs.io/html/virtualenv/1519/index.html

This is now good enough for the first iteration, please feel free to comment, I'll merge later today with potential feedback.

docs/_static/custom.css
```diff
@@ -0,0 +1,60 @@
+.wy-nav-content {
+    max-width: none;
```
pradyunsg: Please don't do this. See https://github.com/python/python-docs-theme/issues/4. We ended up with upstream Sphinx adding a limit to width of content, because it's a much better UX.

gaborbernat: I feel very strongly about this, and I don't agree that limiting the width of the content is better UX... I mean you can always resize your browser if you want smaller, and especially for tables, that size feels very limiting.

gaborbernat: I'll compromise on ``1080px``; 🤷‍♂

pradyunsg: 👀

Can we limit it to somewhere ~120 characters per line?

gaborbernat: That's roughly 1080px 👍

pradyunsg: We chatted out of band, and @gaborbernat has strong opinions and does not wish to stick to the defaults of the theme here.

I'm backing out of this theme-related discussion, because I've had enough of those already (see the aforementioned issue related to docs.python.org) and  don't have the mental bandwidth want to have another discussion pushing for  what is already a norm in most documentation / web interfaces for good reason.

pradyunsg: Also, I've suggested that he remove the 1080px limit, since that's not useful (it's >150 chars vs the ~100 limit suggested in most places) and isn't a viable compromise IMO.

pradyunsg: Screenshot 2020-02-04 at 6 20 35 PM

Attaching a screenshot of the current theme, with the 1080px limit applied, on a full-screen Firefox window on my laptop's screen.

gaborbernat: On my screen, the change is a bit more dramatic 🤔 
image

pradyunsg: 4K vs not-4K :)

gaborbernat: Slightly reduced width to fit in 120 characters.

gaborbernat:
I'll merge so we can release the next beta, but we can do fixes in follow-up PRs.

Issue #1515, opened 2020-01-28
20b1: "RuntimeError: failed to detect cpython3.6.8-64|..."


asottile:
this is breaking some tests for [pre-commit](https://pre-commit.com)

this is similar to #1514 but involves absolute paths

here's a minimal reproduction:

```bash
set -euxo pipefail
rm -rf venv1 venv2
virtualenv venv1
venv1/bin/pip install virtualenv --upgrade --pre
"$PWD/venv1/bin/python" -mvirtualenv venv2 -p "$PWD/venv1/bin/python"
```

```console
$ bash t.sh 
+ rm -rf venv1 venv2
+ virtualenv venv1
Using real prefix '/usr'
Path not in prefix '/home/asottile/opt/venv/include/python3.6m' '/usr'
New python executable in /home/asottile/workspace/pre-commit/venv1/bin/python3
Also creating executable in /home/asottile/workspace/pre-commit/venv1/bin/python
Installing setuptools, pip, wheel...
done.
+ venv1/bin/pip install virtualenv --upgrade --pre
Collecting virtualenv
  Using cached virtualenv-20.0.0b1-py2.py3-none-any.whl (4.6 MB)
Collecting appdirs<2,>=1.4.3
  Using cached appdirs-1.4.3-py2.py3-none-any.whl (12 kB)
Collecting filelock<4,>=3.0.0
  Using cached filelock-3.0.12-py3-none-any.whl (7.6 kB)
Collecting importlib-metadata<2,>=0.12; python_version < "3.8"
  Using cached importlib_metadata-1.4.0-py2.py3-none-any.whl (29 kB)
Collecting importlib-resources<2,>=1.0; python_version < "3.7"
  Using cached importlib_resources-1.0.2-py2.py3-none-any.whl (32 kB)
Collecting six<2,>=1.12.0
  Using cached six-1.14.0-py2.py3-none-any.whl (10 kB)
Collecting zipp>=0.5
  Using cached zipp-2.1.0-py3-none-any.whl (4.6 kB)
Installing collected packages: appdirs, filelock, zipp, importlib-metadata, importlib-resources, six, virtualenv
Successfully installed appdirs-1.4.3 filelock-3.0.12 importlib-metadata-1.4.0 importlib-resources-1.0.2 six-1.14.0 virtualenv-20.0.0b1 zipp-2.1.0
[...]

PR #1522, opened 2020-01-29
Better handling of base executable not found #1515

gaborbernat:
Improve the base executable discovery mechanism:

- print at debug level why we refuse some candidates,
- when no candidates match exactly, instead of hard failing fallback to the closest match where the priority of matching attributes is python implementation, major version, minor version, architecture, patch version, release level and serial (this is to facilitate things to still work when the OS upgrade replace/upgrades the system python with a newer version than what the virtualenv host python was created with),
- always resolve system_executable information during the interpreter discovery, and the discovered environment is the system interpreter instead of the venv/virtualenv (this happened before lazily the first time we accessed and caused reporting that the created virtual environment is of type of the virtualenv host python version, instead of the system pythons version - these two can differ if the OS upgraded the system python underneath and the virtualenv host was created via copy),

Resolves https://github.com/pypa/virtualenv/issues/1515.


src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -26,7 +26,7 @@ def load_host_site():
     # because this is what pip will be using; the host site.py though may contain it's own pattern for where the
     # platform and pure library paths should exist
 
-    # notably on Ubuntu there's a patch for getsitepackages to point to
+    # notably on Ubuntu there's a micro for getsitepackages to point to
```
asottile: oops

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -92,7 +92,7 @@ def rewrite_standard_library_sys_path():
 
 def disable_user_site_package():
     """Flip the switch on enable user site package"""
-    # sys.flags is a c-extension type, so we cannot monkey patch it, replace it with a python class to flip it
+    # sys.flags is a c-extension type, so we cannot monkey micro it, replace it with a python class to flip it
```
asottile: ""

Issue #1516, opened 2020-01-29
20b1: virtualenv creation fails in parallel: "AttributeError: 'NoneType' object has no attribute 'lock_file'"


asottile:
```bash
set -euxo pipefail
rm -rf venv1 venv{2..10} log{2..10}
virtualenv venv1
venv1/bin/pip install git+https://github.com/pypa/virtualenv@rewrite
seq 2 10 | xargs -P8 --replace bash -c 'venv1/bin/virtualenv venv{} -p /usr/bin/python3.6 -vv >& log{} && rm log{}'
```

I expect this to block for a reasonable amount of time (maybe max 60s) and succeed in creating the virtualenvs, the logging around locking appears to be broken

```console
$ bash t.sh 
+ rm -rf venv1 venv2 venv3 venv4 venv5 venv6 venv7 venv8 venv9 venv10 log2 log3 log4 log5 log6 log7 log8 log9 log10
+ virtualenv venv1
Using real prefix '/usr'
Path not in prefix '/home/asottile/opt/venv/include/python3.6m' '/usr'
New python executable in /home/asottile/workspace/pre-commit/venv1/bin/python3
Also creating executable in /home/asottile/workspace/pre-commit/venv1/bin/python
Installing setuptools, pip, wheel...
done.
+ venv1/bin/pip install git+https://github.com/pypa/virtualenv@rewrite
Collecting git+https://github.com/pypa/virtualenv@rewrite
  Cloning https://github.com/pypa/virtualenv (to revision rewrite) to /tmp/pip-req-build-_zc0pdot
  Running command git clone -q https://github.com/pypa/virtualenv /tmp/pip-req-build-_zc0pdot
  Running command git checkout -b rewrite --track origin/rewrite
  Switched to a new branch 'rewrite'
  Branch 'rewrite' set up to track remote branch 'rewrite' from 'origin'.
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting importlib-resources<2,>=1.0; python_version < "3.7"
  Using cached importlib_resources-1.0.2-py2.py3-none-any.whl (32 kB)
Collecting filelock<4,>=3.0.0
  Using cached filelock-3.0.12-py3-none-any.whl (7.6 kB)
Collecting appdirs<2,>=1.4.3
  Using cached appdirs-1.4.3-py2.py3-none-any.whl (12 kB)
Collecting six<2,>=1.12.0
[...]

PR #1524, opened 2020-01-29
Fix failing to acquire lock in case of parallel runs

gaborbernat:
Resolves https://github.com/pypa/virtualenv/issues/1516

### DIFF
--- a/src/virtualenv/util/lock.py
+++ b/src/virtualenv/util/lock.py
@@ -2,12 +2,13 @@
 from __future__ import absolute_import, unicode_literals
 
 import logging
+import os
 from contextlib import contextmanager
 from threading import Lock
 
 from filelock import FileLock, Timeout
 
-from virtualenv.util.path import Path, ensure_dir
+from virtualenv.util.path import Path
 
 
 class _CountedFileLock(FileLock):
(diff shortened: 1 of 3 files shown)


src/virtualenv/util/lock.py
```diff
@@ -72,11 +73,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
     def _lock_file(self, lock):
-        ensure_dir(self.path)
+        # multiple processes might be trying to get a first lock... so we cannot check if this directory exist without
+        # a lock, but that lock might then become expensive, and it's not clear where that lock should live.
+        # Instead here we just ignore if we fail to create the directory.
+        try:
+            os.makedirs(str(self.path))
+        except OSError:
+            pass
```
asottile: this should only catch the exception if it's EEXIST and it's a directory I believe -- other OSErrors should raise here (in py3+ `os.makedirs(..., exist_ok=True)`)

gaborbernat: not too worried about this, as if this fails for any other reason than EEXIST the next line will also fail, and done it like this to handle both platforms 🤔 I know about exists_ok=True... but that does not work on python 2...

PR #1529, opened 2020-01-30
fish prompt: preserve pipestatus

krobelus:
The upcoming release fish 3.1.0 provides a variable $pipestatus, that
exposes the exit code of each process in a pipeline. This $pipestatus
is also used by the new default prompt.

Presently, $status is restored but not $pipestatus, so a prompt displaying
the pipestatus is wrong:

expected:

	(env) $ false | true | false
	(env) [1|0|1] $

actual:

	(env) $ false | true | false
	(env) [0|1] $

The wrong $pipestatus is because `echo 'exit 1' | source` is used
to restore the $status.
This commit solves this problem more elegantly by running the user's
prompt immediately, and printing it later.
Uses the fish builtin "string" command which exists since fish 2.3b1
(released April 19, 2016) so that's unlikely to cause problems.

- [ ] added/updated test(s)


gaborbernat:
This should be done against the rewrite branch... The master branch soon goes to legacy.

krobelus:
Whoops, that blew up when I changed the base branch.
Reopened as #1530.

PR #1547, opened 2020-02-10
Add link to virtualenv < 20.0.0 changelog.

jezdez:

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -87,4 +87,5 @@ v20.0.0b1 (2020-01-28)
 
    The current virtualenv is the second iteration of implementation. From version ``0.8`` all the way to ``16.7.9``
    we numbered the first iteration. Version ``20.0.0b1`` is a complete rewrite of the package, and as such this release
-   history starts from there. The old changelog is still available in the `virtualenv < 20.0.0 documentation <https://virtualenv.pypa.io/en/16.7.9/changes.html>`_.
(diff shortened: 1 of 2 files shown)


docs/changelog.rst
```diff
@@ -73,4 +73,4 @@ v20.0.0b1 (2020-01-28)
 
    The current virtualenv is the second iteration of implementation. From version ``0.8`` all the way to ``16.7.9``
    we numbered the first iteration. Version ``20.0.0b1`` is a complete rewrite of the package, and as such this release
-   history starts from there.
+   history starts from there. The old changelog is still available in the `virtualenv < 20.0.0 documentation <https://virtualenv.pypa.io/en/16.7.9/changes.html>`_.
```
gaborbernat: Let's use the legacy name, see https://virtualenv.pypa.io/en/legacy/

pradyunsg:
```suggestion
   history starts from there. The old changelog is still available in virtualenv's `legacy documentation <https://virtualenv.pypa.io/en/legacy/changes.html>`_.
```

gaborbernat: seems I'm not allowed to push to @jezdez repository, so can't accept suggested changes

jezdez: Good idea!

jezdez: Woops, sorry, only seeing this now :)

PR #1550, opened 2020-02-10
fix system executable discovery

### LINKED ISSUE #1552: virtual environments created via relative paths on Windows creates bad console executables
Seems https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve does not generate the absolute path on Windows, but does everywhere else.

### LINKED ISSUE #1553: incorrect discovery for system executable
Seems sometimes venvs created set their base executable to themselves... In this case, we accept this without question, so we threat virtual environments as system pythons.

gaborbernat:
This might fix our test suite now.

Resolves #1552
Resolves #1553

### DIFF
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -22,7 +22,7 @@ from virtualenv.util.subprocess import run_cmd
 from virtualenv.util.zipapp import ensure_file_on_disk
 from virtualenv.version import __version__
 
-HERE = Path(__file__).absolute().parent
+HERE = Path(os.path.abspath(__file__)).parent
 DEBUG_SCRIPT = HERE / "debug.py"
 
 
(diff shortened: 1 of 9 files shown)


src/virtualenv/create/creator.py
```diff
@@ -131,7 +131,7 @@ def non_write_able(dest, value):
             # pre 3.6 resolve is always strict, aka must exists, sidestep by using os.path operation
             dest = Path(os.path.realpath(raw_value))
         else:
-            dest = value.resolve()
+            dest = value.absolute().resolve()  # on Windows absolute does not imply resolve so use both
```
pradyunsg: `absolute()` is undocumented (see [bpo-29688](https://bugs.python.org/issue29688))

pradyunsg: Also relevant: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573

gaborbernat: It works in practice though, so I'm fine with this approach for now 🤔

gaborbernat: ugh, reading up that thread I've swapped to os.path.abspath now 🤷‍♂

pfmoore: Also relevant is [bpo-39090](https://bugs.python.org/issue39090) - "getting the absolute path is surprisingly subtle and varied depending on your needs" 🙁

Issue #1556, opened 2020-02-10
`virtualenv` should print help and return 2


nsoranzo:
or anything greater than 0.

Behaviour pre v20:
```console
(virtualenv16.7.9_venv) $ virtualenv --version
16.7.9
(virtualenv16.7.9_venv) $ virtualenv 
You must provide a DEST_DIR
Usage: virtualenv [OPTIONS] DEST_DIR

Options:
  --version             show program's version number and exit
...
(virtualenv16.7.9_venv) $ echo $?
2
```

New behaviour in v20.0.1:
```console
(virtualenv20.0.1_venv) $ virtualenv --version
virtualenv 20.0.1 from /tmp/virtualenv20.0.1_venv/local/lib/python2.7/site-packages/virtualenv/__init__.pyc
(virtualenv20.0.1_venv) $ virtualenv 
(virtualenv20.0.1_venv) $ echo $?
0
```

gaborbernat:
The help is printed with ``--help\-h``. All values have defaults so there's no reason why not just execute the create virtual environment command.

nsoranzo:
Ah, it was not clear at all to me that it was creating a virtualenv in `venv`, closing in favour of #1557.

gaborbernat:
Hello, a fix for this issue has been released via virtualenv 20.0.2; see https://pypi.org/project/virtualenv/20.0.2/ (https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-2-2020-02-11) . Please give a try and report back if your issue has not been addressed; if not, please comment here, and we'll reopen the ticket. We want to apologize for the inconvenience this has caused you and say thanks for having patience while we resolve the unexpected bugs with this new major release.

![thanks](https://media.tenor.co/images/1114067881329bb4218880eeb7ec34fd/tenor.gif)

PR #1568, opened 2020-02-11
require passing a destination directory

pradyunsg:
Resolves https://github.com/pypa/virtualenv/issues/1556


pradyunsg:
Note to self: need to update documentation, since there's at least one section w/ prose that suggests passing no-args will work.

nsoranzo:
I think the description should say: "Resolves https://github.com/pypa/virtualenv/issues/1556"

Issue #1557, opened 2020-02-10
Creating a virtualenv gives no feedback to user


nsoranzo:
Behaviour pre v20:
```console
(virtualenv16.7.9_venv) $ virtualenv --version
16.7.9
(virtualenv16.7.9_venv) $ virtualenv test_venv
Using real prefix '/usr'
New python executable in /tmp/test_venv/bin/python2
Also creating executable in /tmp/test_venv/bin/python
Installing setuptools, pip, wheel...
done.
```

New behaviour in v20.0.1:
```console
(virtualenv20.0.1_venv) $ virtualenv --version
virtualenv 20.0.1 from /tmp/virtualenv20.0.1_venv/local/lib/python2.7/site-packages/virtualenv/__init__.pyc
(virtualenv20.0.1_venv) $ virtualenv test_venv
```

It doesn't need to print the same messages of course, but at least something!

It is especially surprising in the case of running just `virtualenv`, which before was printing the help and returning an error code because of the missing `DEST_DIR`, while now creates a virtualenv in the new default `venv` and returns 0 without any feedback.

gaborbernat:
Has been commented during the beta that some people like the silent out of box behaviour, similarly how the rm command does not print just does. If you pass in a -v you'll get more feedback. I'm split between the current behaviour and maybe a one line final report 🤔on success. If you'd have warning/errors during the creation those you would already see.

nsoranzo:
The behaviour of "silent" shell commands is counter-intuitive and needs to be explained to new users, in my experience as Software Carpentry instructor. In virtualenv case, passing from giving feedback to be silent is even more surprising.
I find that `git` is a great example of how a command-line tool should give feedback, for example `git checkout <branch>` could be silent but it's not.

pfmoore:
As someone who said I liked the new quieter behaviour, I will note that I didn't realise that you could omit even the name of the environment (the `dest` parameter). I'm not keen on this behaviour - I'd prefer `dest` to be mandatory - but regardless of that I definitely don't like the fact that a bare `virtualenv` command creates an environment silently.

So, my preference now would be one or both of:

1. Make `dest` required.
2. Display a single line "Created virtual environment XXX" by default, suppressed with `-q`.

We could also bikeshed endlessly over the "correct" default for `dest_dir` (I prefer `.venv`) 😉

gaborbernat:
[...]

PR #1569, opened 2020-02-11
print a message after creating an environment

pradyunsg:
Resolves https://github.com/pypa/virtualenv/issues/1557

### DIFF
--- a/docs/changelog/1557.feature.rst
+++ b/docs/changelog/1557.feature.rst
@@ -1,0 +1,2 @@
+Print out a one line message about the created virtual environment when no :option:`verbose` is set, this can now be
+sileneced to get back the original behaviour via the :option:`quiet` flag - by :user:`pradyunsg`.
--- a/src/virtualenv/run/__init__.py
+++ b/src/virtualenv/run/__init__.py
@@ -20,6 +20,12 @@ def run_via_cli(args, options=None):
     """
     session = session_via_cli(args, options)
     session.run()
+
+    logging.warn(
+        "Created virtual environment via %s, using %s",
+        six.ensure_text(str(session.creator)),
+        six.ensure_text(str(session.seeder)),
+    )
     return session
 
 
--- a/src/virtualenv/run/__init__.py
+++ b/src/virtualenv/run/__init__.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 
+import six
+
 from ..config.cli.parser import VirtualEnvConfigParser
 from ..report import LEVELS, setup_report
 from ..session import Session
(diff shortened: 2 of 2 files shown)


src/virtualenv/run/__init__.py
```diff
@@ -20,6 +20,12 @@ def run_via_cli(args):
     """
     session = session_via_cli(args)
     session.run()
+
+    logging.warn(
+        "Created virtual environment via %s, using %s",
+        six.ensure_text(str(session.creator)),
+        six.ensure_text(str(session.seeder)),
+    )
```
pfmoore: Note that if #1568 is not accepted, the message here should also include the destination directory. (I don't really mind whether it does in the case where the user explicitly specifies the directory name, but it should if the default is used).

gaborbernat: the creator contains the destination directory 👍

pfmoore: Ah, that's great then. Sorry, I didn't have enough time to check the details.

pradyunsg: Yea, I spent quite a few minutes figuring that out myself. :)

gaborbernat: Sad to hear that 😢

gaborbernat:
Given #1568 I'm tempted to mark this won't do now.

docs/cli_interface.rst
```diff
@@ -10,8 +10,8 @@ CLI flags
 
 It modifies the environment variables in a shell to create an isolated Python environment, so you'll need to have a
 shell to run it. You can type in ``virtualenv`` (name of the application) followed by flags that control its
-behaviour. All options have sensible defaults, so you'll get a working virtual environment in a ``venv`` folder even
-if you don't pass any options. The default values for the command line options can be overridden via the
+behaviour. All options have sensible defaults, and there's on required argument: then name/path of the virtual
```
pradyunsg:
```suggestion
behaviour. All options have sensible defaults, and there's one required argument: then name/path of the virtual
```

docs/changelog/1548.feature.rst
```diff
@@ -0,0 +1,2 @@
+Print out a one line message about the created virtual environment when no :option:`verbose` is set, this can now be
+sileneced to get back the original behaviour via the :optio=:`quiet` flag - by :user:`pradyunsg`.
```
nsoranzo:
```suggestion
silenced to get back the original behaviour via the :optio=:`quiet` flag - by :user:`pradyunsg`.
```

docs/changelog/1557.feature.rst
```diff
@@ -0,0 +1,2 @@
+Print out a one line message about the created virtual environment when no :option:`verbose` is set, this can now be
+sileneced to get back the original behaviour via the :optio=:`quiet` flag - by :user:`pradyunsg`.
```
nsoranzo:
```suggestion
silenced to get back the original behaviour via the :optio=:`quiet` flag - by :user:`pradyunsg`.
```

docs/cli_interface.rst
pfmoore: Also, "the name/path", not "then name/path"

PR #1572, opened 2020-02-11
add option to view failure stacktraces

gaborbernat:


EpicWink:
Shouldn't this simply be part of `--verbose`? Maybe instead of [`if options.with_traceback:`](https://github.com/pypa/virtualenv/pull/1572/files#diff-79b6454ee141f415b84cfbe2e037cc60R32) use:
```python
if logging.getLogger("filelock").getEffectiveLevel() <= logging.INFO:
```

gaborbernat:
I view them as different things, so I prefer the explicit nature of how it's currently 🤔

PR #1579, opened 2020-02-11
support long path names for the app-data seeder generates console entry points

### LINKED ISSUE #997: Long env path names cause weird errors after activiting virtual environment

gaborbernat:
Resolves #997.

### DIFF
--- a/tests/unit/create/test_creator.py
+++ b/tests/unit/create/test_creator.py
@@ -297,3 +297,18 @@ def test_creator_input_passed_is_abs(tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     result = Creator.validate_dest("venv")
     assert str(result) == str(tmp_path / "venv")
+
+
+def test_create_long_path(current_fastest, tmp_path):
+    if sys.platform == "darwin":
+        max_shebang_length = 512
+    else:
+        max_shebang_length = 127
+    # filenames can be at most 255 long on macOS, so split to to levels
+    count = max_shebang_length - len(str(tmp_path))
+    folder = tmp_path / ("a" * (count // 2)) / ("b" * (count // 2)) / "c"
+    folder.mkdir(parents=True)
+
+    cmd = [str(folder)]
+    result = run_via_cli(cmd)
+    subprocess.check_call([str(result.creator.exe.parent / "pip"), "--version"])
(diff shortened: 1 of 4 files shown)


tests/unit/create/test_creator.py
```diff
@@ -297,3 +297,18 @@ def test_creator_input_passed_is_abs(tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     result = Creator.validate_dest("venv")
     assert str(result) == str(tmp_path / "venv")
+
+
+def test_create_long_path(current_fastest, tmp_path):
+    if sys.platform == "darwin":
+        max_shebang_length = 512
+    else:
+        max_shebang_length = 127
```
systemlogic: this should also be 512. Our builds are genarating longer path one for example
```
#!/home/gcdo/.cache/bazel/_bazel_gcdo/7973012a3ebb4144bd899078c66d5296/execroot/com_ibm_monorepo/bazel-out/k8-fastbuild/bin/cedp/web/rest_bzl_pipenv_interpreter.sh.runfiles/.venv/bin/python
# -*- coding: utf-8 -*-
import re
import sys

from pip._internal.cli.main import main

if __name__ == "__main__":
    sys.argv[0] = re.sub(r"(-script.pyw?|.exe)?$", "", sys.argv[0])
    sys.exit(main())
```

gaborbernat: These are the limits as taken from ``distlib`` of how it triggers the long-path special casing; those there are only used for tests.

asottile: the test is correct, on posixlike platforms the shebang limit is 127 characters

pradyunsg:
Any reason to not switch make_exe to use ScriptMaker?

gaborbernat:
it does now 👍

PR #1598, opened 2020-02-12
Revert "also disallow altsep beside pathsep inside paths"

### LINKED ISSUE #1582: With Git bash, cannot create virtualenv with a nested path.
With *Git Bash*, virtualenv refuses to create a virtualenv with a nested directory (ex: `foo/bar`), reporting that `/` is a path seperator and shouldn't be in the destination path.

I have no idea why `/` is considered a path separator (usually `;` or `:`, depending on OS, right?)

- [X] Minimal reproducible example or detailed descriptions
- [X] OS and `pip list` output

Environment: Windows 10, with Git Bash, using Python 3.7.
```
$ py -3.7 -m pip list
Package            Version
------------------ -------
appdirs            1.4.3
distlib            0.3.0
filelock           3.0.12
importlib-metadata 1.5.0
pip                20.0.2
setuptools         45.2.0
six                1.14.0
virtualenv         20.0.2
zipp               2.2.0
```

Example:
```
$ py -3.7 -m virtualenv foo/bar
usage: virtualenv [--version] [--with-traceback] [-v | -q] [--discovery {builtin}] [-p py] [--creator {builtin,cpython3-win,venv}] [--seeder {app-data,pip}] [--no-seed] [--activators comma_separated_list] [--clear] [--system-site-packages]
                  [--symlinks | --copies] [--download | --no-download] [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--wheel version] [--no-pip] [--no-setuptools] [--no-wheel] [--clear-app-data] [--symlink-app-data]

                  [--prompt prompt] [-h]
                  dest
virtualenv: error: argument dest: destination 'foo/bar' must not contain the path separator (/) as this would break the activation scripts
```

And FYI:
```
$ py -3.7
Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)] on win32
>>> import os
>>> os.pathsep
';'
>>>
```

vphilippon:
This reverts commit 524d95e782d810243cf9cd344a88711591c94efc.

Fixes #1582
This essentially allows back the usage of `/` in pathname on Windows for virtualenvs for nested dirs.
Ex: `python -m virtualenv foo/bar`.

- [x] added/updated test(s)

### DIFF
--- a/docs/changelog/1582.bugfix.rst
+++ b/docs/changelog/1582.bugfix.rst
@@ -1,0 +1,1 @@
+Allow the use of ``/`` as pathname component separator on Windows - by ``vphilippon``
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -117,12 +117,11 @@ class Creator(object):
                     encoding, "".join(refused.keys()), raw_value
                 )
             )
-        for char in (i for i in (os.pathsep, os.altsep) if i is not None):
-            if char in raw_value:
-                raise ArgumentTypeError(
-                    "destination {!r} must not contain the path separator ({}) as this would break "
-                    "the activation scripts".format(raw_value, char)
-                )
+        if os.pathsep in raw_value:
+            raise ArgumentTypeError(
+                "destination {!r} must not contain the path separator ({}) as this would break "
+                "the activation scripts".format(raw_value, os.pathsep)
+            )
 
         value = Path(raw_value)
         if value.exists() and value.is_file():
--- a/tests/unit/create/test_creator.py
+++ b/tests/unit/create/test_creator.py
@@ -25,13 +25,12 @@ from virtualenv.util.path import Path
 CURRENT = PythonInfo.current_system()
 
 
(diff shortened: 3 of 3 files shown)


gaborbernat:
This needs a test of using Unix paths on windows, and a changelog.

vphilippon:
So I'm trying to understand how/where the path provided is used to write a proper test (rather than sticking the `os.altsep` component in some pre-existing matrix).

As I dig in, I'm getting an omnious feeling with how `via_os_path.Path` might behave in some cases.

I'll take some extra time to look this up. I'd like to make sure that the path get standardized with `os.sep` early and explicitly, rather than "hoping" that the mangling around eventually normalize it.

vphilippon:
In the end, since the creator is the piece providing, validating and sanitizing the dest path input, I've put a test on `Creator.validate_dest`. The activators currently rely on the creator giving a valid path, so I think this is the best way to test this.
Plus, that means there's no need to worry about `os.altsep` in other pieces past that point.

If you think of cases this won't cover let me know.

Issue #1609, opened 2020-02-13
Failed to install virtuallenv from source on a centos-7 container


ssbarnea:
It seems that packaging is no longer compatible with older tool chains and even installing virtualenv from source become an interesting challenge.

```
[root@n0 virtualenv]# pip install wheel
Collecting wheel
  Downloading https://files.pythonhosted.org/packages/8c/23/848298cccf8e40f5bbb59009b32848a4c38f4e7f3364297ab3c3e2e2cd14/wheel-0.34.2-py2.py3-none-any.whl
Installing collected packages: wheel
Successfully installed wheel-0.34.2
You are using pip version 8.1.2, however version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
[root@n0 virtualenv]# python setup.py bdist_wheel
running bdist_wheel
running build
installing to build/bdist.linux-x86_64/wheel
running install
running install_egg_info
running egg_info
writing UNKNOWN.egg-info/PKG-INFO
writing top-level names to UNKNOWN.egg-info/top_level.txt
writing dependency_links to UNKNOWN.egg-info/dependency_links.txt
reading manifest file 'UNKNOWN.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files found matching '.gitattributes'
warning: no previously-included files found matching '.gitignore'
warning: no previously-included files found matching '.github/*'
warning: no previously-included files found matching 'azure-pipelines.yml'
warning: no previously-included files found matching 'readthedocs.yml'
warning: no previously-included files found matching 'tasks/release.py'
warning: no previously-included files found matching 'tasks/upgrade_wheels.py'
writing manifest file 'UNKNOWN.egg-info/SOURCES.txt'
Copying UNKNOWN.egg-info to build/bdist.linux-x86_64/wheel/UNKNOWN-0.0.0-py2.7.egg-info
running install_scripts
creating build/bdist.linux-x86_64/wheel/UNKNOWN-0.0.0.dist-info/WHEEL
creating 'dist/UNKNOWN-0.0.0-py2.py3-none-any.whl' and adding 'build/bdist.linux-x86_64/wheel' to it
adding 'UNKNOWN-0.0.0.dist-info/LICENSE'
adding 'UNKNOWN-0.0.0.dist-info/METADATA'
adding 'UNKNOWN-0.0.0.dist-info/WHEEL'
[...]

PR #1616, opened 2020-02-14
Add packaging environment

ssbarnea:
Adds a packaging environment and documents its use. This should clarify how
to build and install and unreleased version.

Related: #1609

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -45,6 +45,15 @@ supporting Python 2.7 use
 
 .. _compatibility-requirements:
 
+from source
+-----------
+
+If you want to test unreleased version you can run `tox -e packaging` which will produce source and wheel inside
+`dist/` folder.
+
+If you have `pip>=20` you can just do a `pip install git+https://github.com/pypa/virtualenv.git`, older versions will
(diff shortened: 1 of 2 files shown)


ssbarnea:
@gaborbernat If you do not want to advertise the pip+git install approach is ok, but in that case we should at least add some fail-safe code inside setup.py to make it fail quickly when run with outdated pip. Better to fail-fast when we know there is no chance of success instead of receiving bug reports from confused users.

ssbarnea:
@gaborbernat can I find you on freenode? my nick is `zbr`. I am in both #pypa-dev and #pypa channels all the time.

gaborbernat:
Gitter? I don't follow freenode 🤷‍♂

gaborbernat:
> but in that case we should at least add some fail-safe code inside setup.py to make it fail quickly when run with outdated pip.

I'm on board with that; added https://github.com/pypa/virtualenv/issues/1621

gaborbernat:
We now have 👍 for a chat-like interface, https://gitter.im/pypa/virtualenv

PR #1626, opened 2020-02-14
Default testing with lower constrainsts

ssbarnea:
Assure we run with oldest dependencies that we still support by default.

Adds optional `pyXY-upper` environments that bypass the lower
constraints and run tests using latest releases still inside our
supported range.

- [ ] added/updated test(s)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,19 +38,41 @@ jobs:
     jobs:
       py38:
         image: [linux, vs2017-win2016, macOs]
+      py38-lower:
+        name: py38_lower
+        image: [linux, vs2017-win2016, macOs]
       py37:
+        image: [linux, vs2017-win2016, macOs]
+      py37-lower:
+        name: py37_lower
         image: [linux, vs2017-win2016, macOs]
       py36:
         image: [linux, vs2017-win2016, macOs]
+      py36-lower:
+        name: py36_lower
+        image: [linux, vs2017-win2016, macOs]
       py35:
+        image: [linux, vs2017-win2016, macOs]
+      py35-lower:
+        name: py35_lower
         image: [linux, vs2017-win2016, macOs]
       py27:
         image: [linux, windows, macOs]
+      py27-lower:
+        name: py27_lower
+        image: [linux, windows, macOs]
       pypy:
+        image: [linux, windows, macOs]
+      pypy-lower:
+        name: pypy_lower
         image: [linux, windows, macOs]
       pypy3:
         image: [linux, windows, macOs]
-      fix_lint:
+      pypy3-lower:
+        name: pypy3_lower
+        image: [linux, windows, macOs]
+      fix-lint:
+        name: fix_lint
         image: [linux, windows]
       docs:
         image: [linux, windows]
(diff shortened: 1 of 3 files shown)


pip-constraint.txt
```diff
@@ -0,0 +1,4 @@
+# Used to test compatibility with oldest supported dependencies
+appdirs==1.4.3
```
gaborbernat: shouldn't we pin all dependencies here?

tox.ini
```diff
@@ -2,12 +2,7 @@
 minversion = 3.14.0
 envlist =
     fix_lint,
-    py38,
-    py37,
-    py36,
-    py35,
-    py34,
-    py27,
+    py{38,37,36,35,34,27}{,-upper},
```
gaborbernat: I'd prefer the other way around; default to no constraints, have the lower as extra; also we should add here pypy too 👍

gaborbernat:
And as far as CI go, I want to run both upper and lower versions 👍

pip-constraint.txt
ssbarnea: I pinned only a subset which did not cause any troubles. We will do the rest in follow-ups as we may have to make other changes. This change does not change what we ship.

gaborbernat: 😢 I'd personally rather have it fixed now if we have stuff that's causing issues

gaborbernat: I'd recommend making it one commit per dependency though 👍

ssbarnea:
> The CI now should pass; let's add those constraints 👍

Well, you will see some failures around zipapp tests. These seems to be highly affected by the presence of the lower requirements.

gaborbernat:
I don't think the zipapp tests should use the constraints; the zipapp will always package the latest versions of those libraries 🤔 let's not use the pip constraints for zipapp generation

azure-pipelines.yml
```diff
@@ -38,19 +38,41 @@ jobs:
     jobs:
       py38:
         image: [linux, vs2017-win2016, macOs]
+      py38-lower:
```
gaborbernat: please move them, separate group, as I did in my previous PR.

ssbarnea: I am afraid I lost that due to forced push. If you can do it again it would be great. After this I will add a condition to disable zipapp tests with lower constraints.

gaborbernat:
> Well, you will see some failures around zipapp tests. These seems to be highly affected by the presence of the lower requirements.

Thinking this through, maybe we should fix or pin those dependencies higher.

azure-pipelines.yml
stevegomez17: Good

stevegomez17: Good

ssbarnea: What I do not know is how to disable the 3 zipapp tests when using lower, in a way that makes sense as I do not want to rely on environment name.

Any ideas?

gaborbernat: I don't think we should disable the zipapp tests 🤷‍♂ we should instead make them pass 👍

ssbarnea: Zipapp is part of specific packaging, if making them run would require bumping requirements it would be a bad idea as it will prevent other users from using virtualenv, when they do not have any need for zipapp.

I am inclined to believe that a correct approach would be to make zipapp an extra and to install it only on non-lower envs and to update the tests to skip on that case. If it is an extra, we can manually bump min versions there w/o affecting a normal install.

I did not look into why zipapp fails, maybe is something simple but sadly these days I am already overloaded and I have no background of using zipapp. maybe is something simple and we can avoid the more complex approach I mention in previous paragraph.

gaborbernat: Before we talk about any further options we need to understand why zipapp fails, it shouldn't.

gaborbernat:
We moved to Github Actions, so I'll close this for now. Feel free to reopen it.

PR #1628, opened 2020-02-14
Better error message when build dependencies not satisfied

gaborbernat:
This only manifests on old platforms, but as we still support them
makes sense to provide better error messages, and relax a bit our
build requirements.

To validate using the docker build:
```
FROM pycontribs/centos:7
COPY . /w
RUN pip install 'setuptools==41.0.0' 'setuptools_scm==2'
RUN pip install /w -vvv --no-deps --disable-pip-version-check
RUN pip freeze 
```

Resolves https://github.com/pypa/virtualenv/issues/1621

### DIFF
--- a/docs/changelog/1621.feature.rst
+++ b/docs/changelog/1621.feature.rst
@@ -1,0 +1,2 @@
+Fail with better error message if trying to install source with unsupported ``setuptools``, allow ``setuptools-scm >= 2``
+and move to legacy ``setuptools-scm`` format to support better older platforms (``CentOS 7`` and such) - by :user:`gaborbernat`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,16 +2,9 @@
 requires = [
     "setuptools >= 41.0.0",
     "wheel >= 0.30.0",
-    "setuptools_scm[toml]>=3.4",
+    "setuptools_scm >= 2",
 ]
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -2,17 +2,10 @@
 requires = [
     "setuptools >= 41.0.0",
     "wheel >= 0.30.0",
-    "setuptools_scm[toml]>=3.4",
+    "setuptools_scm >= 2",
 ]
 build-backend = 'setuptools.build_meta'
 
-[tool.setuptools_scm]
```
gaborbernat: Sadly using the toml config degrades really bad on old platforms, so we've dropped it's usage in line with https://github.com/pypa/setuptools_scm#setuppy-usage

gaborbernat: @pradyunsg 😢

docs/changelog/1621.feature.rst
```diff
@@ -0,0 +1,2 @@
+Fail with better error message if trying to install source with unsupported ``setuptools``, allow ``setuptools-scm >= 2``
+and move to legacy ``setuptools-scm`` format to support better older platforms (``CentOs`` and such) - by :user:`gaborbernat`.
```
ssbarnea: Correct it to `CentOS 7` (case), the 8 is in much better shape.

Issue #1624, opened 2020-02-14
Discover python in a case insensitive way


xavfernandez:
Thanks your impressive work on the rewrite !

Everything works fine on my part but was actually surprised by:

```
$ virtualenv20 --version
virtualenv 20.0.3 from /home/xafer/.local/bin/virtualenv20/virtualenv/__init__.py

$ virtualenv20 -p cpython3.7 test
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/xafer/.local/bin/virtualenv20/__main__.py", line 169, in <module>
  File "/home/xafer/.local/bin/virtualenv20/__main__.py", line 165, in run
  File "/home/xafer/.local/bin/virtualenv20/virtualenv/__main__.py", line 19, in run
  File "/home/xafer/.local/bin/virtualenv20/virtualenv/run/__init__.py", line 22, in cli_run
  File "/home/xafer/.local/bin/virtualenv20/virtualenv/run/__init__.py", line 29, in session_via_cli
  File "/home/xafer/.local/bin/virtualenv20/virtualenv/run/__init__.py", line 51, in build_parser
RuntimeError: failed to find interpreter for Builtin discover of python_spec='cpython3.7'

$ virtualenv20 -p CPython3.7 test
created virtual environment in 82ms CPython3Posix(dest=/home/xafer/tmp/test, clear=False, global=False) with seeder FromAppData pip=latest setuptools=latest wheel=latest app_data_dir=/home/xafer/.local/share/virtualenv/seed-v1 via=copy
```

IMHO the interpreter discovery shouldn't consider the case.

gaborbernat:
@xavfernandez makes sense, feels like an oversight 👍

PrajwalM2212:
I would like to work on this.

ssbarnea:
I think it would be wrong to be case insensitive. Most platforms are case-insensitive. Even MacOS switched to defaulting to case-insensitive in recent years. 

If you give a param, you better know what you type.

gaborbernat:
[...]

PR #1629, opened 2020-02-14
Discover python interpreter in a case insensitive manner

PrajwalM2212:
Closes https://github.com/pypa/virtualenv/issues/1624

- [ ] added/updated test(s)

### DIFF
--- a/docs/changelog/1624.bugfix.rst
+++ b/docs/changelog/1624.bugfix.rst
@@ -1,0 +1,1 @@
+Discover python interpreter in a case insensitive manner - by :user:`PrajwalM2212`
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -237,7 +237,7 @@ class PythonInfo(object):
                 return False
 
         if impl_must_match:
-            if spec.implementation is not None and spec.implementation != self.implementation:
+            if spec.implementation is not None and spec.implementation.lower() != self.implementation.lower():
                 return False
 
         if spec.architecture is not None and spec.architecture != self.architecture:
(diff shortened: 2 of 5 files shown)


docs/changelog/1624.bugfix.rst
```diff
@@ -0,0 +1 @@
+Discover python interpreter in a case insensitive manner
```
gaborbernat: please add your user name as: `` - by :user:`PrajwalM2212` ``

You'll also need to add test for this to confirm this behaviour; probably by extending https://github.com/pypa/virtualenv/blob/master/tests/unit/discovery/py_info/test_py_info.py#L66

src/virtualenv/discovery/py_info.py
```diff
@@ -237,7 +237,7 @@ def satisfies(self, spec, impl_must_match):
                 return False
 
         if impl_must_match:
-            if spec.implementation is not None and spec.implementation != self.implementation:
+            if spec.implementation is not None and spec.implementation.lower() != self.implementation.lower():
```
gaborbernat: You'll want to do the same at https://github.com/pypa/virtualenv/blob/master/src/virtualenv/discovery/py_spec.py#L101, and also add test for that too 👍

tests/unit/discovery/py_info/test_py_info.py
```diff
@@ -56,7 +56,11 @@ def test_bad_exe_py_info_no_raise(tmp_path, caplog, capsys):
         list(
             "{}{}{}".format(impl, ".".join(str(i) for i in ver), arch)
             for impl, ver, arch in itertools.product(
-                ([CURRENT.implementation] + (["python"] if CURRENT.implementation == "CPython" else [])),
+                (
+                    [CURRENT.implementation]
+                    + (["python"] if CURRENT.implementation == "CPython" else [])
+                    + (["cpython"] if CURRENT.implementation == "CPython" else [])
```
gaborbernat: I think the correct logic here would be to add ``CURRENT.implementation.lower`` if that differs from ``CURRENT.implementation`` independent if it's CPython or not 👍

PR #1636, opened 2020-02-15
tox: call coverage as module

ssbarnea:
Avoids runtime warning and potential issues of using externally
installed coverage. Warning was observed on Azure.


gaborbernat:
Interesting, where this happens? Can you show an example?

gaborbernat:
I think this is a bug within virtualenv, mainly #1561

gaborbernat:
That being said this is alright, there are not that many drawbacks for calling like this so I'll accept

PR #1648, opened 2020-02-21
Prevent sys.prefix from leaking into child process on macOS

cjolowicz:
Fixes https://github.com/pypa/virtualenv/issues/1643

---

- [ ] added/updated test(s)

### DIFF
--- a/src/virtualenv/discovery/cached_py_info.py
+++ b/src/virtualenv/discovery/cached_py_info.py
@@ -103,9 +103,8 @@ def _run_subprocess(cls, exe):
     with ensure_file_on_disk(resolved_path) as resolved_path:
         cmd = [exe, "-s", str(resolved_path)]
 
-        # Prevent sys.prefix from leaking into the child process on macOS.
-        if sys.platform == "darwin":
-            os.environ.pop("__PYVENV_LAUNCHER__", None)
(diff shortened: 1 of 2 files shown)


src/virtualenv/discovery/cached_py_info.py
```diff
@@ -103,6 +103,10 @@ def _run_subprocess(cls, exe):
     with ensure_file_on_disk(resolved_path) as resolved_path:
         cmd = [exe, "-s", str(resolved_path)]
 
+        # Prevent sys.prefix from leaking into the child process on macOS.
+        if sys.platform == "darwin":
```
gaborbernat: I have a feeling that we should do this unconditionally 🤔 that being said I'm not a fan of modifying the os.environ directly; instead copy it, modify it and then pass it into the Popen

gaborbernat: we should also add some test; and changelog for this

gaborbernat: oh and do link the upstream bug ticket as a comment 👍

cjolowicz:
Thanks for merging! ...and sorry, did not get around to writing test and changelog earlier. 

Updated PR description with repro and analysis.

PR #1651, opened 2020-02-21
Fix error printing in bailout for Python < 2.7

AdamWill:
Ironically, the code that tries to print an error message when
we're running on Python < 2.7 can only work on Python 2.7 or
higher! Prior to 2.7, the curly-braces style of string formatting
required all fields to be explicitly numbered; you have to use
{0}, you cannot use {}. So if you actually try to run virtualenv
on Python 2.6, you get an error when trying to print the error!
This should fix that, and print the right error...

Signed-off-by: Adam Williamson <awilliam@redhat.com>

- [ ] added/updated test(s)

### DIFF
--- a/docs/changelog/1651.bugfix.rst
+++ b/docs/changelog/1651.bugfix.rst
@@ -1,0 +1,1 @@
+fix error printing in bailout for Python < 2.7 - by ``AdamWill`
--- a/virtualenv.py
+++ b/virtualenv.py
@@ -50,7 +50,7 @@ __version__ = "16.7.9"
 virtualenv_version = __version__  # legacy
 DEBUG = os.environ.get("_VIRTUALENV_DEBUG", None) == "1"
 if sys.version_info < (2, 7):
-    print("ERROR: {}".format(sys.exc_info()[1]))
+    print("ERROR: {0}".format(sys.exc_info()[1]))
     print("ERROR: this script requires Python 2.7 or greater.")
     sys.exit(101)
 


AdamWill:
Obviously this needs to be excepted from the linter, but I can't quite tell what linter is being used and hence how to except it.

AdamWill:
[pyupgrade issue filed](https://github.com/asottile/pyupgrade/issues/266).

gaborbernat:
FYI this branch is no longer maintained 🤷‍♂️

AdamWill:
oh, kay. Fedora is still shipping virtualenv from this branch, though, apparently...not sure why we're doing that if it isn't maintained...

gaborbernat:
They should be doing it from the master branch 👍

hroncok:
> not sure why we're doing that if it isn't maintained

Because we lack behind due to capacity reasons. The first 20.x virtualenv was released couple weeks ago, expect it will take a while before we migrate to it and expect it will only be done in rawhide. For released Fedoras, we are stuck with the "legacy" code from December. Note that the branch is called "legacy" and is no longer maintained, but it's not ancient. We maintain much much older stuff in Fedora than virtualenv 16.x.

gaborbernat:
@hroncok to be clear here, while the "legacy" branch is no longer maintained is mostly means I don't plan to work on it. That being said if someone puts in PRs I can still review and release them if needed.

hroncok:
Thanks. If you need any help reviewing prs or maintaining the legacy branch for bug fixes, let me know. We'll be on 16.x for at least until Fedora 31 goes EOL, most likely until F32 goes. That's more than a year.

gaborbernat:
It's all yours if you want it I'll tag you on prs opened against the legacy branch 👍

gaborbernat:
Also let me know when you want to cut a release 👍and I'll do it, e.g. for this pr.

hroncok:
We can backport the patch easily, so depends on how much work a release is.

gaborbernat:
5 minutes of my time 👍so alright as long as is not daily 😅

hroncok:
Ok. Let me try updating early next week for latest 16.x to see of everything is ok. I'll report back with some prs if needed or an ack for a release. Thanks again.

hroncok:
About his PR:

Before:

```
$ virtualenv -p python2.6
Running virtualenv with interpreter /usr/bin/python2.6
  File "/usr/lib/python3.8/site-packages/virtualenv.py", line 53, in <module>
    print("ERROR: {}".format(sys.exc_info()[1]))
ValueError: zero length field name in format
```

After:

```
$ virtualenv -p python2.6
Running virtualenv with interpreter /usr/bin/python2.6
ERROR: None
ERROR: this script requires Python 2.7 or greater.
```

This is improved, but the changed line should IMHO be removed entirely instead.

hroncok:
In a6d6ed791d94c39b7348192503866aff6a0ba458 a try-except block was changed for an if. Since that point, the error was probably saying None.

hroncok:
Followup: https://github.com/pypa/virtualenv/pull/1659

hroncok:
Also porting both commits to master.

gaborbernat:
master no longer has such checks 🤷‍♂

hroncok:
I've just realized that my master was old master from my fork. So no need for that indeed :+1:

PR #1659, opened 2020-02-24
Remove a leftover error line with empty exception info

hroncok:
Since a6d6ed791d94c39b7348192503866aff6a0ba458 there is no exception to report about.

Related to https://github.com/pypa/virtualenv/pull/1651

- [ ] added/updated test(s)

### DIFF
--- a/docs/changelog/1651.bugfix.rst
+++ b/docs/changelog/1651.bugfix.rst
@@ -1,1 +1,1 @@
-fix error printing in bailout for Python < 2.7 - by ``AdamWill`
+fix error printing in bailout for Python < 2.7 - by ``AdamWill` and ``hroncok``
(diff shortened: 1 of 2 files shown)


gaborbernat:
The CI is broke, so if you want to fix that you'll need another PR.

hroncok:
> The CI is broke, so if you want to fix that you'll need another PR.

I don't know azure much.

Anyway, I think you can do a release now if that's indeed easy for you, this was the only problem I've found.

gaborbernat:
would you like to bump embedded dependencies first or just cut master as it is?

hroncok:
Cut as is is fine, we embed our own. Thank you.

gaborbernat:
was saying more for general public than Fedora here tbh 🤷‍♂

gaborbernat:
see https://pypi.org/project/virtualenv/16.7.10/

hroncok:
Got it, though the general public most likely just fetches the latest and greatest.

hroncok:
Could you please also upload the sdist?

gaborbernat:
Sure the network just dropped while doing it but will do in the morning 👍

hroncok:
I see the sdist. Thanks again.

PR #1661, opened 2020-02-25
handle application data folder is read only

yakkle:
when using docker, if `user_data_dir()` isn't writable directory,
   `default_data_dir()` use `system temp directory` + `virtualenv`.
   for example, tempdir is `/tmp`, it use `/tmp/virtualenv`

- [x] added/updated test(s)


gaborbernat:
I think we need a better solution for this in a sense that we should clean-up the temporay folder when this is not the app-data.

gaborbernat:
@yakkle thanks for starting the work on this, I'll take up continuing this PR, this requires major changes to make it right.

PR #1673, opened 2020-02-26
Fix setting PYTHONPATH on Python 2 not working

jd:
- ensure we do not rewrite the ``PYTHONPATH``
- ensure we respect the ``-E`` flag.

### DIFF
--- a/src/virtualenv/create/via_global_ref/builtin/python2/site.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/site.py
@@ -97,7 +97,7 @@ def rewrite_standard_library_sys_path():
                 value = "{}{}".format(base_prefix, value[len(prefix) :])
             elif dir_starts_with(value, exec_prefix):
                 value = "{}{}".format(base_exec_prefix, value[len(exec_prefix) :])
-        sys.path[at] = value
+            sys.path[at] = value
 
 
 def dir_starts_with(directory, prefix):
--- a/src/virtualenv/create/via_global_ref/builtin/python2/site.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/site.py
@@ -89,19 +89,25 @@ def rewrite_standard_library_sys_path():
         skip_rewrite = value == exe_dir  # don't fix the current executable location, notably on Windows this gets added
         skip_rewrite = skip_rewrite  # ___SKIP_REWRITE____
         if not skip_rewrite:
-            if dir_starts_with(value, exe_dir):
-                # content inside the exe folder needs to remap to original executables folder
-                orig_exe_folder = base_executable[: base_executable.rfind(sep)]
-                value = "{}{}".format(orig_exe_folder, value[len(exe_dir) :])
-            elif dir_starts_with(value, prefix):
-                value = "{}{}".format(base_prefix, value[len(prefix) :])
-            elif dir_starts_with(value, exec_prefix):
-                value = "{}{}".format(base_exec_prefix, value[len(exec_prefix) :])
-            sys.path[at] = value
+            sys.path[at] = translate_directory(
+                value, base_executable, exe_dir, exec_prefix, base_prefix, prefix, base_exec_prefix
+            )
 
+    # Now that we can access os.environ["PYTHONPATH"], we can revert our
+    # rewrite of thoses paths done above
+    import os
 
-def dir_starts_with(directory, prefix):
-    return directory == prefix or directory.startswith(prefix + sep)
+    python_paths = os.environ.get("PYTHONPATH", "").split(":")
+    for ppath in map(abs_path, python_paths):
+        translated_ppath = translate_directory(
+            ppath, base_executable, exe_dir, exec_prefix, base_prefix, prefix, base_exec_prefix
+        )
+        try:
+            at = sys.path.index(translated_ppath)
+        except ValueError:
+            pass
+        else:
+            sys.path[at] = ppath
 
 
 def abs_path(value):
@@ -116,6 +122,22 @@ def abs_path(value):
         i -= 1
     value = sep.join(keep[::-1])
     return value
+
+
+def translate_directory(directory, base_executable, exe_dir, exec_prefix, base_prefix, prefix, base_exec_prefix):
+    if dir_starts_with(directory, exe_dir):
+        # content inside the exe folder needs to remap to original executables folder
+        orig_exe_folder = base_executable[: base_executable.rfind(sep)]
+        return "{}{}".format(orig_exe_folder, directory[len(exe_dir) :])
+    elif dir_starts_with(directory, prefix):
+        return "{}{}".format(base_prefix, directory[len(prefix) :])
+    elif dir_starts_with(directory, exec_prefix):
+        return "{}{}".format(base_exec_prefix, directory[len(exec_prefix) :])
+    return directory
+
+
+def dir_starts_with(directory, prefix):
+    return directory == prefix or directory.startswith(prefix + sep)
 
 
 def disable_user_site_package():
--- a/src/virtualenv/create/via_global_ref/builtin/python2/site.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/site.py
@@ -89,15 +89,19 @@ def rewrite_standard_library_sys_path():
         skip_rewrite = value == exe_dir  # don't fix the current executable location, notably on Windows this gets added
         skip_rewrite = skip_rewrite  # ___SKIP_REWRITE____
         if not skip_rewrite:
-            if value.startswith(exe_dir):
+            if dir_starts_with(value, exe_dir):
                 # content inside the exe folder needs to remap to original executables folder
                 orig_exe_folder = base_executable[: base_executable.rfind(sep)]
                 value = "{}{}".format(orig_exe_folder, value[len(exe_dir) :])
-            elif value.startswith(prefix):
+            elif dir_starts_with(value, prefix):
                 value = "{}{}".format(base_prefix, value[len(prefix) :])
-            elif value.startswith(exec_prefix):
+            elif dir_starts_with(value, exec_prefix):
                 value = "{}{}".format(base_exec_prefix, value[len(exec_prefix) :])
         sys.path[at] = value
+
+
+def dir_starts_with(directory, prefix):
+    return directory == prefix or directory.startswith(prefix + sep)
 
 
 def abs_path(value):
--- a/src/virtualenv/create/via_global_ref/builtin/python2/site.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/site.py
@@ -78,66 +78,59 @@ def read_pyvenv():
 
 def rewrite_standard_library_sys_path():
     """Once this site file is loaded the standard library paths have already been set, fix them up"""
-    exe = abs_path(sys.executable)
+    exe, prefix, exec_prefix = get_exe_prefixes(base=False)
+    base_exe, base_prefix, base_exec = get_exe_prefixes(base=True)
     exe_dir = exe[: exe.rfind(sep)]
-    prefix, exec_prefix = abs_path(sys.prefix), abs_path(sys.exec_prefix)
-    base_prefix, base_exec_prefix = abs_path(sys.base_prefix), abs_path(sys.base_exec_prefix)
-    base_executable = abs_path(sys.base_executable)
-    for at, value in enumerate(sys.path):
-        value = abs_path(value)
-        # replace old sys prefix path starts with new
-        skip_rewrite = value == exe_dir  # don't fix the current executable location, notably on Windows this gets added
+    for at, path in enumerate(sys.path):
+        path = abs_path(path)  # replace old sys prefix path starts with new
+        skip_rewrite = path == exe_dir  # don't fix the current executable location, notably on Windows this gets added
         skip_rewrite = skip_rewrite  # ___SKIP_REWRITE____
         if not skip_rewrite:
-            sys.path[at] = translate_directory(
-                value, base_executable, exe_dir, exec_prefix, base_prefix, prefix, base_exec_prefix
-            )
+            sys.path[at] = map_path(path, base_exe, exe_dir, exec_prefix, base_prefix, prefix, base_exec)
 
-    # Now that we can access os.environ["PYTHONPATH"], we can revert our
-    # rewrite of thoses paths done above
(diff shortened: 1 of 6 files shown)


src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -114,6 +108,45 @@ def abs_path(value):
...
+        if not skip_rewrite:
+            sys.path[at] = translate_directory(value)
+
+
+def revert_pythonpath_rewrite():
+    # Now that we can access os.environ["PYTHONPATH"], we can revert our
+    # rewrite of thoses paths done in `rewrite_standard_library_sys_path`.
+    import os
+
+    python_paths = os.environ.get("PYTHONPATH", "").split(":")
```
gaborbernat: fairly certain these should be absolute paths too 👍

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -114,6 +108,45 @@ def abs_path(value):
...
+    for at, value in enumerate(sys.path):
+        value = abs_path(value)
+        # replace old sys prefix path starts with new
+        skip_rewrite = value == exe_dir  # don't fix the current executable location, notably on Windows this gets added
+        skip_rewrite = skip_rewrite  # ___SKIP_REWRITE____
+        if not skip_rewrite:
+            sys.path[at] = translate_directory(value)
+
+
+def revert_pythonpath_rewrite():
```
gaborbernat: I prefer python path to be written as ``python_path``

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -114,6 +108,45 @@ def abs_path(value):
     return value
 
 
+def translate_directory(directory):
```
gaborbernat: this should be after ``rewrite_standard_library_sys_path`` to ordered top-down by the occurrence

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -6,15 +6,29 @@
 """
 import sys
 
+base_executable = None
```
gaborbernat: please don't, don't introduce globals they just make code hard to follow

jd: Ok, though the list of arguments is going to be long to carry everwhere :(

src/virtualenv/create/via_global_ref/builtin/python2/site.py
jd: I'm not sure it's going to make the function name clear that we're talking about `PYTHONPATH` though. You sure?

src/virtualenv/create/via_global_ref/builtin/python2/site.py
jd: Fixed

src/virtualenv/create/via_global_ref/builtin/python2/site.py
gaborbernat: I personally prefer that over globals. Also, we don't have to carry everything everywhere.

src/virtualenv/create/via_global_ref/builtin/python2/site.py
gaborbernat: yes

gaborbernat: yes

src/virtualenv/create/via_global_ref/builtin/python2/site.py
gaborbernat: I don't see the change, conversations are usually resolved by the one starting the discussion.

gaborbernat: I don't see. The conversation is resolved by the one starting the conversation for this reason.

jd:
Is there anything left that needs fixing?

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -13,8 +13,14 @@ def main():
...
     sys.base_exec_prefix = config["base-exec-prefix"]
     sys.base_executable = config["base-executable"]
+    base_executable = abs_path(sys.base_executable)
+    base_prefix, base_exec_prefix = abs_path(sys.base_prefix), abs_path(sys.base_exec_prefix)
+    exe = abs_path(sys.executable)
+    exe_dir = exe[: exe.rfind(sep)]
+    prefix, exec_prefix = abs_path(sys.prefix), abs_path(sys.exec_prefix)
     global_site_package_enabled = config.get("include-system-site-packages", False) == "true"
-    rewrite_standard_library_sys_path()
+    rewrite_standard_library_sys_path(base_executable, exe_dir, exec_prefix, base_prefix, prefix, base_exec_prefix)
```
gaborbernat: Feels to me we only need all these variables here because we're doing the rewrite and revert in two functions, cant' we just do it in one to avoid this extra overhead?

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -60,6 +66,10 @@ def load_host_site():
 sep = "\\" if sys.platform == "win32" else "/"  # no os module here yet - poor mans version
 
 
+def dir_starts_with(directory, prefix):
```
gaborbernat: this is out of order, usually, functions are defined after they are called

gaborbernat: so they read top down

jd: You should invest in some automated checking tools. Bothering contributors with such trivial details on each review is kinda frustrating.

src/virtualenv/create/via_global_ref/builtin/python2/site.py
jd: See L140

src/virtualenv/create/via_global_ref/builtin/python2/site.py
jd: Sure

jd:
Could you give me some hints about where to add some tests?
I took a look at https://github.com/pypa/virtualenv/pull/1641 but it's unclear to me how you added/changed the tests.

src/virtualenv/create/via_global_ref/builtin/python2/site.py
gaborbernat: That's why we have pre commit. I haven't got a tool for checking this yet though.

gaborbernat:
In that pr the test were adding brew Python to the ci matrix. Here you want a test inside test_create 👍set Python path, run create, check sys path of the debug property 👍

src/virtualenv/create/via_global_ref/builtin/python2/site.py
gaborbernat: That being said feel free to take this as far as you're willing to. I'll pick up and clean things up once you're done, and ship it. As a maintainer, I feel like I need to guide the code to have some unified style and ensure PRs don't just fix issues but leave the codebase in a maintainable state. Hope you understand.

jd: You need a CI check, not (just) pre-commit.

If your plan is to manually guide people to your style and pick up behind people, you're going to be exhausted before long and even might lose contributors in the process. Which is not something I wish you!

gaborbernat: As I said I'm not aware of a tool that does this at the moment. Pre commit is run part of the ci. So a pre commit test is a ci test.

jd:
I'm blocked by https://github.com/pypa/virtualenv/issues/1683 to try locally to add tests :(

jd:
Any idea how you set UTF-8 strings in a Windows env? 🤔

pfmoore:
Environment variables are Unicode in Windows, so I'm not quite sure the question makes sense. Having said that, if this is Python 2, `os.environ` probably uses the ANSI APIs to read the environment, so the practical answer would likely be "you can't - you always get the default encoding".

jd:
My question is related to the CI failure. IIUC the dir contains Unicode chars and setting that as `PYTHONPATH` is a problem… :(

pfmoore:
Ah, OK. So passing a general Unicode value through to the subprocess' environment. I don't think you can, in Python 2 (IIRC, you can only use characters that are valid in the system ANSI codepage) 🙁 (I may be wrong, it's a long while since I used Python 2).

Sorry I can't be more help.

tests/conftest.py
```diff
@@ -311,6 +311,28 @@ def change_env_var(key, value):
...
+def override_env_var():
+
+    original_env = {}
+
+    def _environ_override(var, value):
+        original_env.setdefault(var, os.environ.get(var, _ENV_EMPTY))
+        os.environ[var] = value
+
+    try:
+        yield _environ_override
```
gaborbernat: any reason not just use the builtin ``monkeypatch.setenv``?

jd: No, I just didn't know about it!

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -78,40 +78,59 @@ def read_pyvenv():
...
-                value = "{}{}".format(orig_exe_folder, value[len(exe_dir) :])
-            elif value.startswith(prefix):
-                value = "{}{}".format(base_prefix, value[len(prefix) :])
-            elif value.startswith(exec_prefix):
-                value = "{}{}".format(base_exec_prefix, value[len(exec_prefix) :])
-        sys.path[at] = value
+            sys.path[at] = map_path(path, base_exe, exe_dir, exec_prefix, base_prefix, prefix, base_exec)
+
+    # the rewrite above may have changed elements from PYTHONPATH, revert these if on
+    if sys.flags.ignore_environment:
```
gaborbernat: @jd added handling for the ``-E`` flag

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -78,40 +78,59 @@ def read_pyvenv():
...
+        return
+    import os
+
+    python_paths = []
+    if "PYTHONPATH" in os.environ and os.environ["PYTHONPATH"]:
+        for path in os.environ["PYTHONPATH"].split(os.pathsep):
+            if path not in python_paths:
+                python_paths.append(path)
+    sys.path[: len(python_paths)] = python_paths
```
gaborbernat: to supports setting paths both within the host prefix and the venv prefix instead of trying to fix up let's just overwrite the rewritten part that belongs to python path

tests/unit/create/test_creator.py
```diff
@@ -386,3 +388,58 @@ def test_create_distutils_cfg(creator, tmp_path, monkeypatch):
 
     package_folder = result.creator.platlib / "demo"  # prefix is set to the virtualenv prefix for install
     assert package_folder.exists()
+
+
+@pytest.mark.parametrize("python_path_on", [True, False], ids=["on", "off"])
+@pytest.mark.skipif(PY3, reason="we rewrite sys.path only on PY2")
+def test_python_path(monkeypatch, tmp_path, python_path_on):
```
gaborbernat: have a dedicated test for this use case

tests/unit/create/test_creator.py
```diff
@@ -386,3 +388,58 @@ def test_create_distutils_cfg(creator, tmp_path, monkeypatch):
...
+        cmd.extend(["-c", "import json; import sys; print(json.dumps(sys.path))"])
+        print(cmd)
+        return json.loads(subprocess.check_output(cmd))
+
+    monkeypatch.delenv(str("PYTHONPATH"), raising=False)
+    base = _get_sys_path()
+
+    # note the value result.creator.interpreter.system_stdlib cannot be set, as that would disable our custom site.py
+    python_paths = [
+        str(Path(result.creator.interpreter.prefix)),
```
gaborbernat: test various edge cases for the python path as just one path within the venv

tests/unit/create/test_creator.py
```diff
@@ -386,3 +388,58 @@ def test_create_distutils_cfg(creator, tmp_path, monkeypatch):
...
+        assert not (set(base) - set(extra_all))  # all base paths are present
+        abs_python_paths = list(OrderedDict((os.path.abspath(ensure_text(i)), None) for i in python_paths).keys())
+
+        extra_as_python_path = extra_all[: len(abs_python_paths)]
+        assert abs_python_paths == extra_as_python_path  # python paths are there at the start
+
+        non_python_path = extra_all[len(abs_python_paths) :]
+        assert non_python_path == [i for i in base if i not in extra_as_python_path]
+    else:
+        assert base == extra_all
```
gaborbernat: also, test that we respect ``-E``

src/virtualenv/create/via_global_ref/builtin/python2/site.py
jd: I'm a bit confused here. How can you be sure the first paths in `sys.path` can be replaced like that?

src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -78,40 +78,59 @@ def read_pyvenv():
...
+    # the rewrite above may have changed elements from PYTHONPATH, revert these if on
+    if sys.flags.ignore_environment:
+        return
+    import os
+
+    python_paths = []
+    if "PYTHONPATH" in os.environ and os.environ["PYTHONPATH"]:
+        for path in os.environ["PYTHONPATH"].split(os.pathsep):
+            if path not in python_paths:
+                python_paths.append(path)
```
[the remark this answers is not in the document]
gaborbernat: Keeping the order is very important, a set would loose that information

gaborbernat:
@jd it's just how pythonpath is parsed at the interpreter level.

gaborbernat:
Hello, a fix for this issue has been released via virtualenv 20.0.8; see https://pypi.org/project/virtualenv/20.0.8/ (https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-8-2020-03-04). Please give a try and report back if your issue has not been addressed; if not, please comment here, and we'll reopen the ticket. We want to apologize for the inconvenience this has caused you and say thanks for having patience while we resolve the unexpected bugs with this new major release.
thanks

PR #1679, opened 2020-02-27
Default prompt of environment name in batch shell

### LINKED ISSUE #1570: No (env) indication visible - Windows 10
In all the time I've used virtualenv (i.e since 2 years now), whenever I activate a virtual environment there is an (env) indication like this:
image

But with the new virtualenv version (20.0.1), I have no indication of whether the virtualenv is active or not: 
image

Was this intended? How do I get the previous functionality back? virtualenv version 20 seems much different than 16.

spetafree:
Fixes #1570, by setting a default prompt for batch (windows cmd.exe) shell based on the environment name.

In bash activation script, the check for using the default prompt is done in the bash shell script. It is more complicated to check that and get the environment folder name in bat script, and because of that I have decided to implement that check in the python code, by inheriting the `replacements` method and adding the check there.

I had a few failures in the local tests I tried to run, but is seems to me that these are not connected to my change - the same tests happen also on the current master.

- [X] ran the linter to address style issues (``tox -e fix_lint``)

Above.

I am actually at loss how to test this, the current tests seem not to test that issue also on other
shells where I assume it works.

- [X] added news fragment in ``docs/changelog`` folder
I do not understand how should I name the fragment, I hope it is ok.

n/a

### DIFF
--- a/src/virtualenv/activation/batch/__init__.py
+++ b/src/virtualenv/activation/batch/__init__.py
@@ -21,9 +21,3 @@ class BatchActivator(ViaTemplateActivator):
         # ensure the text has all newlines as \r\n - required by batch
         base = super(BatchActivator, self).instantiate_template(replacements, template, creator)
         return base.replace(os.linesep, "\n").replace("\n", os.linesep)
-
-    def replacements(self, creator, dest_folder):
-        base_replacements = super(BatchActivator, self).replacements(creator, dest_folder)
-        if not base_replacements["__VIRTUAL_PROMPT__"]:
-            base_replacements["__VIRTUAL_PROMPT__"] = "(%s) " % creator.env_name
-        return base_replacements
--- a/src/virtualenv/activation/batch/__init__.py
+++ b/src/virtualenv/activation/batch/__init__.py
@@ -21,3 +21,9 @@ class BatchActivator(ViaTemplateActivator):
         # ensure the text has all newlines as \r\n - required by batch
         base = super(BatchActivator, self).instantiate_template(replacements, template, creator)
         return base.replace(os.linesep, "\n").replace("\n", os.linesep)
+
+    def replacements(self, creator, dest_folder):
+        base_replacements = super(BatchActivator, self).replacements(creator, dest_folder)
+        if not base_replacements["__VIRTUAL_PROMPT__"]:
+            base_replacements["__VIRTUAL_PROMPT__"] = "(%s) " % creator.env_name
+        return base_replacements
--- a/docs/changelog/1679.bugfix.rst
+++ b/docs/changelog/1679.bugfix.rst
@@ -1,1 +1,1 @@
-    fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour.
+    Fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour. - by :user:`spetafree`.
--- a/docs/changelog/1679.bugfix.rst
+++ b/docs/changelog/1679.bugfix.rst
@@ -1,1 +1,1 @@
-Fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour. - by :user:`spetafree`.
+Fix Batch activation script shell prompt to display environment name by default - by :user:`spetafree`.
--- a/docs/changelog/1679.bugfix.rst
+++ b/docs/changelog/1679.bugfix.rst
@@ -1,1 +1,1 @@
-    Fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour. - by :user:`spetafree`.
+Fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour. - by :user:`spetafree`.
--- a/docs/changelog/1679.bugfix.rst
+++ b/docs/changelog/1679.bugfix.rst
@@ -1,0 +1,1 @@
+    fix windows batch (cmd.exe) shell prompt to display environment name by default, simliar to bash behaviour.
--- a/src/virtualenv/activation/batch/activate.bat
+++ b/src/virtualenv/activation/batch/activate.bat
@@ -13,7 +13,12 @@ if defined _OLD_VIRTUAL_PROMPT (
     )
 )
 if not defined VIRTUAL_ENV_DISABLE_PROMPT (
-    set "PROMPT=__VIRTUAL_PROMPT__%PROMPT%"
+    set "ENV_PROMPT=__VIRTUAL_PROMPT__"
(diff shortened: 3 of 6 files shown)


src/virtualenv/activation/batch/__init__.py
```diff
@@ -21,3 +21,9 @@ def instantiate_template(self, replacements, template, creator):
         # ensure the text has all newlines as \r\n - required by batch
         base = super(BatchActivator, self).instantiate_template(replacements, template, creator)
         return base.replace(os.linesep, "\n").replace("\n", os.linesep)
+
+    def replacements(self, creator, dest_folder):
+        base_replacements = super(BatchActivator, self).replacements(creator, dest_folder)
+        if not base_replacements["__VIRTUAL_PROMPT__"]:
+            base_replacements["__VIRTUAL_PROMPT__"] = "(%s) " % creator.env_name
```
gaborbernat: Shouldn't this change be in the template? Not the parameterization?

spetafree: I am not sure what do you mean, I will try to explain my reasoning:

In all other shells, there is a check in the shell script if the value replaced by virtual prompt is empty, and if so, it is set to the `env_name` in bat it is hard to do it, at least to my limited expertise in bat programing. To overcome this I have made the check here.

I do not think this should be in the `via_template.py` file if that what you mean, I see no need to complicate other shells because of bat limitations.

spetafree: A possible different change would be to change
https://github.com/pypa/virtualenv/blob/7649968fcc673336e510a18d4d1d20feef48fe4b/src/virtualenv/activation/via_template.py#L34
```python
 "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
```
to

````python
 "__VIRTUAL_PROMPT__": '(%s) ' % creator.env if self.flag_prompt is None else self.flag_prompt,
```
But I am not sure and will be unable to check how it effects all other shells, and the only ones I have access to are cmd.exe and (git) bash (both on windows).

spetafree: > 
> 
> Was thinking personally this should be a change within activate.bat 🤔

Correctly checking that a variable is empty and getting the folder name in bat is rather ugly, but I can try doing it, I'll need to learn how to. Seemed to me cleaner that way, similar to what have been in the previous version of vritualenv - the windows bat case was handled in python.

gaborbernat: Leave it as it is then, will investigate next week 👍

gaborbernat:
I'm not saying we absolutely must do it in batch but would like to see the two options 👍and how they would differ and if it's worth doing differently here than for all other scripts.

spetafree:
What do you think about the change in `via_template.py`? Maybe it will fix all the shells at once? The prompt will just always default to '(env_name) ' instead of to empty string in python, and the shells scripts may be simplified.

gaborbernat:
Can't tell from top of my head at the moment. Need to have a think about it.

spetafree:
It may take me some time, but may try to install some of the other shells and see how it works. There is also the question of how to specify an empty prompt in command line. (maybe we will need a --no-prompt flag)
Thanks for reviewing the PR anyway, It will probably be a few days at least until I can learn enough about all these shells, I'm much more familiar with python that with shell scripts 😄

gaborbernat:
No worries, take your time, I'll not be able to take a deeper look at this until next Tuesday.

src/virtualenv/activation/batch/__init__.py
spetafree: Turns out fixing it it in the bat file, at least for the simple case was not so hard. Although, I can't say I understand why is this exact loop works and other I tried failed, so I'm not sure that it will work in edge cases, if there are.

I hope it's better now

spetafree:
I'll try to write the test tomorrow. Do you think that adding another command to print the prompt environment variable in _get_test_lines and assert it in assert_output in `test_batch.py` is the correct way?

spetafree:
Any chance you'll consider re-opening it, I was not able to work on it the last couple of days? Tomorrow I'll hopefully be able to work on it again.

spetafree:
I'll explain the changes:

1. I had to change the handling of the variables in the batch script to support the non ascii characters. The following seems to work:

https://github.com/pypa/virtualenv/blob/dddf7cdacdb0042a19e9a946721f8898a349bd02/src/virtualenv/activation/batch/activate.bat#L17

2.  The folder name in tests:  `e-$ èрт🚒♞中片-j` contains a single `$` character which is an escape character in batch prompts, and causes incorrect rendering. I have changed it to `$$`. I hope it's ok with all other shells, and doesn't drop some edge case that was checked with single `$`.

3. I have added testing with and without `--prompt` to all shells.

4. I have expanded the test for batch activation test to check for correct prompt. I have not added the same test to other shells. I can do it for bash, which I am somewhat familiar with, but I do not know how to test this for other shells.

p.s. @gaborbernat,  This is my second ever attempt at a public open source pull request, and I would like to thank you very much for your handling of it 😄. It is a much nicer treatment than my last attempt...

spetafree:
I see I missed something. Will fix in a few hours.

spetafree:
I have no idea why to the zipapp test fails, and how is this related to my PR. 

Any help will be appreciated, I will continue investigating. But currently I can't reproduce this on my windows machine.

spetafree:
@gaborbernat Thanks a lot 😄.

But, there is no test for the prompt now, why did you decide to approve? 

Also the parameteriztion in conftest.py change has no effect now as far as I understand.

gaborbernat:
Why would it not have any effect, I think this is alright 🤔

spetafree:
Because no test looks at the prompt now as far as I understand. 

The only test that actually checked the prompt was in this commit:

https://github.com/pypa/virtualenv/pull/1679/commits/dddf7cdacdb0042a19e9a946721f8898a349bd02

gaborbernat:
That's not true, the tests already looked at it, just did not run the test suite Without prompt flag.

spetafree:
Ok, then I couldn't understand the tests - I didn't find a test that actually reads the prompt. I'll look again. Thanks anyway.

gaborbernat:
Hello, a fix for this issue has been released via virtualenv 20.0.8; see https://pypi.org/project/virtualenv/20.0.8/ (https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-8-2020-03-04). Please give a try and report back if your issue has not been addressed; if not, please comment here, and we'll reopen the ticket. We want to apologize for the inconvenience this has caused you and say thanks for having patience while we resolve the unexpected bugs with this new major release.
thanks

PR #1698, opened 2020-03-06
Better handling of loader patch on python 3

gaborbernat:


gaborbernat:
@asottile any better this?

PR #1702, opened 2020-03-09
upgrade setuptools to 46.0.0 from 45.3.0 for Python3.5+

gaborbernat:


gaborbernat:
Hello, a fix for this issue has been released via virtualenv 20.0.10; see https://pypi.org/project/virtualenv/20.0.10/ (https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-10-2020-03-10). Please give a try and report back if your issue has not been addressed; if not, please comment here, and we'll reopen the ticket. We want to apologize for the inconvenience this has caused you and say thanks for having patience while we resolve the unexpected bugs with this new major release.
thanks

PR #1707, opened 2020-03-10
Improve installation documentation

### LINKED ISSUE #1618: Document pip requirements for installation
We should document under https://virtualenv.pypa.io/en/latest/installation.html#via-pip pip minimal version requirement for both wheel and sdist.

### LINKED ISSUE #1619: Document not supporting pre PEP-518 build systems and suggest workarounds
Though we don't support ``setup.py bdist_wheel`` we should mention that if one ensures build-requires are present things "should work". But in such cases, the responsibility falls onto whoever does the build to ensure pyproject.toml is kept in sync with such dependencies.

### LINKED ISSUE #1620: Document installing latest from source
Prefer a new enough pip. Not sure if we want them to build the wheel/sdist... if the goal is to have latest, just install via a new enough pip straight from git. See #1616

gaborbernat:
Resolves #1618.
Resolves #1619.
Resolves #1620.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -19,12 +19,30 @@ via pip
 Alternatively you can install it within the global Python interpreter itself (perhaps as a user package via the
 ``--user`` flag). Be cautious if you are using a python install that is managed by your operating system or
 another package manager. ``pip`` might not coordinate with those tools, and may leave your system in an
-inconsistent state.
+inconsistent state. Note, if you go down this path you need to ensure pip is new enough per the subsections below:
 
 .. code-block:: console
 
     python -m pip install --user virtualenv
     python -m virtualenv --help
+
+wheel
+~~~~~
+Installing virtualenv via a wheel (default with pip) requires an installer that can understand the ``python-requires``
+tag (see `PEP-503 <https://www.python.org/dev/peps/pep-0503/>`_), with pip this is version ``9.0.0`` (released 2016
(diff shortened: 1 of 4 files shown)


docs/installation.rst
```diff
@@ -19,13 +19,31 @@ via pip
...
 .. code-block:: console
 
     python -m pip install --user virtualenv
     python -m virtualenv --help
 
+wheel
+~~~~~
+Installing virtualenv via a wheel (default with pip) requires an installer that can understand the ``python-requires``
+tag (see `PEP-503 <https://www.python.org/dev/peps/pep-0503/>`_), with pip this is version ``9.0.0`` (released 2016
+November). Furthermore, in case you're not installing it via the PyPi you need to be using a mirror that correctly
```
pradyunsg: PyPI

docs/installation.rst
```diff
@@ -43,6 +61,23 @@ minor release use the link ``https://bootstrap.pypa.io/virtualenv/x.y/virtualenv
 supporting Python 2.7 use
 `https://bootstrap.pypa.io/virtualenv/2.7/virtualenv.pyz <https://bootstrap.pypa.io/2.7/virtualenv/virtualenv.pyz>`_.
 
+via ``setup.py``
+----------------
+We don't recommend and officially support this method. One should prefer using an installer that supports
```
pradyunsg: s/and/nor/

PR #1714, opened 2020-03-11
Allow missing .py files if .pyc is present

tucked:
Python can run with just `.pyc` files; however, if only a `.pyc` file is present (e.g. `os.pyc`), `virtualenv` will throw a `RuntimeError: No virtualenv implementation for PythonInfo(...)`.

This makes it so the `.py` file is checked if it exists or if _neither_ it nor the `.pyc` file exists (so as to preserve the current behavior when both are missing).

### DIFF
--- a/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
@@ -57,13 +57,13 @@ class Python2(ViaGlobalRefVirtualenvBuiltin, Python2Supports):
             yield src
         # install files needed to run site.py
         for req in cls.modules():
-            stdlib_path = interpreter.stdlib_path("{}.py".format(req))
-            comp = Path(stdlib_path.stem + ".pyc")
+            comp = interpreter.stdlib_path("{}.pyc".format(req))
             comp_exists = comp.exists()
-            if stdlib_path.exists() or not comp_exists:
-                yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
             if comp_exists:
                 yield PathRefToDest(comp, dest=cls.to_stdlib)
+            stdlib_path = interpreter.stdlib_path("{}.py".format(req))
+            if stdlib_path.exists() or not comp_exists:
+                yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
 
     def to_stdlib(self, src):
         return self.stdlib / src.name
--- a/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
@@ -58,9 +58,11 @@ class Python2(ViaGlobalRefVirtualenvBuiltin, Python2Supports):
         # install files needed to run site.py
         for req in cls.modules():
             stdlib_path = interpreter.stdlib_path("{}.py".format(req))
-            yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
-            comp = stdlib_path.parent / "{}.pyc".format(req)
-            if comp.exists():
+            comp = Path(stdlib_path.stem + ".pyc")
+            comp_exists = comp.exists()
+            if stdlib_path.exists() or not comp_exists:
+                yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
+            if comp_exists:
                 yield PathRefToDest(comp, dest=cls.to_stdlib)
 
     def to_stdlib(self, src):
--- a/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/python2.py
@@ -57,13 +57,27 @@ class Python2(ViaGlobalRefVirtualenvBuiltin, Python2Supports):
             yield src
         # install files needed to run site.py
         for req in cls.modules():
-            comp = interpreter.stdlib_path("{}.pyc".format(req))
-            comp_exists = comp.exists()
-            if comp_exists:
-                yield PathRefToDest(comp, dest=cls.to_stdlib)
-            stdlib_path = interpreter.stdlib_path("{}.py".format(req))
-            if stdlib_path.exists() or not comp_exists:
-                yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
+
+            # the compiled path is optional, but refer to it if exists
+            module_compiled_path = interpreter.stdlib_path("{}.pyc".format(req))
+            has_compile = module_compiled_path.exists()
+            if has_compile:
+                yield PathRefToDest(module_compiled_path, dest=cls.to_stdlib)
+
+            # stdlib module src may be missing if the interpreter allows it by falling back to the compiled
+            module_path = interpreter.stdlib_path("{}.py".format(req))
+            add_py_module = cls.needs_stdlib_py_module()
+            if add_py_module is False:
+                if module_path.exists():  # if present add it
+                    add_py_module = True
+                else:
+                    add_py_module = not has_compile  # otherwise only add it if the pyc is not present
+            if add_py_module:
+                yield PathRefToDest(module_path, dest=cls.to_stdlib)
+
+    @classmethod
+    def needs_stdlib_py_module(cls):
+        raise NotImplementedError
 
     def to_stdlib(self, src):
         return self.stdlib / src.name
--- a/docs/changelog/1714.bugfix.rst
+++ b/docs/changelog/1714.bugfix.rst
@@ -1,0 +1,1 @@
+Allow missing ``.py`` files if a compiled ``.pyc`` version is available - by :user:`tucked`.
--- a/docs/changelog/1714.doc.rst
+++ b/docs/changelog/1714.doc.rst
@@ -1,0 +1,2 @@
+:ref:`supports <compatibility-requirements>` details now explicitly what Python installations we support
+- by :user:`gaborbernat`.
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -122,6 +122,6 @@ Windows
 Packaging variants
 ~~~~~~~~~~~~~~~~~~
 - Normal variant (file structure as comes from `python.org <https://www.python.org/downloads/>`_).
-- We support system installations that do not contain the python files for the standard library if the respective
-  compiled files are present (e.g. only ``os.pyc``, not ``os.py``) - relevant mostly for ``Python 2.7``. This can be
-  used by custom systems may want to maximize available storage or obfuscate source code by removing ``.py`` files.
+- We support CPython 2 system installations that do not contain the python files for the standard library if the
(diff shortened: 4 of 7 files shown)


src/virtualenv/create/via_global_ref/builtin/python2/python2.py
```diff
@@ -58,9 +58,11 @@ def sources(cls, interpreter):
         # install files needed to run site.py
         for req in cls.modules():
             stdlib_path = interpreter.stdlib_path("{}.py".format(req))
-            yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
             comp = stdlib_path.parent / "{}.pyc".format(req)
-            if comp.exists():
+            comp_exists = comp.exists()
+            if stdlib_path.exists() or not comp_exists:
```
cemeyer: Why `or not comp_exists`?  In that case, we will generate a broken reference to a file that does not exist, much like the issue this patch (otherwise) addresses.

```
if stdlib_path.exists():
    yield ...
if comp.exists():
    yield ...
```

Seems more correct to me.

If both paths are absent, it may also be worth throwing some kind of exception?  Presumably we cannot possibly succeed without the `req`uirements, so we should error out as soon as we know that.

tucked: The motivation was to preserve the current behavior when both are missing, mostly because I wasn't sure what to do otherwise. I agree that just checking for each is cleaner.

gaborbernat:
Can you describe a use case for this? Why would anyone want this?

tucked:
> Can you describe a use case for this? Why would anyone want this?

Custom systems may want to maximize available storage or obfuscate source code by removing `.py` files.

gaborbernat:
Did this worked with old virtualenv or would this be a new feature here 🤔

tucked:
> Did this worked with old virtualenv or would this be a new feature here 🤔

I cannot reproduce the issue with `virtualenv<20`.

gaborbernat:
I don't think this is an issue but feature 🤔 did it work to create virtual environments with virtualenv 16.x when you did not have the source file? I don't think so.

tucked:
> I don't think this is an issue but feature 🤔 did it work to create virtual environments with virtualenv 16.x when you did not have the source file? I don't think so.

Yes, if I downgrade to `virtualenv<20` on the same system I repro'd this on, I can create a virtualenv.

gaborbernat:
I'm surprised. Can you post a tree view of the result with old virtualenv and the output of -vvv 😲

tucked:
> I'm surprised. Can you post a tree view of the result with old virtualenv and the output of -vvv 😲

How's this (with `virtualenv-16.7.10`)?
- http://ix.io/2dZQ (`-vvv`)
- http://ix.io/2dZR (`ls -lhR` -- Sorry, I don't have `tree` 😢 I'll add an edit if I can get it.)

tucked:
I'm having a tough time making a test for this because writes to `Path(interpreter.system_stdlib)` are usually restricted 🤔 
``` diff
diff --git a/tests/unit/create/test_creator.py b/tests/unit/create/test_creator.py
index 70e0e6a..4091689 100644
--- a/tests/unit/create/test_creator.py
+++ b/tests/unit/create/test_creator.py
@@ -19,6 +19,7 @@ import pytest

 from virtualenv.__main__ import run, run_with_catch
 from virtualenv.create.creator import DEBUG_SCRIPT, Creator, get_env_debug_info
+from virtualenv.create.via_global_ref.builtin.cpython.cpython2 import CPython2
 from virtualenv.discovery.builtin import get_interpreter
 from virtualenv.discovery.py_info import PythonInfo
 from virtualenv.info import IS_PYPY, IS_WIN, PY3, fs_is_case_sensitive, fs_supports_symlink
@@ -448,3 +449,23 @@ def test_python_path(monkeypatch, tmp_path, python_path_on):
         assert non_python_path == [i for i in base if i not in extra_as_python_path]
     else:
         assert base == extra_all
+
+
+@pytest.fixture
+def creator_patched_modules(monkeypatch, current_creators):
+    creator = current_creators[2]
+    module_name = "CPython2_patch_modules_module"
+    module_path = Path(CURRENT.system_stdlib) / "{}.pyc".format(module_name)
+    module_path.touch()
+    @classmethod
+    def modules(cls):
+        return [module_name]
+    monkeypatch.setattr(CPython2, 'modules', modules)
+    yield creator
+    module_path.unlink()
+
+
+
+@pytest.mark.skipif(PY3, reason=".py files are only checked for in py2.")
+def test_pyc_only(creator_patched_modules):
+    assert creator_patched_modules.can_create(CURRENT) is not None
```
```
tests/unit/create/test_creator.py:459:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.tox/py27/lib/python2.7/site-packages/pathlib2/__init__.py:1524: in touch
    fd = self._raw_open(flags, mode)
.tox/py27/lib/python2.7/site-packages/pathlib2/__init__.py:1305: in _raw_open
    return self._accessor.open(self, flags, mode)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

pathobj = PosixPath('/usr/lib/python2.7/CPython2_patch_modules_module.pyc'), args = (65, 438)

    @functools.wraps(strfunc)
    def wrapped(pathobj, *args):
>       return strfunc(str(pathobj), *args)
E       OSError: [Errno 13] Permission denied: '/usr/lib/python2.7/CPython2_patch_modules_module.pyc'

args       = (65, 438)
pathobj    = PosixPath('/usr/lib/python2.7/CPython2_patch_modules_module.pyc')
strfunc    = <built-in function open>

.tox/py27/lib/python2.7/site-packages/pathlib2/__init__.py:585: OSError
```

src/virtualenv/create/via_global_ref/builtin/python2/python2.py
gaborbernat: I agree if neither exists we should yield the non compiled version to trigger an error 👍

gaborbernat:
As a test I'd recommend building a mock python by copying the system files without the py files and running on that 👍

src/virtualenv/create/via_global_ref/builtin/python2/python2.py
cemeyer: Yielding a file that doesn't exist results in a relatively inscrutable and confusing error message.  Why not just raise an exception here, so it is clear what the fault is?

gaborbernat: The world is more nuanced than that. The fact that this creator cannot create a virtual environment does not mean that there is not another registered that can. Think e.g. on Python 3 our builtin logic might not be able to create a venv, but the venv creator can by using the interpreters' builtin venv mechanism. Granted if no other creators are found we should probably display an error with the ones that failed due to source invariants could not be satisfied (e.g. missing, permissions, etc). But that's a feature improvement at the creator discovery level, not here.

ngie-eign:
Dumb question: why isn't this logic using `importlib`, `__import__`, or something similar with a stripped down `sys.path`, then scrutinizing paths for modules it finds?

By the way, the current code is technically code broken with .pyo-only files as well as compressed eggs.

src/virtualenv/create/via_global_ref/builtin/python2/python2.py
```diff
@@ -58,9 +58,11 @@ def sources(cls, interpreter):
...
+                yield PathRefToDest(stdlib_path, dest=cls.to_stdlib)
+            if comp_exists:
```
ngie-eign: Should this be an `elif`?

tucked: Perhaps... The current behavior is to `yield` `.py` _and_ `.pyc` (if it exists), though, so I opted to preserve that.

gaborbernat:
It's because the python executing the code here is not the same as the target python. Importing it here would not give you the modules you're expecting. Those modes are not supported by virtualenv, and would require case by case judgment if we should support them or not.

gaborbernat:
I'm fairly sure compressed eggs are not supported by the ecosystem, e.g. setuptools anymore. They work for backwards compatibility reasons but they should not be used.

src/virtualenv/create/via_global_ref/builtin/python2/python2.py
gaborbernat: No, this should be two separate ifs. If the pyc exists copy it so the first load doesn't have to regenerate it.

src/virtualenv/create/via_global_ref/builtin/python2/python2.py
gaborbernat: Created https://github.com/pypa/virtualenv/issues/1716 to track the above improvement request.

tucked:
> As a test I'd recommend building a mock python by copying the system files without the py files and running on that 👍

I'm not entirely sure how to do this 😅 I was able to get something working, though.
I believe the CI failures are unrelated. (Note: `test_pyc_only` passed in the timed out `pypy macOs` run.)

gaborbernat:
@tucked see https://github.com/pypa/virtualenv/pull/1714/commits/9742921eb2b2cecde0acfa876e704fcd28e13d6b for a more accurate test

tucked:
> @tucked see 9742921 for a more accurate test

Awesome 😎 Thanks for that!
Looks like we're _almost_ there... At least [one of the CI failures](https://dev.azure.com/pypa/virtualenv/_build/results?buildId=20182&view=logs&j=6cd38c3c-3a17-5dab-f848-d6174e69d150&t=b83ea89c-83a1-58b5-981d-aabcf04faede&l=265) looks relevant:
```
>       assert "os.pyc" in result.creator.debug["os"]
E       KeyError: u'os'
```

gaborbernat:
Yeah, pushed a fixed commit just now, seems PyPy does not support not having the source file for a standard library module.

gaborbernat:
I'll do a release tomorrow 👍

gaborbernat:
Hello, a fix for this issue has been released via virtualenv 20.0.11; see https://pypi.org/project/virtualenv/20.0.11/ (https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-11-2020-03-18). Please give a try and report back if your issue has not been addressed; if not, please comment here, and we'll reopen the ticket. We want to apologize for the inconvenience this has caused you and say thanks for having patience while we resolve the unexpected bugs with this new major release.
thanks

tucked:
@gaborbernat Thank YOU for all your help with this 😊 No worries

I suspect this is fixed; however, I'm unable to verify due to #1738!

Issue #1723, opened 2020-03-15
pipx.y missing, instead pip-x.y present


earthgecko:
**Issue**

In virtualenv 20.0.10 at least when a virtualenv environment is created the naming convention of the pip major version file has changed.

In 16.7.x there was `pip3.7` in 20.0.2 this has changed to `pip-3.7` although there is still `pip3` in both, in terms of automation patterns that provision multiple environments and builds there may be many build dependencies that use the `pip${MAJOR_VERSION}` pattern.

Perhaps adding a symlink to honour the older naming convention may prevent many people spending many hours debugging.

**Environment**

- OS: CentOS 8
- OS Python: Python 3.6.8
- python interpretor being used for the virtualenv is Python 3.7.6 compiled from source.
- System pip list

```
pip list --format=columns

Package             Version
------------------- -------
appdirs             1.4.3
asn1crypto          0.24.0
cffi                1.11.5
configobj           5.0.6
cryptography        2.3
dbus-python         1.2.4
decorator           4.2.1
distlib             0.3.0
filelock            3.0.12
gpg                 1.10.0
idna                2.5
importlib-metadata  1.5.0
importlib-resources 1.3.1
iniparse            0.4
isc                 2.0
netifaces           0.10.6
pciutils            2.3.6
perf                0.1
pip                 9.0.3
ply                 3.9
pycparser           2.14
pygobject           3.28.3
pyOpenSSL           18.0.0
pyparsing           2.1.10
python-dateutil     2.6.1
python-dmidecode    3.12.2
python-linux-procfs 0.6
pyudev              0.21.0
rhnlib              2.8.6
rpm                 4.14.2
schedutils          0.6
selinux             2.9
sepolicy            1.1
setools             4.2.2
setuptools          39.2.0
six                 1.11.0
slip                0.6.4
slip.dbus           0.6.4
syspurpose          1.25.17
zipp                3.1.0
```

**Reproduce**

```
cd /opt
pip install virtualenv==20.0.10

pip list --format=columns | grep virtualenv
virtualenv          20.0.10

virtualenv --python=/opt/python_virtualenv/versions/3.7.6/bin/python3.7 venv_20.0.10

ls -al /opt/venv_20.0.10/bin | grep pip
-rwxr-xr-x. 1 root root  233 Mar 15 14:08 pip
-rwxr-xr-x. 1 root root  233 Mar 15 14:08 pip3
[...]

PR #1729, opened 2020-03-18
generate pipx.y besides pip-x.y

gaborbernat:
Resolves https://github.com/pypa/virtualenv/issues/1723


src/virtualenv/seed/via_app_data/pip_install/base.py
```diff
@@ -157,6 +157,8 @@ def _write_script(scriptnames, shebang, script, filenames, ext):
                 "{}{}".format(name, version_info.major),
                 "{}-{}.{}".format(name, version_info.major, version_info.minor),
```
pradyunsg: Why are we generating "scriptname-X.Y" scripts?

gaborbernat: Because that's what distlib does, and the idea is that the virtualenv should shield you from calling os versions of these 👍

PR #1730, opened 2020-03-18
Fix cross-major creation on Windows of activate_this.py throws error

### LINKED ISSUE #1710: Broken activation in Windows for python3
virtualenv activation through `activate_this.py` is broken in Windows for python3. Check this very simple piece of code

```
import os
import subprocess
import sys
from distutils.spawn import find_executable

venv = find_executable("virtualenv")
testdir = os.path.join(os.path.curdir, 'testenv')

subprocess.check_output((venv, testdir, "-p", sys.executable))

bin_path = os.path.join(testdir, "Scripts") if sys.platform in ("win32", "cygwin") else os.path.join(testdir, "bin")

path = os.path.join(bin_path, "activate_this.py")
with open(path) as f:
    exec(f.read(), {"__file__": path})
```

This generates a `AttributeError: 'str' object has no attribute 'decode'`. Taking a look at the `activate_this.py` code:

```
for lib in "..\Lib\site-packages".split(os.pathsep):
    path = os.path.realpath(os.path.join(bin_dir, lib))
    site.addsitedir(path.decode("utf-8") if "yes" else path)
```

it's indeed normal that we get the error because we're always calling `decode()` over a str. Question is where the `__DECODE_PATH__` from https://github.com/pypa/virtualenv/blob/master/src/virtualenv/activation/python/activate_this.py#L28 is assigned to `"yes"`?

gaborbernat:
Resolves #1710

### DIFF
--- a/docs/changelog/1710.bugfix.rst
+++ b/docs/changelog/1710.bugfix.rst
@@ -1,0 +1,2 @@
+Fix ``activate_this.py`` throws ``AttributeError`` on Windows when virtual environment was created via cross python
+mechanism - by :user:`gaborbernat`.
--- a/src/virtualenv/activation/python/__init__.py
+++ b/src/virtualenv/activation/python/__init__.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, unicode_literals
 import os
 from collections import OrderedDict
 
-from virtualenv.info import WIN_CPYTHON_2
 from virtualenv.util.path import Path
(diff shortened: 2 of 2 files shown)


svillar:
Any idea about the first release containing this fix?

gaborbernat:
20.0.11 will go out later today if all goes by plans 👍 or at latest tomorrow, will come back and tell once done

PR #1745, opened 2020-03-23
upgrade embedded setuptools

gaborbernat:
to ``44.1.0`` and ``46.1.1``


gaborbernat:
released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-14-2020-03-25

PR #1762, opened 2020-04-06
Updated Azure Pipelines build shield to match other shields styles.

mondeja:


gaborbernat:
No thank you we'll keep the current representation directly from Azure.

PR #1793, opened 2020-04-28
Bump pip

gaborbernat:


gaborbernat:
Hello, this now has been released via https://pypi.org/project/virtualenv/20.0.19/

heads up

PR #1813, opened 2020-05-04
Fix download is True by default

gaborbernat:
Now part of https://github.com/pypa/virtualenv/pull/1814


gaborbernat:
Hello, this now has been released via https://pypi.org/project/virtualenv/20.0.20/

heads up

Issue #1806, opened 2020-05-03
Generate VCS ignores to avoid version controlling virtual environments


gaborbernat:
As per https://twitter.com/metalikus/status/1256853693153947649

List of VCSs targeted is https://www.python.org/dev/peps/pep-0610/#registered-vcs

SanketDG:
I would like to take this up, where would be this best placed?

https://github.com/pypa/virtualenv/blob/6dffcf2ae7b5b55098cf3c8326472f988277cb70/src/virtualenv/create/creator.py#L151-L156

Is something like `self.write_gitignore`/`self.ensure_gitignore` after `self.set_pyenv_cfg()` be a good way to implement this?

gaborbernat:
Yeah somewhere there 🤔

gaborbernat:
This now has been released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-21-2020-05-20

PR #1817, opened 2020-05-04
Ensure that the virtualenv dir is ignored from Git

SanketDG:
Fixes https://github.com/pypa/virtualenv/issues/1806

- [ ] ensured there are test(s) validating the fix (need help)

### DIFF
--- a/docs/changelog/1806.feature.rst
+++ b/docs/changelog/1806.feature.rst
@@ -1,0 +1,1 @@
+Ensure that the virtualenv dir created is ignored by Git
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -154,6 +154,15 @@ class Creator(object):
             safe_delete(self.dest)
         self.create()
         self.set_pyenv_cfg()
+        self._ensure_gitignore()
+
+    def _ensure_gitignore(self):
+        """
+        Ignore the virtualenv directory from git.
+        """
+        self.dest.joinpath(".gitignore").write_text(
+            "# placed by virtualenv automatically, do not change\n*", encoding="UTF-8"
+        )
 
     def set_pyenv_cfg(self):
         self.pyenv_cfg.content = OrderedDict()


SanketDG:
Question: Do we want this as a flag?

src/virtualenv/create/creator.py
```diff
@@ -154,6 +154,15 @@ def run(self):
             safe_delete(self.dest)
         self.create()
         self.set_pyenv_cfg()
+        self._ensure_gitignore()
+
+    def _ensure_gitignore(self):
```
gaborbernat, line 159: On second thought this needs to be within https://github.com/pypa/virtualenv/blob/master/src/virtualenv/create/via_global_ref/api.py#L89

src/virtualenv/create/creator.py
```diff
@@ -154,6 +154,15 @@ def run(self):
...
+        """
+        Ignore the virtualenv directory from git.
+        """
+        self.dest.joinpath(".gitignore").write_text(
```
gaborbernat, line 163: use the ``/`` instead of join path

gaborbernat: the utf-8 encoding is not needed can leave it empty to use platform default;

the comment is not needed, please remove

docs/changelog/1806.feature.rst
```diff
@@ -0,0 +1 @@
+Ensure that the virtualenv dir created is ignored by Git
```
gaborbernat, line 1: This does not follow the example structure, e.g. missing user

src/virtualenv/create/creator.py
gaborbernat, line 159: let's not make it git specific, let's cover all VCS covered here https://www.python.org/dev/peps/pep-0610/#registered-vcs

a better name is: ``generate_vcs_ignore``, no need to make it private

gaborbernat: And yeah please add an option to disable this via ``--no-vcs-ingore`` 👍

gaborbernat:
Superseded by https://github.com/pypa/virtualenv/pull/1825

Issue #1801, opened 2020-05-01
Revamp Docs


SanketDG:
I have a proposal to revamp the current documentation that stays at https://virtualenv.pypa.io/en/latest/index.html

### Planned Layout

- Installation
- Quick Start
- User Guide
- Reference Guide
- Plugin API
- Contributing
- Release Notes

### Front Page

The front page should not be a list of things about what we do different from `venv` but instead just make a subtle note that `virtualenv` is different from inbuilt `venv` and link to a more exhaustive list of differences that is placed later in the documentation.

Related projects - I am not sure how relevant `pew` still is (also last commit is in 2019)

Should we keep the tutorials? - The `mod_wsgi` seems pretty irrelevant on the first page.

@gaborbernat's talk is helpful here espescially with the rewrite, I am not sure how the other one is, given its been quite some years since the talk.

### Installation

We should first and foremost make the distinction that this is different from `venv` and to the link to the said differences (as we did on the front page)

`pip` is ubiquitous, `pipx` is not, it should still be the recommended form of installation.

The rest of the content just needs to be brushed up a little.

---

I am going to propose required changes of each part of the docs with a comment here, so that folks don't get overwhelmed.

gaborbernat:
I don't agree on the pipx point, on the rest we can come to an agreement.

SanketDG:
From gitter:

> @gaborbernat I have started working on the docs at https://github.com/SanketDG/virtualenv/tree/sanketdg/revamp-docs
I have done a preliminary amount of work and I will be waiting until we found a scalable model for undergoing a full documentation rewrite
My naive idea is to create a upstream branch where I can send periodic PRs which can be reviewed, and at the end we merge the whole branch into master
Let me know what your thoughts are on this.

gaborbernat:
Thanks will check, send it as a pr?

gaborbernat:
Seems no longer an issue as no one reported it in 3 years.

PR #1820, opened 2020-05-06
WIP: Revamp Docs: Rewrite Index Page

SanketDG:

This is some preliminary work towards https://github.com/pypa/virtualenv/issues/1801. Since this will completely break the current docs, I am not sure how should I go about this. I would still advise you to consider the workflow I have suggested in https://github.com/pypa/virtualenv/issues/1801#issuecomment-624678532

### DIFF
--- a/docs/differences_with_venv.rst
+++ b/docs/differences_with_venv.rst
@@ -1,7 +1,17 @@
 Differences with venv
 =====================
-- is slower (by not having the app-data seed method),
-- is not as extendable,
-- cannot create virtual environments for arbitrarily installed python versions (and automatically discover these),
-- is not upgrade-able via pip,
-- does not have as rich programmatic API (describe virtual environments without creating them).
+
+- ``venv`` is bundled with Python itself (starting from 3.3+), ``virtualenv``
+  is a third party package. This allows for ``virtualenv`` to have greater
+  flexibility when dealing with several Python versions.
+- ``virtualenv`` allows further :doc:`extendibility <extend>` to override default
+  functionality through a plugin system.
+- ``virtualenv`` can create virtual environments through its interpreter discovery
+  technique from arbitrary Python locations, they do not be installed or be on the
+  ``PATH``.
+- ``venv`` can only create virtual environments for the Python version it is
+  installed with, ``virtualenv`` supports both Python 2 and 3 and also other flavours
+  like ``CPython`` and ``PyPy``.
+- ``virtualenv`` has a rich programmatic API (describe virtual environments without
+  creating them).
+- ``virtualenv`` has better support across different platforms and shells.
--- a/docs/differences_with_venv.rst
+++ b/docs/differences_with_venv.rst
@@ -1,0 +1,7 @@
+Differences with venv
+=====================
+- is slower (by not having the app-data seed method),
+- is not as extendable,
+- cannot create virtual environments for arbitrarily installed python versions (and automatically discover these),
+- is not upgrade-able via pip,
+- does not have as rich programmatic API (describe virtual environments without creating them).
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,5 @@
-Virtualenv
-==========
+virtualenv - Isolated Python Environments
+=========================================
 
 .. image:: https://img.shields.io/pypi/v/virtualenv?style=flat-square
   :target: https://pypi.org/project/virtualenv/#history
@@ -30,17 +30,15 @@ Virtualenv
   :target: https://pypistats.org/packages/virtualenv
   :alt: Package popularity
 
-``virtualenv`` is a tool to create isolated Python environments. Since Python ``3.3``, a subset of it has been
-integrated into the standard library under the `venv module <https://docs.python.org/3/library/venv.html>`_. The
-``venv`` module does not offer all features of this library, to name just a few more prominent:
+``virtualenv`` is a tool to create isolated Python environments with its own copy of
+a working Python environment that is isolated from the rest of the system.
 
-- is slower (by not having the ``app-data`` seed method),
-- is not as extendable,
-- cannot create virtual environments for arbitrarily installed python versions (and automatically discover these),
-- is not upgrade-able via `pip <https://pip.pypa.io/en/stable/installing/>`_,
-- does not have as rich programmatic API (describe virtual environments without creating them).
+.. note::
+  Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.python.org/3/library/venv.html>`_. The
+  ``venv`` module does not offer all features of this library. <Difference link here>
 
-The basic problem being addressed is one of dependencies and versions, and indirectly permissions.
+The need for isolation of python environments comes when there are two python
+applications with their own set of dependencies, python versions and permissions.
 Imagine you have an application that needs version ``1`` of ``LibFoo``, but another application requires version
 ``2``. How can you use both these libraries? If you install everything into your host python (e.g. ``python3.8``)
 it's easy to end up in a situation where two packages have conflicting requirements.
@@ -53,28 +51,10 @@ In all these cases, ``virtualenv`` can help you. It creates an environment that 
 that doesn't share libraries with other virtualenv environments (and optionally doesn't access the globally installed
 libraries either).
 
-Useful links
-------------
+Quick Start
+-----------
 
-**Related projects, that build abstractions on top of virtualenv**
-
-* :pypi:`virtualenvwrapper` - a useful set of scripts for creating and deleting virtual environments
-* :pypi:`pew` - provides a set of commands to manage multiple virtual environments
-* :pypi:`tox` - a generic virtualenv management and test automation command line tool, driven by a ``tox.ini``
-  configuration file
-* :pypi:`nox` - a tool that automates testing in multiple Python environments, similar to tox,
-  driven by a ``noxfile.py`` configuration file
-
-**Tutorials**
-
-* `Corey Schafer tutorial <https://www.youtube.com/watch?v=N5vscPTWKOk>`_ on how to use it
-* `Using virtualenv with mod_wsgi <http://code.google.com/p/modwsgi/wiki/VirtualEnvironments>`_
-
-**Presenting how the package works from within**
-
-* `Bernat Gabor: status quo of virtual environments <https://www.youtube.com/watch?v=o1Vue9CWRxU>`_
-* `Carl Meyer: Reverse-engineering Ian Bicking's brain: inside pip and virtualenv
-  <http://pyvideo.org/video/568/reverse-engineering-ian-bicking--39-s-brain--insi>`_
+TODO
 
 .. comment: split here
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -35,7 +35,7 @@ a working Python environment that is isolated from the rest of the system.
 
 .. note::
   Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.python.org/3/library/venv.html>`_. The
-  ``venv`` module does not offer all features of this library. <Difference link here>
+  ``venv`` module does not offer all features of this library. See the :doc:`differences <differences_with_venv>` betwwen virtualenv and venv.
 
 The need for isolation of python environments comes when there are two python
 applications with their own set of dependencies, python versions and permissions.
@@ -54,7 +54,7 @@ libraries either).
 Quick Start
 -----------
 
-TODO
+.. _differences: differences_with_venv
 
 .. comment: split here
 
@@ -67,3 +67,5 @@ TODO
    extend
    development
    changelog
+   differences_with_venv
+   useful_links
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -34,25 +34,25 @@ virtualenv - Isolated Python Environments
 a working Python environment that is isolated from the rest of the system.
 
 .. note::
-  Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.python.org/3/library/venv.html>`_. The
+  Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.Python.org/3/library/venv.html>`_. The
   ``venv`` module does not offer all features of this library. See the :doc:`differences <differences_with_venv>` betwwen virtualenv and venv.
 
-The need for isolation of python environments comes when there are two python
-applications with their own set of dependencies, python versions and permissions.
-Imagine you have an application that needs version ``1`` of ``LibFoo``, but another application requires version
-``2``. How can you use both these libraries? If you install everything into your host python (e.g. ``python3.8``)
-it's easy to end up in a situation where two packages have conflicting requirements.
+The need for isolation of Python environments comes when there are two Python
+libraries with their own set of dependencies, Python versions and permissions.
+Python applications will often use external libraries that are not part of the standard Python
+library. The applications will often require a specific version  of that library where a particular
+bug has been fixed. Imagine you have an application that needs version ``1.2.3`` of ``foo``,
+but another application requires version ``2.3.4``. Installing one version of the library would
+make the other application unable to run. So, If you install everything into your system Python
+it's easy to end up in a situation where two applications have conflicting requirements.
 
 Or more generally, what if you want to install an application *and leave it be*? If an application works, any change
 in its libraries or the versions of those libraries can break the application. Also, what if you can't install packages
-into the global ``site-packages`` directory, due to not having permissions to change the host python environment?
+into the global ``site-packages`` directory, due to not having permissions to change the system Python environment?
 
 In all these cases, ``virtualenv`` can help you. It creates an environment that has its own installation directories,
 that doesn't share libraries with other virtualenv environments (and optionally doesn't access the globally installed
 libraries either).
-
-Quick Start
------------
 
 .. _differences: differences_with_venv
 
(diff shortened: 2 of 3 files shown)


docs/index.rst
```diff
@@ -30,17 +30,15 @@ Virtualenv
...
+a working Python environment that is isolated from the rest of the system.
 
-- is slower (by not having the ``app-data`` seed method),
-- is not as extendable,
-- cannot create virtual environments for arbitrarily installed python versions (and automatically discover these),
-- is not upgrade-able via `pip <https://pip.pypa.io/en/stable/installing/>`_,
-- does not have as rich programmatic API (describe virtual environments without creating them).
+.. note::
+  Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.python.org/3/library/venv.html>`_. The
+  ``venv`` module does not offer all features of this library. <Difference link here>
```
gaborbernat, line 38: do add the difference link

docs/index.rst
```diff
@@ -30,17 +30,15 @@ Virtualenv
...
 
-The basic problem being addressed is one of dependencies and versions, and indirectly permissions.
+The need for isolation of python environments comes when there are two python
```
gaborbernat, line 40: I liked better this before. We're not just talking here about applications. For example when I'm testing a library that's not an application, but I still need an isolated python environment, not?

SanketDG: @gaborbernat Absolutely, I agree. I put in application because that's how it's present in the rest of the paragraph below:
> magine you have an application that needs version ``1`` of ``LibFoo``, but another application requires version
``2``. How can you use both these libraries? If you install everything into your host python (e.g. ``python3.8``)
it's easy to end up in a situation where two packages have conflicting requirements.

Should I rephrase this too?

SanketDG: I rewrote this loosely inspired from https://docs.python.org/3/tutorial/venv.html#introduction

docs/index.rst
SanketDG: Done in the latest iteration.

docs/index.rst
```diff
@@ -30,51 +30,64 @@ Virtualenv
...
 
 In all these cases, ``virtualenv`` can help you. It creates an environment that has its own installation directories,
 that doesn't share libraries with other virtualenv environments (and optionally doesn't access the globally installed
 libraries either).
 
-Useful links
-------------
+Quick Start
+-----------
+The recommended way to install virtualenv is through pip: (for other methods see :doc:`Installation <installation>`)
```
gaborbernat, line 59: I strongly disagree with this. installing virtualenv with pip causes all kind of issues I'm not comfortable recommending.

docs/index.rst
```diff
@@ -30,51 +30,64 @@ Virtualenv
...
+.. code-block:: console
 
-* :pypi:`virtualenvwrapper` - a useful set of scripts for creating and deleting virtual environments
-* :pypi:`pew` - provides a set of commands to manage multiple virtual environments
-* :pypi:`tox` - a generic virtualenv management and test automation command line tool, driven by a ``tox.ini``
-  configuration file
-* :pypi:`nox` - a tool that automates testing in multiple Python environments, similar to tox,
-  driven by a ``noxfile.py`` configuration file
+    $ Python -m pip install --user virtualenv
+    $ virtualenv --version
```
gaborbernat, line 64: Not sure what's the point of this. Also probably ``python -m virtualenv is the sane way to access this`` as on Windows the installed bin folder path might not be on the path.

docs/index.rst
```diff
@@ -30,51 +30,64 @@ Virtualenv
...
 
-**Tutorials**
+To create a virtualenv using the default Python instance available on the system
```
gaborbernat, line 66: Well not sure what qualifies default python. It's more like the python under what virtalenv is installed.

docs/index.rst
```diff
@@ -30,51 +30,64 @@ Virtualenv
...
 
-**Presenting how the package works from within**
+    $ virtualenv venv/
 
-* `Bernat Gabor: status quo of virtual environments <https://www.youtube.com/watch?v=o1Vue9CWRxU>`_
-* `Carl Meyer: Reverse-engineering Ian Bicking's brain: inside pip and virtualenv
-  <http://pyvideo.org/video/568/reverse-engineering-ian-bicking--39-s-brain--insi>`_
+To activate a virtualenv
+
+Bash
```
gaborbernat, line 74: I'd not enumerate any of these. Makes a lot of assumption that the user's shell is bash, often it's not.

docs/index.rst
```diff
@@ -30,51 +30,64 @@ Virtualenv
...
+.. code-block:: console
+
+    $ source venv/bin/activate
+
+Windows
+^^^^^^^
+
+.. code-block:: console
+
+    venv/bin/activate.bat
```
gaborbernat, line 86: Also no for this. Assumes user runs with command prompt, but that's not the case with git bash or PowerShell.

docs/index.rst
SanketDG: I will replace with `pipx` then, thanks (y)

gaborbernat: I don't think needs replaced, needs to be removed. Maybe a link to the install section but no need to point out any method in particular. The user might even pull it in from its operating system package manager, it should not matter.

docs/index.rst
SanketDG: The idea is to target the most commonly used shells/platforms. But I understand.

A list of exhaustive activators are not present in the documentation.

Do you want me to create that? Is there an alternative way that you have in mind for specifying the activators in a platform agnostic way?

gaborbernat: Mainly because it's not set in stone users might create more via https://virtualenv.pypa.io/en/latest/extend.html#python-discovery. A default shell targeted is documented under https://virtualenv.pypa.io/en/latest/cli_interface.html#section-activators. Not sure what more you would want. https://virtualenv.pypa.io/en/latest/user_guide.html#activators details the contract.

docs/index.rst
SanketDG: I wanted to write a "Quick Start" and a "Tutorial" mostly aimed at beginners, who would come to the docs. Since the installation and activation part is best delegated to its original documentation (which is a little verbose IMO), do you want me to remove this section entirely?

gaborbernat: I'd say yes.

SanketDG: Sure, then what would you suggest would be the best way to have a guide/tutorial that is well suited for beginners, and doesn't have to go through pages of documentation to figure out how to use `virtualenv`

SanketDG: I have removed the Quick Start section

docs/index.rst
SanketDG: I have removed the Quick Start section

docs/index.rst
SanketDG: I have removed the Quick Start section

docs/index.rst
SanketDG: I have removed the Quick Start section

docs/index.rst
SanketDG: I have removed the Quick Start section

docs/useful_links.rst
```diff
@@ -0,0 +1,19 @@
+Useful links
+============
+
+**Related projects, that build abstractions on top of virtualenv**
+
+* `virtualenvwrapper <https://virtualenvwrapper.readthedocs.io/en/latest/>`_ - a useful set of scripts for creating and deleting virtual environments
```
gaborbernat, line 6: Please restore these links to pypi. The pypi contains links to the documentation, code base etc.

gaborbernat: Add back pew.

docs/useful_links.rst
```diff
@@ -0,0 +1,19 @@
...
+  driven by a ``noxfile.py`` configuration file
+
+**Tutorials**
+
+* `Corey Schafer tutorial <https://www.youtube.com/watch?v=N5vscPTWKOk>`_ on how to use it
+* `Using virtualenv with mod_wsgi <http://code.google.com/p/modwsgi/wiki/VirtualEnvironments>`_
+
+**Presenting how the package works from within**
+
+* `Bernat Gabor: status quo of virtual environments <https://www.youtube.com/watch?v=o1Vue9CWRxU>`_
```
gaborbernat, line 19: Add back Carl Meyers video; it's an interesting case study on how things were in the past.

docs/index.rst
```diff
@@ -30,51 +30,31 @@ Virtualenv
...
+library. The applications will often require a specific version  of that library where a particular
+bug has been fixed. Imagine you have an application that needs version ``1.2.3`` of ``foo``,
+but another application requires version ``2.3.4``. Installing one version of the library would
+make the other application unable to run. So, If you install everything into your system Python
+it's easy to end up in a situation where two applications have conflicting requirements.
 
 Or more generally, what if you want to install an application *and leave it be*? If an application works, any change
 in its libraries or the versions of those libraries can break the application. Also, what if you can't install packages
-into the global ``site-packages`` directory, due to not having permissions to change the host python environment?
+into the global ``site-packages`` directory, due to not having permissions to change the system Python environment?
```
gaborbernat, line 51: The term host is accurate, not the system. Doesn't have to be system python install; it can be side-load installed too.

docs/index.rst
```diff
@@ -30,51 +30,31 @@ Virtualenv
...
-- does not have as rich programmatic API (describe virtual environments without creating them).
+.. note::
+  Since Python ``3.3``, a subset of it has been integrated into the standard library under the `venv module <https://docs.Python.org/3/library/venv.html>`_. The
+  ``venv`` module does not offer all features of this library. See the :doc:`differences <differences_with_venv>` betwwen virtualenv and venv.
 
-The basic problem being addressed is one of dependencies and versions, and indirectly permissions.
-Imagine you have an application that needs version ``1`` of ``LibFoo``, but another application requires version
-``2``. How can you use both these libraries? If you install everything into your host python (e.g. ``python3.8``)
-it's easy to end up in a situation where two packages have conflicting requirements.
+The need for isolation of Python environments comes when there are two Python
```
gaborbernat, line 40: I prefer the initial formulation. I find the new sentences hard to follow and inaccurate in some places.

docs/differences_with_venv.rst
```diff
@@ -0,0 +1,17 @@
+Differences with venv
+=====================
+
+- ``venv`` is bundled with Python itself (starting from 3.3+), ``virtualenv``
+  is a third party package. This allows for ``virtualenv`` to have greater
+  flexibility when dealing with several Python versions.
```
gaborbernat, line 6: I don't think the flexibility comes from the fact that it's bundled.

docs/differences_with_venv.rst
```diff
@@ -0,0 +1,17 @@
...
+- ``virtualenv`` allows further :doc:`extendibility <extend>` to override default
+  functionality through a plugin system.
+- ``virtualenv`` can create virtual environments through its interpreter discovery
+  technique from arbitrary Python locations, they do not be installed or be on the
```
gaborbernat, line 10: This is not true.

docs/differences_with_venv.rst
```diff
@@ -0,0 +1,17 @@
...
+  ``PATH``.
+- ``venv`` can only create virtual environments for the Python version it is
+  installed with, ``virtualenv`` supports both Python 2 and 3 and also other flavours
+  like ``CPython`` and ``PyPy``.
+- ``virtualenv`` has a rich programmatic API (describe virtual environments without
+  creating them).
+- ``virtualenv`` has better support across different platforms and shells.
```
gaborbernat, line 17: This is not true at platform level; shells maybe, but that is not the point here. venv adds support for these extra shells... but they take time to ship.

docs/differences_with_venv.rst
```diff
@@ -0,0 +1,17 @@
...
+- ``venv`` is bundled with Python itself (starting from 3.3+), ``virtualenv``
+  is a third party package. This allows for ``virtualenv`` to have greater
+  flexibility when dealing with several Python versions.
+- ``virtualenv`` allows further :doc:`extendibility <extend>` to override default
+  functionality through a plugin system.
+- ``virtualenv`` can create virtual environments through its interpreter discovery
+  technique from arbitrary Python locations, they do not be installed or be on the
+  ``PATH``.
+- ``venv`` can only create virtual environments for the Python version it is
+  installed with, ``virtualenv`` supports both Python 2 and 3 and also other flavours
```
gaborbernat, line 13: You mention other flavours, but then you point out two we support. The other implies you already mentioned one and you're including new ones.

docs/differences_with_venv.rst
```diff
@@ -0,0 +1,17 @@
+Differences with venv
+=====================
+
+- ``venv`` is bundled with Python itself (starting from 3.3+), ``virtualenv``
```
gaborbernat, line 4: You removed the performance entry which is the main reason one would pick this.

docs/useful_links.rst
SanketDG: I had initially written the reasonings for this in my initial spec. Apologies, I thought you had approved those.

> Please restore these links to pypi. The pypi contains links to the documentation, code base etc.

PyPI is not an interface that a lot of people visit everyday.

> Add back pew.

`pew`'s last release/commit is from 2019. How is it relevant?

docs/useful_links.rst
SanketDG: Given the video's quality and age, I am much more inclined to write a document how things were before in the past.

docs/differences_with_venv.rst
SanketDG: This comes naively from your presentation.

https://gaborbernat.github.io/euro_python_2019/#/

docs/differences_with_venv.rst
SanketDG: This also comes naively from your presentation https://gaborbernat.github.io/euro_python_2019/#/

But I could be wrong on the "installation" part.

docs/differences_with_venv.rst
SanketDG: It defines no reason for said performance.

And that performance only comes into play if someone uses the `app-seed` feature, right?

docs/differences_with_venv.rst
SanketDG: I don't understand, it's not bundled, that's why it has greater flexibility?

I also deduced this from your resentation, given there are multiple installation options and support for any Python version.

docs/differences_with_venv.rst
SanketDG: I will rephrase this.

docs/index.rst
SanketDG: This is adapted from https://docs.python.org/3/library/venv.html, and it seems readable to me.

docs/index.rst
SanketDG: `host` doesn't mean intuitive at all, and so doesn't `system`.

I will revert the change.

docs/useful_links.rst
SanketDG: I had described this in my initial spec on the revamp issue

> Please restore these links to pypi. The pypi contains links to the documentation, code base etc.

PyPI isn't a interface that is visited a lot, I am not sure why anything else over the official docs is preferred here



> Add back pew.

Pew's last commit/release was in 2019, I don't see how it's relevant. There are probably a bunch of other tools that are used in the ecosystem today that deserves a name here

docs/useful_links.rst
SanketDG: Given the quality and the age of the video, I'd be inclined to write a document on how things were in the past.

SanketDG:
@gaborbernat 

Apologies if this has been less convincing. I have replied to all the comments above.

Most of the points here were adapted from your talk in EuroPython, but I understand me being not familiar with the system has caused a lot of changes to not make sense.

I will close this PR if these changes do not align with your vision and thought process.

docs/useful_links.rst
gaborbernat: Not everything needs a new release to be relevant. pew is even more popular than nox, so should we remove that too; see https://pepy.tech/project/nox vs https://pepy.tech/project/pew

docs/index.rst
gaborbernat: We'll have to agree to disagree on this then 😊

docs/differences_with_venv.rst
gaborbernat: It's what most visible to end-users though and is on by default.

gaborbernat:
@SanketDG maybe this type of changes are not effective via PR. Feels this could be fleshed out much quicker in a sprint (face-to-face discussion). This way feels arbitrary what gets removed, where and how; so we're going to have a lot of back and forth on it; as IMHO some of them are a matter of taste. 

> Let's step back and understand what you try to achieve here? Is it to make it more accessible to someone totally new to Python? If so maybe we just need to add a single page that addresses the need. Something like `High level introduction` and mention on the front page if you want a high level why and how to use, read this page. For who is interested in the details read other sections.

Still waiting for an answer on this 👍

SanketDG:
> Still waiting for an answer on this

There definitely seems to be a place for this.

How do you want me to go about it?

gaborbernat:
Instead of trying to rewrite parts, just add a new beginner friendly page?

SanketDG:
> Instead of trying to rewrite parts, just add a new beginner friendly page?

I can do that, but as you said a little bit of back and forth discussion would help here (would gitter be the right thing here?)

Is there anything relevant in this PR that you wish to keep? I will otherwise close it then.

gaborbernat:
I think let's close it.

gaborbernat:
> I can do that, but as you said a little bit of back and forth discussion would help here (would gitter be the right thing for this?)

If you keep it as an addition of new document rather than changing existing this might be less disruptive, and you can start with a draft proposal.

PR #1827, opened 2020-05-13
Upgrade embeded setuptools to 46.3.0

gaborbernat:


gaborbernat:
This now has been released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-21-2020-05-20

Issue #1821, opened 2020-05-08
Ensure bundled pip does not get too out of date


dstufft:
**Issue**

Prior to the rewrite of virtualenv, it would ensure to always install the latest version of pip. If I understand correctly, with the rewrite this behavior has regressed back to installing ONLY the version of pip bundled with virtualenv.

The health of the packaging ecosystem depends in part by getting users onto newer versions of the packaging tools as quickly as possible. The decision to install the latest version of pip in virtualenv was made in part to support that. By reverting this feature back to it's legacy behavior we're making it harder to progress python packaging as a whole.

From a user point of view, before we enabled download by default a common complaint we had was that people would create a virtual environment and the first time they run a pip command they're told to upgrade pip. A number of users reported that the very first thing they did upon entering a new virtual env was upgrade pip which slowed things down because they had to uninstall and upgrade over the pip that was just installed.

Users who wish to have reproducibility were in the minority and It is, in my opinion, a fairly major regression in virtualenv and I would stongly suggest bringing back the intended behavior.

gaborbernat:
> Users who wish to have reproducibility were in the minority

I have not seen any research to back up this claim, so I'll consider this as your opinion.

> The health of the packaging ecosystem depends in part by getting users onto newer versions of the packaging tools as quickly as possible.

With PEP-517 introduced a few years back build backends and its dependencies are handled by the build frontend so the sole beneficiary of an auto-upgrade feature would be the front end, in this case, pip. pip A major reason we stopped download by default is that slows down the creation process significantly, especially on slow networks. Pip releases a new version every four-month, there really is no reason to slow it down every time (pip should IMHO not check to upgrade every time, but instead something sane like every three months and maybe warn for a day, and then give up). To make sure people use the latest pip it's easier to tell them to upgrade the virtualenv then force pip upgrade every time. We do repackage virtualenv shortly after pip does (giving them a few days after release to allow pip bugs to be fixed).

> If I understand correctly, with the rewrite this behavior has regressed back to installing ONLY the version of pip bundled with virtualenv.

This was no regression. It was a conscious decision made, and one we changed with the freedom of a new major version that allowed us to redefine defaults, without considering them regression or breaking.

pfmoore:
> It was a conscious decision made, and one we changed with the freedom of a new major version

[...]

PR #1831, opened 2020-05-18
App data seeder: make pip by default periodically self-upgrade

gaborbernat:
A conservative implementation of the outcome from https://discuss.python.org/t/how-should-virtualenv-behave-respective-of-seed-packages-e-g-pip-by-default/4146 poll.

The goal of the PR is to try to pull in new pip versions, however:
- do not degrade virtual environment creation performance significantly,
- do not trigger during CI runs (the feature is aimed mostly at end-users, that install virtualenv themselves).

The proposed change would make pip upgrade periodically:
- every two weeks try to upgrade pip to a newer version
- the upgrade runs not inline, but instead as a new background process (this makes the performance overhead not significant)
- upgrade the seeded pip to a new version if there's a newer version that satisfies the following criteria

  - was discovered at least an hour ago (this is to ensure that CI environments don't start with an old version and mid-way through they switch to a newer version)
  - was released at least 28 days ago (allow 4 weeks for pip to do a bugfix release).

I consider this a reasonable middle-ground. We will mostly use bundled versions (and almost always in the CI), but for end-users, once a pip version is out there for a long time we do switch to it, as we consider that a proof of stability.

People who would like to always get the latest version should either set the environment variable ``VIRTUALENV_DOWNLOAD=1`` or set ``download`` inside the ``virtualenv.ini`` file (or pass in the ``--download`` flag).

People who would like to always pin to a given version should either set the environment variable ``VIRTUALENV_PIP=20.1`` or set ``pip`` inside the ``virtualenv.ini`` file (or pass in the ``--pip`` flag).

Resolves https://github.com/pypa/virtualenv/issues/1821

### DIFF
--- a/src/virtualenv/__main__.py
+++ b/src/virtualenv/__main__.py
@@ -39,7 +39,18 @@ class LogSession(object):
             "  creator {}".format(ensure_text(str(self.session.creator))),
         ]
         if self.session.seeder.enabled:
-            lines += ("  seeder {}".format(ensure_text(str(self.session.seeder))),)
+            lines += (
+                "  seeder {}".format(ensure_text(str(self.session.seeder))),
+                "    added seed packages: {}".format(
+                    ", ".join(
+                        sorted(
+                            "==".join(i.stem.split("-"))
+                            for i in self.session.creator.purelib.iterdir()
+                            if i.suffix == ".dist-info"
+                        ),
+                    ),
+                ),
+            )
         if self.session.activators:
             lines.append("  activators {}".format(",".join(i.__class__.__name__ for i in self.session.activators)))
         return os.linesep.join(lines)
--- a/src/virtualenv/seed/embed/pip_invoke.py
+++ b/src/virtualenv/seed/embed/pip_invoke.py
@@ -6,7 +6,7 @@ from contextlib import contextmanager
 from virtualenv.discovery.cached_py_info import LogCmd
 from virtualenv.info import PY3
 from virtualenv.seed.embed.base_embed import BaseEmbed
-from virtualenv.seed.embed.wheels.acquire import get_bundled_wheel, pip_wheel_env_run
+from virtualenv.seed.embed.wheels.acquire import Version, get_bundled_wheel, pip_wheel_env_run
 from virtualenv.util.subprocess import Popen
 from virtualenv.util.zipapp import ensure_file_on_disk
 
@@ -17,6 +17,13 @@ else:
 
 
 class PipInvoke(BaseEmbed):
+
+    packages = {
+        "pip": Version.latest,
+        "setuptools": Version.latest,
+        "wheel": Version.latest,
+    }
+
     def __init__(self, options):
         super(PipInvoke, self).__init__(options)
 
@@ -38,10 +45,12 @@ class PipInvoke(BaseEmbed):
             cmd.append("--no-index")
         pkg_versions = self.package_version()
         for key, ver in pkg_versions.items():
-            cmd.append("{}{}".format(key, "=={}".format(ver) if ver is not None else ""))
+            cmd.append("{}{}".format(key, "" if ver == Version.latest else "=={}".format(ver)))
         with ExitStack() as stack:
             folders = set()
-            for context in (ensure_file_on_disk(get_bundled_wheel(p, version), self.app_data) for p in pkg_versions):
+            for context in (
+                ensure_file_on_disk(get_bundled_wheel(p, version).path, self.app_data) for p in pkg_versions
+            ):
                 folders.add(stack.enter_context(context).parent)
             for folder in folders:
                 cmd.extend(["--find-links", str(folder)])
--- a/src/virtualenv/seed/embed/wheels/periodic_update.py
+++ b/src/virtualenv/seed/embed/wheels/periodic_update.py
@@ -1,0 +1,195 @@
+"""
+Periodically update bundled versions.
+"""
+
+import calendar
+import json
+import logging
+import subprocess
+import sys
+import tempfile
+from contextlib import contextmanager
+from datetime import datetime, timedelta
+from shutil import copy2
+
+from six.moves.urllib.request import urlopen
+
+from virtualenv.seed.embed.wheels.util import Wheel
+from virtualenv.util.lock import ReentrantFileLock, Timeout
+from virtualenv.util.path import Path, safe_delete
+from virtualenv.util.subprocess import DETACHED_PROCESS, Popen
+
+
+def periodic_update(distribution, for_py_version, wheel, cache_dir, app_data):
+    if distribution != "pip":
+        raise RuntimeError("only pip may be periodically updated")
+
+    needs_update = False
+    with update_log_for_distribution(cache_dir.parent, distribution, no_block=True) as update_log:
+        if update_log.needs_update:
+            update_log.started = datetime.now()
+            update_log.update()
+            needs_update = True
+    if needs_update:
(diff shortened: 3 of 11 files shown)


src/virtualenv/seed/embed/wheels/periodic_update.py
```diff
@@ -0,0 +1,195 @@
...
+    for filename, release_date in (
+        (update_log.updated, update_log.updated_release_date),
+        (update_log.previous_updated, update_log.previous_updated_release_date),
+    ):
+        if filename is None or filename == wheel.name:
+            continue
+        updated_wheel = Wheel(cache_dir / filename)
+        # use it only if released long enough time ago - use version number to approximate release
+        # only use if it has been released for at least 28 days
+        if datetime.now() - release_date > timedelta(days=28):
```
gaborbernat, line 45: use periodically upgraded bundle if it's older release than 28 days

src/virtualenv/seed/embed/wheels/periodic_update.py
```diff
@@ -0,0 +1,195 @@
...
+            u_log.update()
+
+    finally:
+        safe_delete(temp_dir)
+
+
+def _get_release_date(dest):
+    wheel = Wheel(dest)
+    # the most accurate is to ask PyPi - https://pypi.org/pypi/pip/json
+    try:
```
gaborbernat, line 180: get release date primarily from PyPi

src/virtualenv/seed/embed/wheels/periodic_update.py
```diff
@@ -0,0 +1,195 @@
...
+        with urlopen("https://pypi.org/pypi/{}/json".format(wheel.distribution)) as file_handler:
+            content = json.load(file_handler)
+        return datetime.strptime(content["releases"][wheel.version][0]["upload_time"], "%Y-%m-%dT%H:%M:%S")
+    except Exception:  # noqa
+        pass
+    # otherwise can approximate from the version number https://pip.pypa.io/en/latest/development/release-process/
+    released = datetime(year=2000 + wheel.version_tuple[0], month=wheel.version_tuple[1] * 3 + 1, day=1)
```
gaborbernat, line 187: or if that's not available then approximate from version number

src/virtualenv/seed/embed/wheels/periodic_update.py
```diff
@@ -0,0 +1,195 @@
...
+def trigger_update(distribution, for_py_version, wheel, app_data):
+    cmd = [
+        sys.executable,
+        "-c",
+        "from virtualenv.seed.embed.wheels.periodic_update import do_update;"
+        "do_update({!r}, {!r}, {!r}, {!r})".format(distribution, for_py_version, str(wheel.path), str(app_data.path)),
+    ]
+    kwargs = {"stdout": subprocess.PIPE, "stderr": subprocess.PIPE}
+    if sys.platform == "win32":
+        kwargs["creation_flags"] = DETACHED_PROCESS
```
gaborbernat, line 135: The update is a detached new process to be run in the background (to not slow down creation).

src/virtualenv/seed/embed/wheels/periodic_update.py
```diff
@@ -0,0 +1,195 @@
...
+        return datetime.strftime(value, self.datetime_fmt)
+
+    @property
+    def needs_update(self):
+        now = datetime.now()
+        if self.completed is None:  # never completed
+            return self._check_start(now)
+        else:
+            delta_since_last_completed = now - self.completed
+            if delta_since_last_completed < timedelta(days=13):
```
gaborbernat, line 101: trigger periodic update once every two week

src/virtualenv/seed/embed/pip_invoke.py
```diff
@@ -17,6 +17,13 @@
 
 
 class PipInvoke(BaseEmbed):
+
+    packages = {
+        "pip": Version.latest,
```
gaborbernat, line 22: When using the legacy pip seeder, we always use the latest version. This will be latest version in bundled + search dir if running in no download mode, otherwise upstream latest.

src/virtualenv/__main__.py
```diff
@@ -39,7 +39,18 @@ def __str__(self):
             "  creator {}".format(ensure_text(str(self.session.creator))),
         ]
         if self.session.seeder.enabled:
-            lines += ("  seeder {}".format(ensure_text(str(self.session.seeder))),)
+            lines += (
+                "  seeder {}".format(ensure_text(str(self.session.seeder))),
+                "    added seed packages: {}".format(
+                    ", ".join(
+                        sorted(
```
gaborbernat, line 46: In our report now we print exactly the version numbers of the seeded packages.

asottile:
my 2c: I don't think the upgrade should happen magically but should be an explicit command a user runs in order to pull in an upgrade.

always magically downloading the latest and only once every few weeks magically downloading the latest are the same problem set imo

pradyunsg:
I'm gonna cop out of this discussion, since I'd like to avoid adding this to the things I'm actively thinking about. :)

pfmoore:
I'm also going to opt out - my view as expressed on Discourse was that I prefer using the seeded version by default (for performance reasons, not reproducibility) but I'm OK with updating in the background if that's useful for the people who want the latest version to be used. I'll leave those people to say whether this solution suits them or not.

gaborbernat:
Considering we have two maintainers of this and pip that don't actually want this now I'm strongly considering now letting the always embedded camp win. If pip maintainers are not worried about this, Donalds worry of this holding back the evolution of the ecosystem feels unwarranted. (I'm also in favour of always embedded, but as the one driving the discussion I wanted to remove myself from the count 🤷‍♂️ sorry Donald)

gaborbernat:
Superseded by https://github.com/pypa/virtualenv/pull/1833

pradyunsg:
> Considering we have two maintainers of this and pip that don't actually want this now I'm strongly considering letting the always embedded camp win

Hmph. I'm not saying that "I don't want this now" -- but rather that *now* isn't a good time for me to spend working/pushing on this specific thing, since I have other things to on my plate currently.

I do think the approach in this PR is better than always-embedded (which is better than status-quo). This is a compromise solution which makes everyone slightly unhappy but accomodate for every issue "well enough", vs any other choice which'll make some folks very unhappy and some folks very happy.

Further, and I've stated this before elsewhere, I don't think the public vote is a good argument for picking what functionality we provide, since I'm not sure we'd provided enough context and (more importantly) we had no metrics/checks in place to determine if/how it was swayed based on what "kinds" of users "showed up to the vote". As an example, all pip maintainers/distro packagers (except Paul and Debian's maintainer .-.) voted for the middle ground option. I don't know *why* these folks voted they way that they did, and I have no context on other votes as well.

gaborbernat:
> I don't think the public vote is a good argument for picking what functionality we provide since I'm not sure we'd provided enough context

I don't agree on not providing enough context. There is a long wall of text before that poll (https://discuss.python.org/t/how-should-virtualenv-behave-respective-of-seed-packages-e-g-pip-by-default/4146) explaining the context, and multiple made follow-up posts to expand further on some context. 

I also don't agree on public vote not being a good place. I agree it has some shortcomings but I'm not aware of any other better methods. Donald raised in the original issue that back then everyone was on board with always download. I challenged that statement, and made a more up-to-date request for comment/survey; which shows as things stand today there seem to be more people in favour of always embedded than download or periodic update. I've reached out on the usual PEP channels to advertise the topic, so anyone who had strong feelings about the topic could join in.

> I'd much rather prefer a what-are-the-arguments-and-trade-offs based choice,

This has been debated in long on that topic.

Issue #1834, opened 2020-05-19
`--seeder pip` does not work with `--extra-search-dir`


frenzymadness:
**Issue**

The problem is [here](https://github.com/pypa/virtualenv/blob/master/src/virtualenv/seed/embed/pip_invoke.py#L48). When one or more extra search dir is specified, the resulting command looks like this `[<python>, '-m', 'pip', '-q', 'install', '--only-binary', ':all:', '--no-index', 'pip', 'setuptools', 'wheel', '--find-links', '/usr/lib/python3.8/site-packages/virtualenv/seed/embed/wheels', [extra dir one], [extra dir two]]`. Which leads to pip tries to install `[extra dir one]` as a package and error message: `ERROR: Directory '[extra dir one]' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.`

pip needs to have each directory specified with `--find-links` so the command should look like this: ``[<python>, '-m', 'pip', '-q', 'install', '--only-binary', ':all:', '--no-index', 'pip', 'setuptools', 'wheel', '--find-links', '/usr/lib/python3.8/site-packages/virtualenv/seed/embed/wheels', '--find-links', [extra dir one], '--find-links', [extra dir two]]``.

I'll open a PR for this.

gaborbernat:
This now has been released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-21-2020-05-20

PR #1835, opened 2020-05-19
Fix pip seeder with --extra-search-dir specified

frenzymadness:
pip needs to have `--find-links` CLI argument for each directory
not just once for a list of directories.

Fixes: https://github.com/pypa/virtualenv/issues/1834

If you agree with the fix, I'll try to prepare a test for it and a changelog entry.

### DIFF
--- a/src/virtualenv/seed/embed/pip_invoke.py
+++ b/src/virtualenv/seed/embed/pip_invoke.py
@@ -43,7 +43,7 @@ class PipInvoke(BaseEmbed):
             folders = set()
             for context in (ensure_file_on_disk(get_bundled_wheel(p, version), self.app_data) for p in pkg_versions):
                 folders.add(stack.enter_context(context).parent)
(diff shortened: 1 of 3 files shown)


src/virtualenv/seed/embed/pip_invoke.py
```diff
@@ -45,5 +45,6 @@ def get_pip_install_cmd(self, exe, version):
                 folders.add(stack.enter_context(context).parent)
             for folder in folders:
```
gaborbernat, line 46:
```python
folders.update(set(self.extra_search_dir))
for folder in folders:
    cmd.extend(["--find-links", str(folder)])
```

frenzymadness:
Fix updated, test also updated, changelog added. The change to test is basically no-op but it verifies that the command generated in `get_pip_install_cmd()` is correct — updated test fails without the fix and passes with it.

PR #1841, opened 2020-05-28
Implement periodic update feature

### LINKED ISSUE #1821: Ensure bundled pip does not get too out of date

### LINKED ISSUE #1840: Plug-in for bundled wheels
Hello.

First of all, thank you for your work. New virtualenv looks very good and its codebase is clean and readable :+1: Second thing is that I don't want to start another discussion about bundled wheels and their support and updates. I just have a problem to solve and I'd like to do it in cooperation with you because I think that it might help other maintainers.

We'd like to update the python-virtualenv RPM package in Fedora from version 16.7.10 to 20.0.20. The problem we have to solve is that we want the virtualenv provided by the distribution to use our own wheels for pip/setuptools/wheel.

When searching for wheels, we first need to take a look into ensurepip folder like `/usr/lib64/python3.4/ensurepip/_bundled` — this folder exists only for some Pythons like 3.4 or 2.7 and the path is obtained from ensurepip module `import ensurepip; print(ensurepip.__path__[0])`. If we don't find a proper wheel there, we take a look into `/usr/share/python-wheels` where we always find a wheel for all three bundled packages. The latter might contain wheels not compatible with an older Python so ensurepip always has to take precedence.

We always solve this problem by a downstream patch but you can probably imagine that it's not an easy task to maintain a patch like this. So, I'd like to discuss our solutions with you and we might find a better way which makes it easier for us to use the latest virtualenv with our own wheels.

The main problem with the biggest impact on how the patches mentioned below look is that we always need to have a path to Python interpreter we create a virtual environment for to be able to get the ensurepip path mentioned above.

In solution 1, I've implemented a module with a simple function that yields paths to bundled wheels. I then use it in `BaseEmbed` to alter `extra_search_dir` and in `pip_wheel_env_run` for the same purpose. The patch is here: https://github.com/frenzymadness/virtualenv/pull/1/files

Then I have had an idea to implement it in a different way. I've implemented `BUNDLE_SUPPORT` as an object which gives me a correct path for each wheel. This looked promising but then, because it also needs to have a path to Python interpreter, it gets complicated. I had to change a lot of functions going down to `get_bundled_wheel` to pass creator or interpreter to it. The patch is here https://github.com/frenzymadness/virtualenv/pull/2/files

The important thing is that we cannot save the paths somewhere because the wheels are generated by other packages and they might be updated without python-virtualenv update.

What do you think about it? Might it be possible to provide an alternative mechanism (some kind of plug-in) to determine the correct path to bundled wheels?

I am not asking you to implement it for us but you might see something I don't and if we will find a good solution that makes it possible to extend virtualenv this way, I'll be glad to implement it.

gaborbernat:
- print out seeded versions in report
- periodically check in a background out of band process if the embed
  wheels are out of date
- promote app data at root level, with three implementations - via disk
  folder, via temp folder and via not available
- introduce a common wheelhouse to not replicate wheels accross python
  versions, decreasing size of the app data folder
- allow manually upgrading the embeded wheels by using the
  --upgrade-embed-wheels flag, in this case no virtual environment is
  created
- upgrade setuptools to 47.1.0

Resolves #1821.
Resolves #1840.


docs/user_guide.rst
```diff
@@ -131,6 +131,39 @@ main seed mechanism available:
...
+  end users might not be able to upgrade virtualenv at the same speed. Therefore, a user might request to upgrade the
+  list of embedded wheels by invoking virtualenv with the :option:`upgrade-embed-wheels` flag. This operation will
+  trigger automatically, as a background process, if no upgrade has been performed in the last 14 days and upgrade the
+  embedded wheels if they have been our for more than 28 days. This 28 days period should guarantee end users are not
+  pulling in automatically releases that have known bugs within. This automatic behaviour might be disabled via the
+  :option:`no-periodic-update` configuration flag/option. To acquire the release date of a package virtualenv will
+  perform the following:
+
+  - lookup ``https://pypi.org/pypi/<distribution>/json`` (primary truth source),
+  - save the date the version was first discovered, and wait until 28 days passed.
```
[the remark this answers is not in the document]
pradyunsg: On the JSON page, `data["releases"][version]["upload_time_iso_8601"]` should get us this information. :)

The question is whether 28 days since release is better, or 28 days since release seen is better? (also, how careful do we want to be, that we don't block upgrades if there's a release every 27 days or something like that)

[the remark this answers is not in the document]
gaborbernat: If you take a look at the implementation code, see https://github.com/gaborbernat/virtualenv/blob/bundle/src/virtualenv/seed/wheels/periodic_update.py, this is already implemented. We take 28 days since release if available, otherwise 28 days since seen. This ensures at least 28 days of grace period which is what we want here.

> also, how careful do we want to be, that we don't block upgrades if there's a release every 27 days or something like that

This is not a problem because in such cases we'll pick the first available minor that's at least 28 days old, see https://github.com/gaborbernat/virtualenv/blob/bundle/src/virtualenv/seed/wheels/periodic_update.py#L44

setup.py
```diff
@@ -4,6 +4,9 @@
     raise RuntimeError("setuptools >= 41 required to build")
 
 setup(
-    use_scm_version={"write_to": "src/virtualenv/version.py", "write_to_template": '__version__ = "{version}"'},
+    use_scm_version={
+        "write_to": "src/virtualenv/version.py",
+        "write_to_template": 'from __future__ import  unicode_literals;\n\n__version__ = "{version}"',
```
webknjaz, line 9: did you mean
```suggestion
        "write_to_template": 'from __future__ import  unicode_literals;\n\n__version__ = u"{version}"',
```

setup.py
```diff
@@ -4,6 +4,9 @@
     raise RuntimeError("setuptools >= 41 required to build")
 
 setup(
-    use_scm_version={"write_to": "src/virtualenv/version.py", "write_to_template": '__version__ = "{version}"'},
+    use_scm_version={
```
webknjaz, line 7: By the way, modern setuptools-scm now advertise `pyproject.toml` as a main location for this config entries: https://github.com/pypa/setuptools_scm#pyprojecttoml-usage.
That is if you bump the deps to `"setuptools_scm[toml]>=3.4"`.

setup.py
gaborbernat: Not needed, we import unicode literals just before.

setup.py
gaborbernat: Sadly centos do not like yet that, we had modern but reverted the change to be more graceful on centos.

setup.py
webknjaz: Ah, wasn't reading it right. FWIW using the prefix is shorter than that future import...

gaborbernat: I've used the import because that's what we use everywhere else, so consistency.

setup.py
gaborbernat: We must take seriously our support list https://virtualenv.pypa.io/en/latest/installation.html#python-and-os-compatibility

gaborbernat:
This is now ready to go. I expect to merge it over the weekend and ship it on Monday morning.

dstufft:
If I'm reading this correctly, this will download the wheel but hold onto it for 14 days or so before using it? Is that right?

gaborbernat:
If it's downloaded automatically hold onto it until it becomes 28 days old release (one hour after download). You can always make it use right away by manually invoking the upgrade process via the flag.

dstufft:
Would it be useful if the simple index included the release date to use that instead of downloading and holding onto it? If not that’s fine, just thought it might be a easy optimization.

gaborbernat:
> Would it be useful if the simple index included the release date to use that instead of downloading and holding onto it?

It would be useful, though I'd guess we'd still hold onto it, but we'd not need to go to the PyPi API to fetch it. That being said we'd likely need to keep around the current implementation until the simple index is able to provide this information. Alternatively, it would also work if we could extract this information from the wheel/sdist itself somehow.

asottile:
I'm seeing breaking (presumably) due to this change.  virtualenv is no longer reproducible and I've only just noticed setuptools>=50 despite not changing my virtualenv version.

I think this should be reverted back to when virtualenv was reproducible by default, we're not making the situation better for edge-breaking and in may ways it's more confusing now that it's delayed by several weeks.

gaborbernat:
This topic has been over discussed, and both sides made their point. There has been a majority desire both on pypa and general public side of some form of auto upgrade, which has been implemented here. There are configurations available to achieve the old behavior, but we don't plan to revert this feature. (For what it's worth I was also against this feature personally, but I consider it a decent middle ground over the old always update behavior).

PR #1844, opened 2020-06-02
Fix Github Actions CI - introduce codecov

gaborbernat:


.github/workflows/check.yml
```diff
@@ -32,44 +31,61 @@ jobs:
...
+    - name: install OS dependencies
+      run:   |
+        if [ "$RUNNER_OS" == "Linux" ]; then
+             sudo apt-get update -y && sudo apt-get install fish csh
+        elif [ "$RUNNER_OS" == "Windows" ]; then
+             choco install vcpython27 --yes -f
+        elif [ "$RUNNER_OS" == "macOS" ]; then
+             brew update -vvv && brew install fish tcsh
+        fi
+      shell: bash
```
webknjaz, line 48: I personally prefer these as separate steps so that they are exposed better in the log:
```suggestion
    - name: install GNU/Linux dependencies
      if: runner.os == 'Linux'
      run: sudo apt-get update -y && sudo apt-get install fish csh
    - name: install macOS dependencies
      run: brew update -vvv && brew install fish tcsh
      if: runner.os == 'macOS'
    - name: install Windows dependencies
      run: choco install vcpython27 --yes -f
      if: runner.os == 'Windows'
```

.github/workflows/check.yml
```diff
@@ -32,44 +31,61 @@ jobs:
...
     - name: setup python for tox
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: install tox
       run: python -m pip install tox
+    - uses: actions/checkout@v2
+    - name: pick to environment to run
+      run: |
+        python -c "print('::set-env name=TOXENV::' + {'3.8': 'py38', '3.7': 'py37', '3.6': 'py36', '3.5': 'py35', '2.7': 'py27', 'pypy2': 'pypy', 'pypy3': 'pypy3'}['${{ matrix.py}}'])"
```
webknjaz, line 58: Shocking but you don't really need this terrible nesting:
```suggestion
      run: |
        print(
            '::set-env name=TOXENV::' +
            {'3.8': 'py38', '3.7': 'py37', '3.6': 'py36', '3.5': 'py35', '2.7': 'py27', 'pypy2': 'pypy', 'pypy3': 'pypy3'}['${{ matrix.py}}']
        )
      shell: python
```

webknjaz: (I personally prefer putting this into the matrix tho)

webknjaz:
You may be interested in taking a look at the workflow I'm working on right now: https://github.com/ansible/pylibssh/blob/devel/.github/workflows/build-test-n-publish.yml.
It has a separate build stage that goes first and publishes the artifacts. Then, test and publish jobs download those artifacts (sdist+wheels) and work with them instead of testing the source (or rebuilding things separately).

.github/workflows/check.yml
gaborbernat: Yeah had that first but this makes it clearer that only one of them is active at a time 🤔😎

webknjaz: if-clause makes other steps be marked grey so usually it's okay combined with the os in the step name

.github/workflows/check.yml
webknjaz: By the way, why don't you just do static `TOXENV=python` and manage pythons on the job matrix level? Any problems with that?

.github/workflows/check.yml
```diff
@@ -32,44 +31,61 @@ jobs:
...
+        fi
+      shell: bash
     - name: setup python for tox
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: install tox
       run: python -m pip install tox
+    - uses: actions/checkout@v2
+    - name: pick to environment to run
```
webknjaz, line 56: s/to/an/ ?

.github/workflows/check.yml
gaborbernat: I prefer no step present over gray step 😉 also decreases the step count this way

.github/workflows/check.yml
```diff
@@ -32,44 +31,61 @@ jobs:
         - 3.8
         - 3.7
         - 3.6
+        - 3.5
+        - pypy3
         - 2.7
         - pypy2
-        - pypy3
-
     steps:
```
webknjaz, line 38: How about
```suggestion
    env:
      TOXENV: python

    steps:
```

.github/workflows/check.yml
gaborbernat: Because very soon tox will not support running with old pythons, and I prefer tox running on latest python as opposed to whatever the test python needs to be (python 2.7 and such)

.github/workflows/check.yml
gaborbernat: That wouldn't work and especially wouldn't active the conditional setenv parts.

.github/workflows/check.yml
```diff
@@ -32,44 +31,76 @@ jobs:
...
+      run: tox -vv --notest
+    - name: run test suite
+      run: tox
+      env:
+        PYTEST_ADDOPTS: "-vv --durations=20"
+        CI_RUN: 'yes'
+        DIFF_AGAINST: HEAD
+    - name: pick environment to run
+      run: |
+        import os; os.rename('.tox/coverage.{}.xml'.format(os.environ['TOXENV']), '.tox/coverage.xml')
```
webknjaz, line 77: Do you have anything against multilining? This works too:
```suggestion
        import os
        os.rename('.tox/coverage.{}.xml'.format(os.environ['TOXENV']), '.tox/coverage.xml')
```

.github/workflows/check.yml
```diff
@@ -32,44 +31,76 @@ jobs:
...
+      shell: python
+    - uses: codecov/codecov-action@v1
+      with:
+        file: ./.tox/coverage.xml
+        flags: tests
+        env_vars: OS,PYTHON
+        name: ${{ matrix.py }} - ${{ matrix.os }}
 
   check:
```
webknjaz, line 86: Since GHA provide a lot of jobs, I usually put linters into a completely different workflow so that they don't influence/cancel test jobs.

.github/workflows/check.yml
webknjaz: @gaborbernat it should work unless I'm overlooking something. I use it all the time. It only requires `actions/setup-python` to use a var from the matrix: https://github.com/ansible/pylibssh/blob/c13f2f5/.github/workflows/build-test-n-publish.yml#L358. And no dynamic TOXENV mapping needed at all.

webknjaz: Oh, are you referring to the conditionals in `tox.ini`? Missed that.

gaborbernat: Yup.

.github/workflows/check.yml
gaborbernat: In this case, I feel like it does not decreases readability as makes the file longer

.github/workflows/check.yml
gaborbernat: I've already disabled the cancel feature. So that does not affect me.

.github/workflows/check.yml
```diff
@@ -32,44 +31,78 @@ jobs:
...
+      run:   |
+        if [[ "$RUNNER_OS" == "Linux" ]]; then
+             sudo apt-get update -y && sudo apt-get install fish csh
+        elif [[ "$RUNNER_OS" == "Windows"]]; then
+             if [ "${{ matrix.py }}" == "2.7" || "${{ matrix.py }}" == "pypy2" ]; then
+                 choco install vcpython27 --yes -f
+             fi
+        elif [[ "$RUNNER_OS" == "macOS" ]]; then
+             brew update -vvv && brew install fish tcsh
+        fi
```
webknjaz, line 49: You could also try actions level conditionals:
```suggestion
        ${{ runner.os == "Linux" && "sudo apt-get update -y && sudo apt-get install fish csh" }}
        ${{ runner.os == "macOS" && "brew update -vvv && brew install fish tcsh" }}
        ${{ runner.os == "Windows" && ( matrix.py == "2.7" || matrix.py == "pypy2" ) && "choco install vcpython27 --yes -f" }}
```

gaborbernat: needed an || true at end, but this works 👍

.github/workflows/check.yml
```diff
@@ -32,44 +31,84 @@ jobs:
...
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: install tox
       run: python -m pip install tox
+    - uses: actions/checkout@v2
+    - name: use local virtualenv for tox
+      run: python -m pip install ${{ runner.os == 'Windows' && matrix.py == 'pypy3' && 'virtualenv<20' || '.'}}
     - name: setup python for test ${{ matrix.py }}
+      if: "!startsWith(matrix.py,'brew@py')"
```
webknjaz, line 58: Pro tip: converting such strings to multiline syntax helps to avoid extra quotes (which may become hard to escape at times)
```suggestion
      if: >-
        !startsWith(matrix.py, 'brew@py')
```

.github/workflows/check.yml
```diff
@@ -32,44 +31,84 @@ jobs:
...
+      shell: bash
     - name: setup python for tox
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
     - name: install tox
       run: python -m pip install tox
+    - uses: actions/checkout@v2
+    - name: use local virtualenv for tox
+      run: python -m pip install ${{ runner.os == 'Windows' && matrix.py == 'pypy3' && 'virtualenv<20' || '.'}}
```
webknjaz, line 56: FYI you can put logic into separate steps and pass values via outputs: https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter

.github/workflows/check.yml
```diff
@@ -32,44 +31,84 @@ jobs:
...
-      run: tox -vve py --notest
-    - name: run test suite for ${{ matrix.os }}-${{ matrix.py }}
-      run: tox -e py
+    - name: load python from ${{ matrix.py }}
+      if: startsWith(matrix.py,'brew@py')
+      run: |
+        import os; import subprocess
+        version = "${{matrix.py}}"[-1]
+        subprocess.check_call(["brew", "install", f"python@{version}"])
+        print('::set-env name=PATH::' + (":".join([f"/usr/local/opt/python@{version}"] + os.environ["PATH"].split(":"))))
```
webknjaz, line 68: Using `add-path` command would be a bit more elegant https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path
```suggestion
        print(f'::add-path::/usr/local/opt/python@{version}')
```

.github/workflows/check.yml
gaborbernat: I prefer it more condensed as long as the expression is as simple as here 👍

.github/workflows/check.yml
gaborbernat: I think in this case this is the best way, but thanks 👍

.github/workflows/check.yml
gaborbernat: Oh did not knew about it, will change this. Is this prepend path or add? I really need prepend here 🤔

webknjaz: Yep, if you click on that link above, here's the explanation you'll see:

> Prepends a directory to the system `PATH` variable for all subsequent actions in the current job. The currently running action cannot access the new path variable.

PR #1855, opened 2020-06-12
Upgrade embeded setuptools

gaborbernat:


setup.cfg
```diff
@@ -61,7 +61,7 @@ virtualenv.activate =
     batch = virtualenv.activation.batch:BatchActivator
     fish = virtualenv.activation.fish:FishActivator
     powershell = virtualenv.activation.powershell:PowerShellActivator
-    python = virtualenv.activation.python:PythonActivator
```
RowdyHowell, line 64: I'm getting an error that might be related to this merge.  Is the change expected?

`AttributeError: module 'virtualenv.activation.python' has no attribute 'PestythonActivator'`

victor-torres:
Same issue here:

```
AttributeError: module 'virtualenv.activation.python' has no attribute 'PestythonActivator'
```

setup.cfg
brutimus: This just broke all of my production build pipelines as well.

gaborbernat: I have to be honest, have no idea how this change made it in. Late Friday releases are always dangerous but wanted to get out a fix release to address peoples issues due to https://github.com/pypa/virtualenv/pull/1846 🤦 🤷‍♂️

victor-torres: It happens. Thanks for the quick fix! @gaborbernat

phrfpeixoto:
Haha. It seems we're missing tests here.

PR #1867, opened 2020-06-21
Improve error message for no .dist-info

gaborbernat:
Inside the copy app-data installer.


gaborbernat:
This now has been released under https://virtualenv.pypa.io/en/20.0.24/changelog.html#v20-0-24-2020-06-22

PR #1880, opened 2020-06-29
Remove distlib monkeypatches

gaborbernat:
Adapt the new features added in distlib instead.


gaborbernat:
@pradyunsg can pip not ship the x.y format within the console scripts?

Issue #1818, opened 2020-05-05
Issues creating and activating a virtualenv in git bash


asottile:
**Issue**

`PATH` (and maybe `VIRTUAL_ENV`?) seems to be set incorrectly on windows when using git bash, this worked with virtualenv 16.x

**Environment**

Provide at least:
- OS: windows 10
- ``pip list`` of the host python where ``virtualenv`` is installed:

```console
$ virtualenv --version
virtualenv 20.0.20 from c:\python37\lib\site-packages\virtualenv\__init__.py

$ python --version --version
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]

$ pip freeze
appdirs==1.4.3
colorama==0.4.3
distlib==0.3.0
filelock==3.0.12
importlib-metadata==1.6.0
packaging==20.3
pluggy==0.13.1
py==1.8.1
pyparsing==2.4.7
six==1.14.0
toml==0.10.0
tox==3.14.6
virtualenv==20.0.20
zipp==3.1.0
  ```

**Output of the virtual environment creation**

Make sure to run the creation with `-vvv --with-traceback`:

```console
$ virtualenv venv -vvv
87 setup logging to NOTSET [DEBUG report:42]
94 find interpreter for spec PythonSpec(path=c:\python37\python.exe) [INFO builtin:44]
94 proposed PythonInfo(spec=CPython3.7.3.final.0-64, exe=c:\python37\python.exe, platform=win32, version='3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]', encoding_fs_io=utf-8-cp1252) [INFO builtin:50]
94 accepted PythonInfo(spec=CPython3.7.3.final.0-64, exe=c:\python37\python.exe, platform=win32, version='3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]', encoding_fs_io=utf-8-cp1252) [DEBUG builtin:52]
99 symlink on filesystem does work [DEBUG info:47]
99 filesystem is not case-sensitive [DEBUG info:28]
162 create virtual environment via CPython3Windows(dest=C:\Users\Anthony\AppData\Local\Temp\x\venv, clear=False, global=False) [INFO session:52]
[...]

PR #1924, opened 2020-08-15
Use unix line-endings in bash activate script

saytosid:
Fixes: #1818

Signed-off-by: Siddhant Kumar <skumar619@bloomberg.net>

- [ ] updated/extended the documentation - N/A, this is a bug fix


src/virtualenv/activation/bash/__init__.py
```diff
@@ -1,13 +1,35 @@
...
 from virtualenv.util.path import Path
+from virtualenv.util.six import ensure_text
 
 from ..via_template import ViaTemplateActivator
 
 
 class BashActivator(ViaTemplateActivator):
+    def generate(self, creator):
+        generated = super(BashActivator, self).generate(creator)
+        if IS_WIN:
```
gaborbernat, line 13: Why do this conversion and not just always use linux line endings? 🤔

saytosid: To keep the change localized to `BashActivator`, else we'd have to make the change in `ViaTemplateActivator` which will affect many activators.

gaborbernat: Seems the line normalization happens because we use read_text/write_text. In case of these activator templates I don't think we want any of newline normalization, so a better solution is to just switch to reading binary data for the templates, and encode it as UTF-8 ourselves.

PR #1946, opened 2020-09-22
Fix activator path in cygwin and msys2

### LINKED ISSUE #1940: Cygwin and msys2 bash activator compatibility
**Issue**

Run the following via msys2:

```
$ /c/Users/builder/AppData/Local/Programs/Python/Python39/Scripts/virtualenv.exe --no-pip --no-setuptools --no-wheel -p /c/Users/builder/AppData/Local/Programs/Python/Python39/python.exe /tmp/tmp.uw7eZfvJIO
$ source /tmp/tmp.uw7eZfvJIO/Scripts/activate
$ python
python: command not found
```

The content of `/tmp/tmp.uw7eZfvJIO/Scripts/activate` contains `VIRTUAL_ENV='C:\tools\msys64\tmp\tmp.uw7eZfvJIO'`, but it's expected to be `VIRTUAL_ENV='/c/tools/msys64/tmp/tmp.uw7eZfvJIO'`, thus it can't locate `python`.

I believe the legacy version handles this correctly here: https://github.com/pypa/virtualenv/blob/4c9899f471139670b0f1bfd359c37a67c507447d/virtualenv.py#L1742-L1748

which contains:

`VIRTUAL_ENV="$(if [ "$OSTYPE" "==" "cygwin" ]; then cygpath -u 'C:\tools\msys64\tmp\tmp.uw7eZfvJIO'; else echo '/C/tools/msys64/tmp/tmp.uw7eZfvJIO'; fi;)"`

**Environment**

Provide at least:
- OS: Windows 7, using msys2 v20200903.0.0

danyeaw:
Closes #1940. In cygwin and MSYS2, the path is in POSIX format. This PR converts the Windows path to POSIX format using the cygpath utility so that it is added to the bash activation script in the correct format.

- [X] ran the linter to address style issues (``tox -e fix_lint``)
- [X] ensured there are test(s) validating the fix
- [X] added news fragment in ``docs/changelog`` folder

### DIFF
--- a/src/virtualenv/activation/via_template.py
+++ b/src/virtualenv/activation/via_template.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
 import os
+import re
 import sys
+import sysconfig
 from abc import ABCMeta, abstractmethod
 
 from six import add_metaclass
@@ -31,9 +33,18 @@ class ViaTemplateActivator(Activator):
         return generated
 
     def replacements(self, creator, dest_folder):
+        if "mingw" or "cygwin" or "msys" in sysconfig.get_platform():
+            pattern = re.compile("^([A-Za-z]):(.*)")
+            match = pattern.match(str(creator.dest))
+            if match:
+                virtual_env = "/" + match.group(1).lower() + match.group(2)
+            else:
+                virtual_env = str(creator.dest)
+        else:
+            virtual_env = str(creator.dest)
         return {
             "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
-            "__VIRTUAL_ENV__": ensure_text(str(creator.dest)),
+            "__VIRTUAL_ENV__": ensure_text(virtual_env),
             "__VIRTUAL_NAME__": creator.env_name,
             "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
             "__PATH_SEP__": ensure_text(os.pathsep),
--- a/docs/changelog/1940.bugfix.rst
+++ b/docs/changelog/1940.bugfix.rst
@@ -1,0 +1,1 @@
+Provide correct path for bash activator in cygwin or msys2 - by :user:`danyeaw`.


src/virtualenv/activation/via_template.py
```diff
@@ -31,9 +33,13 @@ def generate(self, creator):
         return generated
 
     def replacements(self, creator, dest_folder):
+        if "mingw" or "cygwin" or "msys" in sysconfig.get_platform():
+            virtual_env = subprocess.check_output(["cygpath", str(creator.dest)]).decode().strip()
```
gaborbernat, line 37: Do we need to call this subprocess command? Can't we compute from within python land?

danyeaw: @gaborbernat Thanks for your feedback, I was just trying to reuse the external tools that would be available to all users with cygwin and msys2 that have already solved this problem. We could do this in Python directly, I didn't realize the other platforms were going to fail tests because of this. I think the main change is going from:
"C:/msys64/home/"
to "/c/msys64/home/"
cygpath is smart and converts it to just "/", but that isn't really necessary since the full path works fine as well.
I could do this is a regex pretty easily, is that the approach you would take as well?

gaborbernat: If is a simple regex, then yes let's do it inline.

danyeaw:
@gaborbernat I have updated the PR with the regex 👍 
I noticed on the tests side that the activator isn't run in Windows because of WSL on GitHub Actions. Any thoughts on if additional tests could be added somehow?

gaborbernat:
> Any thoughts on if additional tests could be added somehow?

Write a uni test. Mock ``sysconfig.get_platform``, check generated scripts contain correct path.

danyeaw:
@gaborbernat Thanks for the feedback, I have been exploring how to structure mock tests since I don't have much experience mocking things out. I'm hoping to finish it this weekend. Can we reopen this PR?

gaborbernat:
@danyeaw I did not close this PR on purpose. Was a side-effect of renaming the default branch to main. Please reopen the PR against the main branch. Sorry.

danyeaw:
@gaborbernat No problem, I like the main branch rename! I have submitted #1952.

Issue #1959, opened 2020-10-01
python2: site .pth files are not processed if PYTHONPATH points to site-packages


navytux:
**Issue**

Hello up there. I've discovered that with python2 virtualenv 20 does not process .pth files under site-packages if $PYTHONPATH points to site-packages directory. It is a regression compared to virtualenv 16. This causes testing breakage of real projects, for example pygolong tox tests with new virtualenv become broken due to this issue.

Below is a brief demonstration for the problem:

```console
# create virtualenv with python2 and install zope.interfaces
kirr@deco:~/tmp/trashme$ virtualenv -p python2 x.venv
created virtual environment CPython2.7.18.final.0-64 in 134ms
  creator CPython2Posix(dest=/home/kirr/tmp/trashme/x.venv, clear=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/kirr/.local/share/virtualenv)
    added seed packages: pip==20.1.1, pkg_resources==0.0.0, setuptools==44.1.1, wheel==0.35.1
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator
kirr@deco:~/tmp/trashme$ unset PYTHONPATH
kirr@deco:~/tmp/trashme$ . x.venv/bin/activate
(x.venv) kirr@deco:~/tmp/trashme$ pip install zope.interface
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting zope.interface
  Using cached zope.interface-5.1.1-cp27-cp27mu-manylinux2010_x86_64.whl (232 kB)
Requirement already satisfied: setuptools in ./x.venv/lib/python2.7/site-packages (from zope.interface) (44.1.1)
Installing collected packages: zope.interface
Successfully installed zope.interface-5.1.1
```
```console
# make sure `import zope.interfaces` works
(x.venv) kirr@deco:~/tmp/trashme$ python
Python 2.7.18 (default, Apr 20 2020, 20:30:41) 
[GCC 9.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import zope.interface
>>> import zope
>>> zope.__path__
[...]

PR #1960, opened 2020-10-01
Process pth files even if $PYTHONPATH points to site-packages/

navytux:
Fixes https://github.com/pypa/virtualenv/issues/1959

- [ ] updated/extended the documentation  (not needed)

### DIFF
--- a/src/virtualenv/create/via_global_ref/builtin/python2/site.py
+++ b/src/virtualenv/create/via_global_ref/builtin/python2/site.py
@@ -52,8 +52,7 @@ def load_host_site():
     add_site_dir = sys.modules["site"].addsitedir
     for path in json.loads(site_packages):
         full_path = os.path.abspath(os.path.join(here, path.encode("utf-8")))
-        if full_path not in sys.path:
-            add_site_dir(full_path)
+        add_site_dir(full_path)
 
 
 sep = "\\" if sys.platform == "win32" else "/"  # no os module here yet - poor mans version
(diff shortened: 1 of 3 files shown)


src/virtualenv/create/via_global_ref/builtin/python2/site.py
```diff
@@ -52,8 +52,7 @@ def load_host_site():
     add_site_dir = sys.modules["site"].addsitedir
     for path in json.loads(site_packages):
         full_path = os.path.abspath(os.path.join(here, path.encode("utf-8")))
-        if full_path not in sys.path:
-            add_site_dir(full_path)
+        add_site_dir(full_path)
```
gaborbernat, line 55: This feels wrong to me. The path here is the last resort to ensure these paths are on the sys.path. Can we find out why the site package reload does not puts the python path onto the sys path and load the site-packages?

gaborbernat (changes requested):

src/virtualenv/create/via_global_ref/builtin/python2/site.py
navytux: @gaborbernat, I'm not sure I understand. In commit 1e869a97 you adjusted exactly this place to make sure pth files are processed _at_ _all_. That was on Ubuntu and I'm on Debian. Python maintainer for Ubuntu and Debian is the same person, so those packages are likely to have the same properties. Are you suggesting to me to find out why pth files were not processed before 1e869a97 and find a "more proper" fix for the problem?

gaborbernat (approved):

navytux:
Thanks

PR #1998, opened 2020-10-28
Fix `VIRTUALENV_PYTHON` environment lookup

pneff:
The previous change which introduced the fallback discovery (#1995) broke this behaviour.

I originally thought to change the `ListType` class which uses `self.as_type` to cast the individual list values into the target value. At the moment the default for that type is a `list` which means the default behaviour for the class is to return a list of lists. But that became a bigger change than I'm comfortable doing on this code base, so I came up with this simpler change.

### DIFF
--- a/tests/unit/config/test_env_var.py
+++ b/tests/unit/config/test_env_var.py
@@ -46,6 +46,20 @@ def test_python_multi_value_via_env_var(monkeypatch):
     assert options.python == ["python3", "python2"]
 
 
+def test_python_multi_value_newline_via_env_var(monkeypatch):
+    options = VirtualEnvOptions()
+    monkeypatch.setenv(str("VIRTUALENV_PYTHON"), str("python3\npython2"))
+    session_via_cli(["venv"], options=options)
+    assert options.python == ["python3", "python2"]
+
+
+def test_python_multi_value_prefer_newline_via_env_var(monkeypatch):
+    options = VirtualEnvOptions()
+    monkeypatch.setenv(str("VIRTUALENV_PYTHON"), str("python3\npython2,python27"))
+    session_via_cli(["venv"], options=options)
+    assert options.python == ["python3", "python2,python27"]
+
+
 def test_extra_search_dir_via_env_var(tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     value = "a{}0{}b{}c".format(os.linesep, os.linesep, os.pathsep)
--- a/tests/unit/config/test_env_var.py
+++ b/tests/unit/config/test_env_var.py
@@ -4,6 +4,7 @@ import os
 
 import pytest
 
+from virtualenv.config.cli.parser import VirtualEnvOptions
 from virtualenv.config.ini import IniConfig
 from virtualenv.run import session_via_cli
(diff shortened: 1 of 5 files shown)


tests/unit/config/test_env_var.py
```diff
@@ -31,6 +32,13 @@ def test_value_bad(monkeypatch, caplog, empty_conf):
     assert "invalid literal" in caplog.messages[0]
 
 
+def test_python_via_env_var(monkeypatch):
+    options = VirtualEnvOptions()
+    monkeypatch.setenv(str("VIRTUALENV_PYTHON"), str("python3"))
```
gaborbernat, line 37: Can you add a test to validate that ``python2,python3`` works as expected too?

gaborbernat (changes requested):

tests/unit/config/test_env_var.py
pneff: I extended the code to provide that functionality. Multi-value was split with newline so far, though it seems the only other multi-value field is `extra-search-dir`.

Let me know if you are happy with the approach taken or would prefer a different way to do this.

tests/unit/config/test_env_var.py
```diff
@@ -39,6 +39,13 @@ def test_python_via_env_var(monkeypatch):
     assert options.python == ["python3"]
 
 
+def test_python_multi_value_via_env_var(monkeypatch):
+    options = VirtualEnvOptions()
+    monkeypatch.setenv(str("VIRTUALENV_PYTHON"), str("python3,python2"))
```
gaborbernat, line 44: Wait, I did not realize that ``VIRTUALENV_EXTRA_SEARCH_DIR`` accepts newlines. Can we instead make this one use the same logic, and perhaps alter the logic for that to also accept the comma, not just newlines?

gaborbernat (changes requested):

tests/unit/config/test_env_var.py
gaborbernat: Actually, what we have is correct, but instead of having different split character depending on the type, let's always accept both. I don't expect people to use comma in paths, so IMHO, this is alright.

pneff: I am pushing a commit which implements this change. I opted for only accepting one split option at a time - so if the environment variable contains newlines, then we don't also accept commas. Do you agree with that decision?

I quickly went back to the legacy code as well, and it seems that newline support has only been added with the rewrite. The previous version had space separation for path names. Thus I would expect the backwards compatibility impact of this change to be minimal.

gaborbernat: That makes sense. Can you add a test demonstrating/validating this? Also, the documentation should explicitly state that only of the two will be accepted, and not both at the same time.

gaborbernat (changes requested):

tests/unit/config/test_env_var.py
pneff: Those tests have been added now.

gaborbernat (approved):

pneff:
Thank you @gaborbernat!

Issue #2003, opened 2020-11-01
Optionally skip creation of .gitignore in virtualenv directory


compwiztobe:
While version controlling the virtualenv binaries etc. is not desirable, some of us version control pyvenv.cfg as well as other metadata we store about each environment (such as a pip requirements file), to facilitate portability of venv definitions.  Before the merge of #1825 in response to #1806, that problem was already easily solvable with a top level .gitignore listing the typical Python bin/, lib/, share/, etc. directories to be ignored.

The `*` directive in the virtualenv subdirectories cannot be overridden by any .gitignore files in parent directories.  If we could pass an option to virtualenv to disable the creation of this blanket `*` .gitignore, then existing .gitignore files in the venv directory AND all parent directories would continue to work as expected, and we could continue to track chosen config files stored in each environment directory.

For now, here's a workaround:
```
if [[ ! -f "$venv/.gitignore" ]]; then
  virtualenv $venv
  rm $venv/.gitignore
else
  virtualenv $venv
fi
```
but I would much rather be able to do this with something like
```
virtualenv $venv --no-gitignore
```
to recover the old behavior.

gaborbernat:
@compwiztobe if you want to force adding some files you can just use:

```
git add -f venv/pyenv.cfg
```

The gitignore file is advisory for the add and can be ignored.

compwiztobe:
This is part of a de- and re-hydrating workflow with virtualenvwrapper, where all my venvs are in a venv repository I share between systems.  Creating a separate directory for the dried metadata would work but is messier (and might interfere with venv discovery tools and collide with existing venv names).

Forcibly tracking each file by hand means I have to do this for every file, and I still don't get `git diff` and `git status` to tell me when new environments or metadata needs indexing (for example when I first create or dehydrate a venv).  What I'm looking for is the ignore directives to discriminate between my dehydrated metadata and the hydrated venv binaries.

I'm hoping to upstream this stuff into virtualenvwrapper, including my workaround wrapper `premkvirtualenv`
```
if ! grep -q "# track venv metadata" "$1/.gitignore"; then
cat >> "$1/.gitignore" << EOF

# track venv metadata
!.gitignore
!pip-packages
!pythonexec
!pyvenv.cfg
EOF
fi
```

[...]

PR #2004, opened 2020-11-02
Optionally skip creation of .gitignore in virtualenv directory

compwiztobe:
Addresses #2003.  Creation of `.gitignore` with non-overridable * directive in the environment directory interferes with possibly more specific ignore directives in higher directories.  When the environment directory is a subdirectory of a repo and we need tracking of a subset of its files, allow to skip creation of this `.gitignore` with a command line option to `virtualenv`.

### DIFF
--- a/docs/changelog/2003.feature.rst
+++ b/docs/changelog/2003.feature.rst
@@ -1,0 +1,3 @@
+Optionally skip generation of ``.gitignore`` with * directive in virtualenv
+directory, using command line option :option:`no-gitignore`.
+Default behavior is to generate the file as before.
--- a/docs/changelog/2003.feature.rst
+++ b/docs/changelog/2003.feature.rst
@@ -1,3 +1,1 @@
-Optionally skip VCS ignore directive for entire virtualenv
-directory, using command line option :option:`no-vcsignore`.
-Default behavior is to generate the file as before.
+Optionally skip VCS ignore directive for entire virtualenv directory, using option :option:`no-vcs-ignore`, by default ``False``.
--- a/docs/changelog/2003.feature.rst
+++ b/docs/changelog/2003.feature.rst
@@ -1,3 +1,3 @@
-Optionally skip generation of ``.gitignore`` with * directive in virtualenv
-directory, using command line option :option:`no-gitignore`.
+Optionally skip VCS ignore directive for entire virtualenv
+directory, using command line option :option:`no-vcsignore`.
 Default behavior is to generate the file as before.
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -44,7 +44,7 @@ class Creator(object):
         self._debug = None
         self.dest = Path(options.dest)
         self.clear = options.clear
-        self.vcs_ignore = options.vcs_ignore
+        self.no_vcs_ignore = options.no_vcs_ignore
         self.pyenv_cfg = PyEnvCfg.from_folder(self.dest)
         self.app_data = options.app_data
 
@@ -58,7 +58,7 @@ class Creator(object):
         return [
             ("dest", ensure_text(str(self.dest))),
             ("clear", self.clear),
-            ("vcs_ignore", self.vcs_ignore),
+            ("no_vcs_ignore", self.no_vcs_ignore),
         ]
 
     @classmethod
@@ -94,10 +94,10 @@ class Creator(object):
         )
         parser.add_argument(
             "--no-vcs-ignore",
-            dest="vcs_ignore",
-            action="store_false",
+            dest="no_vcs_ignore",
+            action="store_true",
             help="don't create VCS ignore directive in the destination directory",
-            default=True,
+            default=False,
         )
 
     @abstractmethod
@@ -169,7 +169,7 @@ class Creator(object):
             safe_delete(self.dest)
         self.create()
         self.set_pyenv_cfg()
-        if self.vcs_ignore:
+        if not self.no_vcs_ignore:
             self.setup_ignore_vcs()
 
     def set_pyenv_cfg(self):
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -44,7 +44,7 @@ class Creator(object):
         self._debug = None
         self.dest = Path(options.dest)
         self.clear = options.clear
-        self.vcsignore = options.vcsignore
+        self.vcs_ignore = options.vcs_ignore
         self.pyenv_cfg = PyEnvCfg.from_folder(self.dest)
         self.app_data = options.app_data
 
@@ -58,7 +58,7 @@ class Creator(object):
         return [
             ("dest", ensure_text(str(self.dest))),
             ("clear", self.clear),
-            ("vcsignore", self.vcsignore),
+            ("vcs_ignore", self.vcs_ignore),
         ]
 
     @classmethod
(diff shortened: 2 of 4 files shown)


src/virtualenv/create/creator.py
```diff
@@ -160,7 +169,8 @@ def run(self):
             safe_delete(self.dest)
         self.create()
         self.set_pyenv_cfg()
-        self.setup_ignore_vcs()
```
gaborbernat, line 163: As you can see from the name of this function the feature is nowhere bound to git. Is bound to VCS in general. Please rename all code/flag naming accordingly.

gaborbernat (changes requested):

src/virtualenv/create/creator.py
compwiztobe: Except that the feature is only implemented for Git ignore functionality, and there is no possibility of implementation for other VCS in the foreseeable future, due to the designs of those systems?

Can certainly change to `--no-vcsignore` but wondering if maintaining that agnosticism towards VCS is justified.

gaborbernat: We cannot think of just today and the system we support just today. When designing API for a widely used tool, we must be future proof. So yes, it is justified.

> there is no possibility of implementation for other VCS in the foreseeable future, due to the designs of those systems

I don't think this to be the case.

docs/changelog/2003.feature.rst
```diff
@@ -0,0 +1,3 @@
+Optionally skip VCS ignore directive for entire virtualenv
+directory, using command line option :option:`no-vcsignore`.
```
gaborbernat, line 2: Please drop the ``command line`` words - as https://virtualenv.pypa.io/en/latest/cli_interface.html#defaults outlines this is just an option that's configurable from either env var, config file, or cli flag.

docs/changelog/2003.feature.rst
```diff
@@ -0,0 +1,3 @@
...
+Default behavior is to generate the file as before.
```
gaborbernat, line 3: You can drop this sentence, and just end the previous with: by default `False`.

src/virtualenv/create/creator.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, options, interpreter):
         self._debug = None
         self.dest = Path(options.dest)
         self.clear = options.clear
+        self.vcsignore = options.vcsignore
```
gaborbernat, line 47: Add separator between VCS and ignore to follow the existing style.

src/virtualenv/create/creator.py
```diff
@@ -90,6 +92,13 @@ def add_parser_arguments(cls, parser, interpreter, meta, app_data):
             help="remove the destination directory if exist before starting (will overwrite files otherwise)",
             default=False,
         )
+        parser.add_argument(
+            "--no-vcsignore",
+            dest="vcsignore",
+            action="store_false",
+            help="skip creation of coverall VCS ignore directive in destination directory",
```
gaborbernat, line 99: `don't create VCS ignore directive in the destination directory`

gaborbernat (changes requested):

src/virtualenv/create/creator.py
compwiztobe: yes it looks Italian doesn't it

gaborbernat: 😕

src/virtualenv/create/creator.py
```diff
@@ -90,6 +92,13 @@ def add_parser_arguments(cls, parser, interpreter, meta, app_data):
             help="remove the destination directory if exist before starting (will overwrite files otherwise)",
             default=False,
         )
+        parser.add_argument(
+            "--no-vcs-ignore",
+            dest="vcs_ignore",
+            action="store_false",
+            help="don't create VCS ignore directive in the destination directory",
+            default=True,
```
compwiztobe, line 100: Does this break convention and more importantly the CLI option-environment variable interop, by renaming the argument?  I wasn't aware of the config-file/env-var pathway to set these, and now it's not clear to me if those will expect a value at `no_vcs_ignore`/`VIRTUALENV_NO_VCS_IGNORE` or `vcs_ignore`/`VIRTUALENV_VCS_IGNORE` or some combo.

src/virtualenv/create/creator.py
```diff
@@ -90,6 +92,13 @@ def add_parser_arguments(cls, parser, interpreter, meta, app_data):
             help="remove the destination directory if exist before starting (will overwrite files otherwise)",
             default=False,
         )
+        parser.add_argument(
+            "--no-vcs-ignore",
```
gaborbernat, line 96: ``VIRTUALENV_NO_VCS_IGNORE`` and ``no_vcs_ignore`` will be the names as documented under https://virtualenv.pypa.io/en/latest/cli_interface.html#environment-variables. The destination variable doesn't matter only the CLI flag.

src/virtualenv/create/creator.py
gaborbernat: So if you put:

```
[virtualenv]
no_vcs_ignore = false
```

in your users ``virtualenv.ini`` file you enabled this flag always on your machine.

src/virtualenv/create/creator.py
```diff
@@ -90,6 +92,13 @@ def add_parser_arguments(cls, parser, interpreter, meta, app_data):
...
+            dest="vcs_ignore",
```
gaborbernat, line 97: I don't like how you negated the flag name, but kept the destination positive. Please keep them in sync. This flag should be ``--no-vcs-ignore``, destination `no_vcs_ignore` and should `store_true` with default `False`.

gaborbernat (changes requested):

src/virtualenv/create/creator.py
compwiztobe: Perfect

src/virtualenv/create/creator.py
compwiztobe: Done.  Wasn't really any reason for old way, other than the double negative on [line 172](https://github.com/pypa/virtualenv/pull/2004/files#diff-f055c70dd99867f8264dfbeb8f7b54b7355be4c3dc36ae576ba06940fa316d12R172), and "because I can".  But bringing into sync also will make the default value in auto-docs less confusing (actually correct).

pradyunsg:
I usually find is really awkward to have double-negatives for configuration of tooling (pip is HORRIBLE on this front).

Here's my 2 cents: if we're doing this, let's add `--vcs-ignore/--no-vcs-ignore` (I don't care about the exact naming) such that the configuration options look like `vcs_ignore = false`. By having two knobs for booleans, we'd also allow folks to override environment-level "no" with a `--yes` CLI flag.

gaborbernat:
> Here's my 2 cents: if we're doing this, let's add `--vcs-ignore/--no-vcs-ignore` (I don't care about the exact naming) such that the configuration options look like `vcs_ignore = false`. By having on/off knobs for the boolean, we'd also allow folks to override environment-level "no" with a "yes" CLI flag.

Supporting ``--flag/--no-flag`` would be a generic feature we don't do at the moment and I'm personally not totally sold on it (or interested in implementing it). Either way, would this be done I would not single out a new flag, but implement it for all boolean flags, which is clearly out of the scope of this PR.

gaborbernat (approved):

PR #2008, opened 2020-11-13
conftest.py: Make coverage optional

hroncok:
This allows to execute the tests without coverage installed.
We plan to do that in Fedora, where we run the tests, but not measure coverage.

   - does not apply here
   - this is trivial
   - does not apply here


hroncok:
Thanks

PR #2015, opened 2020-11-19
Add --seed-packages.

AllSeeingEyeTolledEweSew:
Hello,

I added a `--seed-packages` option to the builtin seeders. This tells them to seed additional packages, with the intent being to provide a way to bootstrap any packages required to make `pip` work.

My use case is that my machine can only access the internet through a socks proxy (in fact I use [tails](https://tails.boum.org), which routes all internet traffic through tor). This causes problems for most virtualenvs, because `pip` requires [PySocks](https://pypi.org/project/PySocks) to download anything, so I have to bootstrap every virtualenv with a `.whl` file downloaded from a web browser.

This is fine for my own virtualenvs, but tools like `tox` and `pre-commit` create virtualenvs with little opportunity to customize bootstrapping. See pre-commit/pre-commit#1697 for some discussion.

I considered writing a custom seeder plugin, but I expect that virtualenv-creating tools would override my plugin with `--seeder`, based on considerations of *how* to seed, rather than *what* to seed. So I thought it was better to modify the builtin seeders.

For my case, I can do
```sh
export VIRTUALENV_SEED_PACKAGE=PySocks
export VIRTUALENV_EXTRA_SEARCH_DIR=/path/to/wheel
```
then, `pre-commit` and `tox` work normally everywhere.

I'm not sure if this is immediately useful for other cases, but I see a lot of try-import blocks in `pip`'s source code, so I expect it's useful to someone somewhere.

This PR is still WIP while I add tests, but I thought I should open it now for discussion purposes.

### DIFF
--- a/src/virtualenv/seed/embed/base_embed.py
+++ b/src/virtualenv/seed/embed/base_embed.py
@@ -13,6 +13,13 @@ from ..wheels import Version
 PERIODIC_UPDATE_ON_BY_DEFAULT = True
 
 
+def _parse_basic_req(line):
+    parts = line.split("==")
+    if len(parts) == 1:
+        return (line, None)
+    return (parts[0], parts[1])
+
+
 @add_metaclass(ABCMeta)
 class BaseEmbed(Seeder):
     def __init__(self, options):
@@ -30,6 +37,7 @@ class BaseEmbed(Seeder):
         self.no_wheel = options.no_wheel
         self.app_data = options.app_data
         self.periodic_update = not options.no_periodic_update
+        self.seed_package = options.seed_package
 
         if not self.distribution_to_versions():
             self.enabled = False
@@ -43,11 +51,14 @@ class BaseEmbed(Seeder):
         }
 
     def distribution_to_versions(self):
-        return {
-            distribution: getattr(self, "{}_version".format(distribution))
-            for distribution in self.distributions()
-            if getattr(self, "no_{}".format(distribution)) is False
-        }
+        distributions = dict(self.seed_package)
+        for distribution in self.distributions():
+            if getattr(self, "no_{}".format(distribution)):
+                distributions.pop(distribution, None)
+            else:
+                version = getattr(self, "{}_version".format(distribution))
+                distributions[distribution] = version
+        return distributions
 
     @classmethod
     def add_parser_arguments(cls, parser, interpreter, app_data):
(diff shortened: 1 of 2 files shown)


src/virtualenv/seed/embed/base_embed.py
```diff
@@ -98,6 +109,15 @@ def add_parser_arguments(cls, parser, interpreter, app_data):
             help="disable the periodic (once every 14 days) update of the embedded wheels",
             default=not PERIODIC_UPDATE_ON_BY_DEFAULT,
         )
+        parser.add_argument(
```
gaborbernat, line 112: IMHO this should be a separate plugin. This needs a lot more thought, on e.g. how this interacts with periodic update feature, transitive dependencies and where do we get these custom wheels you're injecting here. Perhaps you're after a pip functionality here, not virtualenv one. In general would be nice to open an issue to discuss such proposal before you lead on with a solution.

gaborbernat (changes requested):

src/virtualenv/seed/embed/base_embed.py
AllSeeingEyeTolledEweSew: > IMHO this should be a separate plugin. This needs a lot more thought, on e.g. how this interacts with periodic update feature, transitive dependencies and where do we get these custom wheels you're injecting here. Perhaps you're after a pip functionality here, not virtualenv one. In general would be nice to open an issue to discuss such proposal before you lead on with a solution.

Should the discussion be here? I find it's often useful to propose a problem and solution together.

The wheels are supplied externally, hence my `VIRTUALENV_EXTRA_SEARCH_DIR`.

It shouldn't interact with periodic update; AFAICT the intent of that is only for the embedded wheels (maybe I should ensure that code is explicitly restricted to those).

Based on #1923 it seems like virtualenv won't support traversing dependencies for now. Until it does, you need to specify your whole dependency set in `--seed-package`, but this seems fine for now.

Updating pip would solve my problem as well; I thought that `--seed-package` was a smaller ask than `pip` vendoring more code. I thought they would want to be as lightweight as possible. If you think I'm more likely to succeed with pip, I can try there instead.

gaborbernat: > Based on #1923 it seems like virtualenv won't support traversing dependencies for now. 

This is not true, just waiting for implementation. I'd consider this a blocking functionality request for this PR.

>  Until it does, you need to specify your whole dependency set in --seed-package, but this seems fine for now.

I disagree.

> Updating pip would solve my problem as well; I thought that `--seed-package` was a smaller ask than `pip` vendoring more code.

Wishful thinking. 😊 

If you want to go ahead with the current works for your half-baked solution I recommend using a plugin, but for acceptance in the core, we'd need at the very least traversing and handling dependencies too.

AllSeeingEyeTolledEweSew: > If you want to go ahead with the current works for your half-baked solution I recommend using a plugin, but for acceptance in the core, we'd need at the very least traversing and handling dependencies too.

Fair enough! thanks for your time.

PR #2026, opened 2020-11-28
Switch documentation theme to Furo

pradyunsg:
@gaborbernat wanted a PR, to see how this looks. :)


gaborbernat:
Closing until some of the concerns raised offline are addressed.

pradyunsg:
FWIW, I don't think any of those concerns would be addressed anytime soon.

All of those are stylistic decisions made as part of Furo's design that I won't be changing in Furo; and some are a bit more fundamental to how we disagree on this documentation should work (eg: tables for options and optimising desktop-like resolutions to look good on mobile).

This makes a little sad, because from the feedback that I've seen, folks do prefer Furo's styling over how most PyPA documentation looks. None the less, Sphinx-RTD-theme is pretty great, and I disable the (IMO bad) custom styling that's injected as part of virtualenv's docs anyway -- so meh. :)

gaborbernat:
>  the feedback that I've seen, folks do prefer Furo's styling over how most PyPA documentation looks.

This statement feels anecdotal and I don't think can be considered without a proper survey. For once I'm not sure virtualenv is how most PyPa documentation looks, and for second people tend to write their praises in public rather than their improvement suggestions (which IMHO is a good thing). Most points I've raised for me feels ends up being rendered not as easy to read as the current one, and I'd love you to chime in them individually (or other maintainers/readers) to reach a conclusion, rather than entirely abandon the topic and me seemingly being unreasonably stubborn.

PR #2035, opened 2020-12-12
Bugfix: PR #2034 - Fix Segfault 11 on OS X:ksh93

cshilton:
With the embedded `unset -f deactivate` call, the deactivate function
segfaulted on Mac OS X ksh93. This patch removes the `unset -f` call
from the function which fixes the segfault. It restores the
functionality by hiding the function behind an underscore, `deactivate`
-> `_deactivate` and then uses an alias to get rid of both the function
and the alias.

### DIFF
--- a/src/virtualenv/activation/bash/activate.sh
+++ b/src/virtualenv/activation/bash/activate.sh
@@ -7,7 +7,7 @@ if [ "${BASH_SOURCE-}" = "$0" ]; then
     exit 33
 fi
 
-deactivate () {
+_deactivate () {
     unset -f pydoc >/dev/null 2>&1
 
     # reset old environment variables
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
I think calling unset is not unrealistic to do. IMHO here the bugfix should be within ksh93, and not within virtualenv.

cshilton:
I'm not ever sure if this is a ksh93 issue? It could be whatever malloc() and free() that OS X is using. But, I defer. I do wish there was a shorter option to deactivate than `nondestructive`. That's a lot more than `deactivate -n` for example to commit to finger memory.

cshilton:
A casual search for this issue in the ksh93 github repo says that the ksh93 people agree with you. Sorry to take up part of your Saturday. Thanks for virtualenv!

gaborbernat:
No problem, thanks!

PR #2049, opened 2021-01-12
Allow unfunctioing pydoc to fail

d125q:
Otherwise, one cannot activate a virtualenv under Zsh with set -e.

As an example,
```shell-session
$ zsh -c 'set -ex; source /path/to/virtualenv/bin/activate; echo activated'
+zsh:1> source /path/to/virtualenvbin/activate
+/path/to/virtualenvbin/activate:5> [ '' '=' /path/to/virtualenvbin/activate ']'
+/path/to/virtualenvbin/activate:47> deactivate nondestructive
+deactivate:1> unset -f pydoc
```
fails with exit code 1 because there is nothing to unset.  (The relevant version are: zsh 5.8.0.2-dev (x86_64-pc-linux-gnu), Python 3.9.1, and virtualenv 20.3.0.)

### DIFF
--- a/docs/changelog/2049.bugfix.rst
+++ b/docs/changelog/2049.bugfix.rst
@@ -1,0 +1,4 @@
+Allow unfunctioning of pydoc to fail freely so that virtualenvs can be
+activated under Zsh with set -e (since otherwise ``unset -f`` and
+``unfunction`` exit with 1 if the function does not exist in Zsh) - by
+:user:`d125q`.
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
Please add a changelog entry, thanks!

d125q:
> Please add a changelog entry, thanks!

Added; thanks likewise!

gaborbernat (approved):

gaborbernat:
Released now via https://virtualenv.pypa.io/en/latest/changelog.html#v20-3-1-2021-01-13

PR #2050, opened 2021-01-12
Clear stash if system executable is missing

gaborbernat:


gaborbernat:
Released now via https://virtualenv.pypa.io/en/latest/changelog.html#v20-3-1-2021-01-13

Issue #2071, opened 2021-03-01
virtualenv PyPy Windows creates wrong path


gaborbernat:
https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.7/lib-python/3/venv/__init__.py#L128-137

```
 venv3.6
├──  lib
│  └──  python3.6
│     └──  site-packages
├──  pyvenv.cfg
```

vs 

```
 emv3.6
├──  Include
├──  Lib
│  ├──  site-packages
│  └──  tcl8.5
│     └──  init.tcl
├──  pyvenv.cfg
```

https://github.com/pypa/virtualenv/blob/main/src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py#L26

A PR fixing this is welcomed.

PR #2072, opened 2021-03-01
Fix PyPy3 Windows stlib path

gaborbernat:
Closes https://github.com/pypa/virtualenv/issues/2071


altendky:
@gaborbernat, thanks for this.  I took a different path in towncrier that didn't expose this issue but I'll give it a go in twisted now.

Issue #2048, opened 2021-01-12
`pip_cert` fixture breaks tests if `PIP_CERT` is in os.environ


frenzymadness:
The new fixture `pip_cert` implemented in https://github.com/pypa/virtualenv/commit/bc787b2da64ab3dbe5bb9c5966ff411dc8a41705 does not work if `PIP_CERT` is in os.environ. Pytest expect it to yield a value because it's recognized as a generator but if `PIP_CERT` is present in os.environ, the fixture returns `None` which causes pytest to fail with:
```
fixturefunc = <function pip_cert at 0x7f14f12ef3a0>
request = <SubRequest 'pip_cert' for <Function test_python[no_prompt]>>
kwargs = {'tmp_path_factory': TempPathFactory(_given_basetemp=None, _trace=<pluggy._tracing.TagTracerSub object at 0x7f14f0197490>, _basetemp=PosixPath('/tmp/pytest-of-mockbuild/pytest-0'))}

    def call_fixture_func(
        fixturefunc: "_FixtureFunc[_FixtureValue]", request: FixtureRequest, kwargs
    ) -> _FixtureValue:
        if is_generator(fixturefunc):
            fixturefunc = cast(
                Callable[..., Generator[_FixtureValue, None, None]], fixturefunc
            )
            generator = fixturefunc(**kwargs)
            try:
                fixture_result = next(generator)
            except StopIteration:
>               raise ValueError(
                    "{} did not yield a value".format(request.fixturename)
                ) from None
E               ValueError: pip_cert did not yield a value

fixturefunc = <function pip_cert at 0x7f14f12ef3a0>
generator  = <generator object pip_cert at 0x7f14ed790580>
kwargs     = {'tmp_path_factory': TempPathFactory(_given_basetemp=None, _trace=<pluggy._tracing.TagTracerSub object at 0x7f14f0197490>, _basetemp=PosixPath('/tmp/pytest-of-mockbuild/pytest-0'))}
request    = <SubRequest 'pip_cert' for <Function test_python[no_prompt]>>

/usr/lib/python3.9/site-packages/_pytest/fixtures.py:917: ValueError

```

An easy fix might be to use the condition in the decorator like `@pytest.fixture(autouse=ensure_str("PIP_CERT") not in os.environ, scope="session")`.

hroncok:
What about changing the return to yield and moving the rest of the function to an elif branch?

[...]

PR #2083, opened 2021-03-17
Fix pip_cert pytest fixture

frenzymadness:
pip_cert is recognized as a generator so it should not use
return if PIP_CERT is in os.environ.

Fixes: https://github.com/pypa/virtualenv/issues/2048

I'm not sure this is testable easily given the fact that a CA cert bundle might be in various locations. Is a change in tests worth mentioning in the changelog?

Before this change:
```
$ PIP_CERT=/etc/ssl/certs/ca-bundle.crt tox -e py39 -- -x
…
ValueError: pip_cert did not yield a value
```

after this change:
```
$ PIP_CERT=/etc/ssl/certs/ca-bundle.crt tox -e py39 -- -x
…
  py39: commands succeeded
  congratulations :)
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -153,14 +153,15 @@ def pip_cert(tmp_path_factory):
     # workaround for https://github.com/pypa/pip/issues/8984 - if the certificate is explicitly set no error can happen
     key = ensure_str("PIP_CERT")
     if key in os.environ:
-        return
-    cert = tmp_path_factory.mktemp("folder") / "cert"
-    import pkgutil
-
-    cert_data = pkgutil.get_data("pip._vendor.certifi", "cacert.pem")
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Can you add some tests to this, please?

frenzymadness:
> Can you add some tests to this, please?

Do you know about a universal CA bundle I can use across all the systems you have in CI? It needs to be set before pytest is executed so I can imagine a new tox environment running only some tests with this environment variable set.

Is it worth it given the fact that the whole fixture is just a workaround for a problem in pip and its author did not add tests for it?

I can also imagine some dummy test exercising just the unused part of the fixture and actually testing nothing.

gaborbernat (approved):

PR #2086, opened 2021-03-31
Remove unnecessary special cases for bash/zsh

orent:
The internal shell utility hash is standard posix. There is no need to special-case it only for bash and zsh.
In fact, this check can break the script on shells such as the built-in sh on BSD systems, ksh, ash or 
busybox sh which are otherwise perfectly capable of running it. The script is a valid posix shell script, as
verified by shellcheck. The one feature that depends on bash (checking that the script is sourced rather 
than executed) will work only on bash, but it is only a safety check.

 https://pubs.opengroup.org/onlinepubs/9699919799/utilities/hash.html

### DIFF
--- a/src/virtualenv/activation/bash/activate.sh
+++ b/src/virtualenv/activation/bash/activate.sh
@@ -23,12 +23,10 @@ deactivate () {
         unset _OLD_VIRTUAL_PYTHONHOME
     fi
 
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

gaborbernat:
Released via https://pypi.org/project/virtualenv/20.4.4/

PR #2088, opened 2021-04-08
Built in discovery class is always preferred over plugin supplied classes

### LINKED ISSUE #2087: Use the builtin plugin classes unless another plugin class is specifically asked for.
**What's the problem this feature will solve?**

I have a private plugin for virtualenv where I add an additional discovery class. This discovery class adds a new argument to the command line parser that is required but only when the discovery class is chosen. However I'm seeing an issue where using virtualenv via the command line as normal is now asking for this argument. The reason seems to be that virtualenv is picking a default discovery class but in a non-deterministic way and sometimes the additional discovery class is chosen as the default discovery class and so the argument is required. The default class is chosen depending on which entry point is discovered first. I believe entry points give no guarantees about order of discovery.

The order of entry points discovery seems to change in different installs of virtualenv and the plugin, rather than changing in the same environment between different invocations of virtualenv.

I believe the problem will be the same for creators, seeders, and activators as well.

**Describe the solution you'd like**

I would expect the builtin discovery class to be chosen as the default discovery class unless explicitly set otherwise.

**Alternative Solutions**

These classes could have a priority set at the class level. The builtin classes would have a priority set such that a plugin class could opt to set it's priority lower or higher than the builtins. virtualenv would then order these classes by their priority. Classes would be allowed to have the same priority with the understanding that the order of classes with the same priority value would be non-deterministic.

AWhetter:

This ensures that the builtin Discovery class is always chosen over any discovery class provided by plugins, as discussed in #2087

Closes #2087

### DIFF
--- a/src/virtualenv/run/plugin/discovery.py
+++ b/src/virtualenv/run/plugin/discovery.py
@@ -13,10 +13,13 @@ def get_discover(parser, args):
         title="discovery",
         description="discover and provide a target interpreter",
     )
+    choices = _get_default_discovery(discover_types)
+    # prefer the builtin if present, otherwise fallback to first defined type
+    choices = sorted(choices, key=lambda a: 0 if a == "builtin" else 1)
     discovery_parser.add_argument(
         "--discovery",
-        choices=_get_default_discovery(discover_types),
-        default=next(i for i in discover_types.keys()),
+        choices=choices,
+        default=next(iter(choices)),
         required=False,
         help="interpreter discovery method",
     )
--- a/docs/changelog/2087.bugfix.rst
+++ b/docs/changelog/2087.bugfix.rst
@@ -1,0 +1,1 @@
+Built in discovery class is always preferred over plugin supplied classes.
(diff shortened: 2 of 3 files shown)


src/virtualenv/run/plugin/discovery.py
```diff
@@ -13,10 +13,13 @@ def get_discover(parser, args):
         title="discovery",
         description="discover and provide a target interpreter",
     )
+    choices = _get_default_discovery(discover_types)
+    # prefer the builtin if present, otherwise fallback to first defined type
+    choices = sorted(choices, key=lambda a: 0 if a == "builtin" else 1)
```
AWhetter, line 18: This these lines have been borrowed from here: https://github.com/pypa/virtualenv/blob/0262fa6d9a0f8b45c80e6b2f82109ab1250ec336/src/virtualenv/run/plugin/creators.py#L52-L53

gaborbernat (approved):

gaborbernat:
Released via https://pypi.org/project/virtualenv/20.4.4/

PR #2107, opened 2021-05-04
Fix site.getsitepackages() ignoring --system-site-packages on python2

### LINKED ISSUE #2106: site.getsitepackages() doesn't respect --system-site-packages on python2
**Issue**

site.getsitepackages() doesn't respect --system-site-packages being set on python2. System site-package paths are never included. 
I came across this while working on #2105. In contrast to #2105 this is not specific to debian, which is why I am creating a separate issue for it.

```
python -m virtualenv --system-site-packages -ppython2 testenv
. testenv/bin/activate
python -c "import site; print(site.getsitepackages())"
['/home/adrian/Documents/dev/uni/bachelor/mypy/testenv/lib/python2.7/site-packages', '/home/adrian/Documents/dev/uni/bachelor/mypy/testenv/lib/site-python']
```

**Environment**

Provide at least:
- OS: Tested on both Ubuntu 18.04.3 LTS and Arch Linux
- ``pip list`` of the host python where ``virtualenv`` is installed:

Same as #2105 

Same as #2105

freundTech:
Fixes #2106.
This PR also adds a regression test.
There might be a merge conflict between this PR and #2108, because they both add a test to the end of tests/unit/create/test_creator.py

### DIFF
--- a/docs/changelog/2106.bugfix.rst
+++ b/docs/changelog/2106.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``site.getsitepackages()`` ignoring ``--system-site-packages`` on python2 - by :user:`freundTech`.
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

freundTech:
I noticed that this PR sometimes leads to failing tests. I manually verified that the change is correct, but the test added for it might have a bug or a race-condition. I'm looking into it.

PR #2109, opened 2021-05-04
Legacy: Use sysconfig.get_default_scheme() where available

hroncok:
sysconfig._get_default_scheme() is gone on Python 3.10

(We support Fedora 32 for another 3 weeks and we still have legacy virtualenv there. I will patch this in, but I wanted to share it. Not tested yet.)


hroncok:
The bot is changing irrelevant things here :/

gaborbernat (changes requested):
You'll need to fix the CI here.

hroncok:
Do you want me to add a style commit first, or to disable the pre-commit CI on the legacy branch?

gaborbernat:
Definitely don't disable the style check. We need to understand why it's triggered here and not on the master (or other) branches.

gaborbernat:
It's ok to disable pre-commit here.

hroncok:
Thanks.

PR #2141, opened 2021-07-11
add libffi-7.dll to pypy on windows

mattip:

This is a continuation of [PyPy issue 1922](https://foss.heptapod.net/pypy/pypy/-/issues/1922) where virtualenv has the shared objects needed to run PyPy hardcoded, and PyPy has not yet adopted CPython's launchers on windows. I am not sure how best to solve this. I was hoping a change in PyPy's venv module would have helped, but it seems not.

The immediate problem is that PyPy wishes to add `libffi-7.dll` to the windows build for the next release (in 6 weeks or so?). I have tested this fix locally for the [yet-as-unmerged-to-master change](https://foss.heptapod.net/pypy/pypy/-/tree/branch/libffi-win64) in PyPy's sources.


src/virtualenv/create/via_global_ref/builtin/pypy/pypy2.py
```diff
@@ -112,7 +112,7 @@ def modules(cls):
 
     @classmethod
     def _shared_libs(cls):
-        return ["libpypy-c.dll"]
+        return ["libpypy-c.dll", "libffi-7.dll"]
```
gaborbernat, line 115: Wouldn't this need to be version limited to newer pypy versions?

gaborbernat (changes requested):

src/virtualenv/create/via_global_ref/builtin/pypy/pypy2.py
mattip: no, the code to copy ignores missing files

gaborbernat: Oh, 🤔 wonder why 😆

mattip: https://github.com/pypa/virtualenv/blob/17caadc974cb2a24f70374525596cd1160579594/src/virtualenv/create/via_global_ref/builtin/pypy/common.py#L45-L49

mattip:
Is the pypy2 windows test run known to be flaky or did I mess something up?

docs/changelog/2141.misc.rst
```diff
@@ -0,0 +1 @@
+add libffi-7.dll to the hard-coded list of dlls for PyPy
```
gaborbernat, line 1: Should start with capital and end with a period.

src/virtualenv/create/via_global_ref/builtin/pypy/pypy2.py
gaborbernat: Will it work without this DLL? Should we make it hard dependency rather than soft that it is now?

gaborbernat (changes requested):

docs/changelog/2141.misc.rst
mattip: fixed

src/virtualenv/create/via_global_ref/builtin/pypy/pypy2.py
mattip: Currently PyPy does not have the extra DLL, so making it required will break backward comatibility.

gaborbernat: Currently this DLL is soft depdency (so assuming that's not a bug), if in the future this is not the case we should make it hard dependency and version guard the hard dependency part.

mattip: It would be nice to avoid this code altogether in favor of a more robust solution

gaborbernat: That wasn't my question. Can we say a usecase where pypy would work without this DLL with the current less robust solution? If yes, we can keep it a soft dependency. If not, it must be hard depedency.

mattip: I am not sure what you mean, sorry. PyPy v7.3.5 will not have the libffi-7.dll and will work without it. Once the branch to add the dll gets merged, any future PyPy versions will require the dll. I am hopeful that branch will be merged before the PyPy v7.3.6 release, but it is dependent on a new release of virtualenv with this PR or one similar.

mattip: On the other hand, I can envision a scenario where a packager like conda uses some [OS-level utilities](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order) to avoid copying the dll. In this case even though the dlls are required, they are not copied by this code into the new virtualenv.

mattip:
Thanks

gaborbernat:
Released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-5-0-2021-07-13

mattip:
Wow, that was fast, thanks!

PR #2142, opened 2021-07-12
Use platformdirs instead of appdirs

gaborbernat:


gaborbernat:
Blocked on https://github.com/pypa/virtualenv/pull/2143

PR #2143, opened 2021-07-12
Drop Python 3.4 support

gaborbernat:


xmo-odoo:
@gaborbernat this apparently made tox stop working entirely for 3.4 because tox apparently now decides that the seed packages for 3.4 are pip==21.1.3, setuptools==57.1.0, wheel==0.36.2.

Is there a way to make this work again however rickety? Not to be an officially supported configuration, just a way to tell tox to either not do anything or... use whatever version of those fit the interpreter? `pip install -U pip` does install a version compatible with 3.4.

xmo-odoo:
Whoops sorry about that I did not notice I had crossed over from tox to virtualenv.

Issue #1910, opened 2020-07-29
Support Pythons without distutils


gaborbernat:
At the moment interpreter lookup fails when ``distutils`` is not available, e.g. on some Ubuntu Python distributions. Error:

```
113 failed to query /usr/bin/python3.6 with code 1 err: 'Traceback (most recent call last):\n  File "/home/manos/.local/lib/python3.8/site-packages/virtualenv/discovery/py_info.py", line 16, in <module>\n    from distutils.command.install import SCHEME_KEYS\nModuleNotFoundError: No module named \'distutils.command\'\n' [INFO cached_py_info:38]
```

See https://github.com/pypa/virtualenv/issues/1643#issuecomment-665691388.

hynek:
To help desperate Googlers (me, yesterday): if you get `RuntimeError: failed to find interpreter for Builtin discover of python_spec='3.9'` on Ubuntu or debian and you positively know that the interpreter is installed, it might be that you're missing `python3.9-distutils`.

jamesjjcondon:
FYi https://bugs.launchpad.net/ubuntu/+source/python-distutils-extra/+bug/1892159

jamesjjcondon:
I'm having a similar issue trying to create a virtualenv with python3.7 on Ubuntu 20.04. 
Not missing python3.9-distutils. and python3.7-distutils not available. Above didn't help.

Have got working on machine a) but not b). verbose traces below. Seems to be related to cpython version and proposed python path.

machine a)
![image](https://user-images.githubusercontent.com/33615047/108798713-69042180-75de-11eb-81f6-c4f5b8397ce6.png)

machine b)
![image](https://user-images.githubusercontent.com/33615047/108798527-e4b19e80-75dd-11eb-85f1-4aa976c32701.png)

Any pointers @gaborbernat ?

jamesjjcondon:
Don't know how but removing and purging python3.7 and reinstalling from deadsnakes seems to have fixed

freundTech:
Note that python 3.10 will deprecate distutils. Python 3.12 will completely remove it.
See https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated

urkh:
For those who have a similar problem with python 3.7 or lower and use a .deb distro, the solution is to install the `python*-distutils` package according to the version of python that is failing. A quick and easy option is to add deadsnakes ppa and from there install the required python version along with distutils

[...]

PR #2146, opened 2021-07-13
Support without distutils

gaborbernat:


gaborbernat:
Resolves https://github.com/pypa/virtualenv/issues/1910.

tests/unit/discovery/test_discovery.py
```diff
@@ -27,7 +27,7 @@ def test_discovery_via_path(monkeypatch, case, tmp_path, caplog, session_app_dat
     elif case == "upper":
         name = name.upper()
     exe_name = "{}{}{}".format(name, current.version_info.major, ".exe" if sys.platform == "win32" else "")
-    target = tmp_path / current.distutils_install["scripts"]
+    target = tmp_path / current.install_path("scripts")
```
Ark-kun, line 30: Can we test both execution flows paths of `install_path`?

gaborbernat: What do you mean?

PR #2168, opened 2021-08-10
Bump pip and wheel

gaborbernat:


harshkumar-dev:
Hi @gaborbernat

Is there a way to install a particular version of virtualenv? on 
https://virtualenv.pypa.io/en/latest/installation.html#via-zipapp, I only see https://bootstrap.pypa.io/virtualenv.pyz 
For example, is there a url for 20.7.2? I am looking for a link that is python version independent.

gaborbernat:
You can omit the version from the link, e.g. https://github.com/pypa/get-virtualenv/blob/20.4.1/public/virtualenv.pyz

harshkumar-dev:
Thank you!

PR #2170, opened 2021-08-16
Nushell activation scripts

elferherrera:
Hi all,

I'm including a brand new activation script for nushell. I think I've got the base structure but I was wondering if someone could help me set up some things. 

- The first issue is that because of the nature of nushell, a deactivation function cannot be created. In order to deactivate an environment we will need to source a deactivate script (included with this PR). How can I check that that file is created together with the `activate.nu` file?
- Can someone help me set up the tests for activating the scripts for nushell?

### DIFF
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -17,7 +17,6 @@ class NushellActivator(ViaTemplateActivator):
         # Due to nushell scoping, it isn't easy to create a function that will
         # deactivate the environment. For that reason a __DEACTIVATE_PATH__
         # replacement pointing to the deactivate.nu file is created
-        #
 
         return {
             "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
@@ -25,5 +24,5 @@ class NushellActivator(ViaTemplateActivator):
             "__VIRTUAL_NAME__": creator.env_name,
             "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
             "__PATH_SEP__": ensure_text(os.pathsep),
-            "__DEACTIVATE_PATH__": ensure_text(os.path.join(dest_folder, "deactivate.nu")),
+            "__DEACTIVATE_PATH__": ensure_text(str(Path(dest_folder) / Path("deactivate.nu"))),
         }
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -1,6 +1,8 @@
 from __future__ import absolute_import, unicode_literals
 
+import os
 from virtualenv.util.path import Path
+from virtualenv.util.six import ensure_text
 
 from ..via_template import ViaTemplateActivator
 
@@ -9,3 +11,19 @@ class NushellActivator(ViaTemplateActivator):
     def templates(self):
         yield Path("activate.nu")
         yield Path("deactivate.nu")
+
+    def replacements(self, creator, dest_folder):
+        # Due to nushell scoping, it isn't easy to create a function that will
+        # deactivate the environment. For that reason a __DEACTIVATE_PATH__
+        # replacement pointing to the deactivate.nu file is created
+        virtual_env = ensure_text(str(creator.dest))
+        bin_name = ensure_text(str(creator.bin_dir.relative_to(creator.dest)))
+
+        return {
+            "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
+            "__VIRTUAL_ENV__": virtual_env,
+            "__VIRTUAL_NAME__": creator.env_name,
+            "__BIN_NAME__": bin_name,
+            "__PATH_SEP__": ensure_text(os.pathsep),
+            "__DEACTIVATE_PATH__": str(Path(virtual_env, bin_name, "deactivate.nu")),
+        }
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -16,14 +16,12 @@ class NushellActivator(ViaTemplateActivator):
         # Due to nushell scoping, it isn't easy to create a function that will
         # deactivate the environment. For that reason a __DEACTIVATE_PATH__
         # replacement pointing to the deactivate.nu file is created
-        virtual_env = ensure_text(str(creator.dest))
-        bin_name = ensure_text(str(creator.bin_dir.relative_to(creator.dest)))
 
         return {
             "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
-            "__VIRTUAL_ENV__": virtual_env,
+            "__VIRTUAL_ENV__": ensure_text(str(creator.dest)),
             "__VIRTUAL_NAME__": creator.env_name,
-            "__BIN_NAME__": bin_name,
+            "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
             "__PATH_SEP__": ensure_text(os.pathsep),
-            "__DEACTIVATE_PATH__": str(Path(virtual_env, bin_name, "deactivate.nu")),
+            "__DEACTIVATE_PATH__": str(Path(dest_folder, "deactivate.nu")),
         }
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -24,5 +24,5 @@ class NushellActivator(ViaTemplateActivator):
             "__VIRTUAL_NAME__": creator.env_name,
             "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
             "__PATH_SEP__": ensure_text(os.pathsep),
-            "__DEACTIVATE_PATH__": ensure_text(str(Path(dest_folder, "deactivate.nu"))),
+            "__DEACTIVATE_PATH__": ensure_text(os.path.join(dest_folder, "deactivate.nu")),
         }
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -17,6 +17,7 @@ class NushellActivator(ViaTemplateActivator):
         # Due to nushell scoping, it isn't easy to create a function that will
         # deactivate the environment. For that reason a __DEACTIVATE_PATH__
         # replacement pointing to the deactivate.nu file is created
+        #
 
         return {
             "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -24,5 +24,5 @@ class NushellActivator(ViaTemplateActivator):
             "__VIRTUAL_NAME__": creator.env_name,
             "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
             "__PATH_SEP__": ensure_text(os.pathsep),
-            "__DEACTIVATE_PATH__": str(Path(dest_folder, "deactivate.nu")),
+            "__DEACTIVATE_PATH__": ensure_text(str(Path(dest_folder, "deactivate.nu"))),
         }
--- a/src/virtualenv/activation/nushell/__init__.py
+++ b/src/virtualenv/activation/nushell/__init__.py
@@ -1,0 +1,13 @@
+from __future__ import absolute_import, unicode_literals
+
+from virtualenv.util.path import Path
+
+from ..via_template import ViaTemplateActivator
+
+
+class NushellActivator(ViaTemplateActivator):
+    def templates(self):
+        yield Path("activate.nu")
+
+    def as_name(self, template):
+        return template.stem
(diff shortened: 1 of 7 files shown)


gaborbernat (changes requested):
For this to get accepted you'll need to setup the test suite for it, as per other shells do, see https://github.com/pypa/virtualenv/blob/main/tests/unit/activation/test_batch.py

elferherrera:
thanks @gaborbernat for pointing me to that script. I see that I would need to set up in the CI an instance of nushell as well to test it. Is that correct?

gaborbernat:
I'd also like to mention that condition of accepting this PR is:

- nushell will be stable and not break the CI every month,
- does not add the significant overhead of the CI.

If either of those is true now or in the future we'll remove it from the core, and encourage you to instead provide it as a plugin.

src/virtualenv/activation/nushell/deactivate.nu
```diff
@@ -0,0 +1,7 @@
+# Setting the old path
+let-env $path-name = $nu.env.VENV_OLD_PATH
+
+# Unleting the environment variables that where created when activating the env
```
kubouch, line 4: Just a minor typo here on "where"

elferherrera:
@gaborbernat I was wondering if you could help me understand something. When asserting this [line](https://github.com/pypa/virtualenv/blob/main/tests/unit/activation/conftest.py#L144) I get this value 
`c:\users\123123\documents\repos\virtualenv\venv\scripts\python.exe` 
but it expects this
`c:\users\123123\appdata\local\temp\pytest-of-123123\pytest-100\activation-tester-env0\e-$ èрт🚒♞中片-j\scripts\python.exe`.

I've looked everywhere but dont understand where that is specified in the unit tests. In the logs I see that the files are copied to the correct location. Can you help me out?

gaborbernat:
Those match up with these https://github.com/pypa/virtualenv/blob/main/tests/unit/activation/conftest.py#L123-L135. So what you see there is that post-activation we still get the original path and not the virtual environment one.

elferherrera:
@gaborbernat thanks. That was useful. However, I'm having trouble understanding why the output from this line 
```
source 'C:\Users\123123\AppData\Local\Temp\pytest-of-123123\pytest-131\activation-tester-env1\e-$ Ã¨Ñ€Ñ‚ðŸš’â™žä¸­ç‰‡-j\Scripts\activate.nu'
```

has to be 
```
c:\users\123123\appdata\local\temp\pytest-of-123123\pytest-131\activation-tester-env1\e-$ èрт🚒♞中片-j\scripts\python.exe
```

gaborbernat:
To support paths with Unicode characters. What you see in the line above is a sign of garbled coding by the shell (meaning your activator might work for ASCII paths), but will not for Unicode ones (notably when users have Unicode char in their name, and they work in their home folder).

elferherrera:
@gaborbernat sorry, I copied that string from the terminal and it got jumbled. That is not my problem now. It seems that because Im testing my changes in a virtual environment the test is not overriding the VIRTUAL_ENV from the test, that's why one file is pointing to the venv in my shell and the other is pointing to the venv from the test.

Is it possible to monkeypatch the variable for the test?

gaborbernat:
I don't think that's the case, because then why would this not happen for
the other shells 😬

elferherrera:
you are right. I'm running the powershell test and I dont get that error. This is really weird

elferherrera:
Ok. I think I found the issue. The problem is that when nushell sources that file, it changes the env variables (PATH) only for its scope and because the test is looking outside nushell's scope, then it finds the previous env variable.  I need to figure out how to maintain nushell's scope during the test

elferherrera:
@gaborbernat again, more questions. I have been testing the script that is run in the test, which is
```
python.exe -c 'import sys; print(sys.executable)'
python.exe -c 'import os; import sys; v = os.environ.get("VIRTUAL_ENV"); print(v)'
source 'C:\Users\123123\Documents\repos\virtualenv\something\Scripts\activate.nu'
python.exe -c 'import sys; print(sys.executable)'
python.exe -c 'import os; import sys; v = os.environ.get("VIRTUAL_ENV"); print(v)'
pydoc -w os
deactivate
python.exe -c 'import sys; print(sys.executable)'
python.exe -c 'import os; import sys; v = os.environ.get("VIRTUAL_ENV"); print(v)'
```
and it runs perfectly in nushell and even inside a python shell using Popen
the result I get is
```
C:\Users\123123\Documents\repos\virtualenv\venv\Scripts\python.exe      <---- old venv
C:\Users\123123\Documents\repos\virtualenv\venv                         <---- old venv
C:\Users\123123\Documents\repos\virtualenv\something\Scripts\python.exe <---- new venv
something                                                               <---- new venv
wrote os.html                                                           <---- result from pydoc
C:\Users\123123\Documents\repos\virtualenv\venv\Scripts\python.exe      <---- old venv after deactivate
C:\Users\123123\Documents\repos\virtualenv\venv                         <---- old venv
```
which is what the test expect

however, when running the test, in this [line](https://github.com/elferherrera/virtualenv/blob/nushell/tests/unit/activation/conftest.py#L87) the result we get is this:

```
c:\users\123123\documents\repos\virtualenv\venv\scripts\python.exe      <---- This should be none
None                                                                    <---- correct
c:\users\123123\documents\repos\virtualenv\venv\scripts\python.exe      <---- this should be the new env
e-$ èрт🚒♞中片-j                                                       <---- this is correct. Name of env
wrote os.html                                                           <---- result from pydoc
c:\users\123123\documents\repos\virtualenv\venv\scripts\python.exe      <---- old env
None                                                                    <---- correct
```
which obviously shows that the new env is not loaded.

So, this leads me to the next question, do you remember if there is an special configuration that I need to set for the `ActivationTester`? Do I need to patch some other env variables?

gaborbernat:
If they are environment variables at play that must be specific to nu, and those I'm not aware of. 😬 Perhaps - someone who knows in detail how nu operates can help 🤔

elferherrera:
@gaborbernat I see your requested change but the nushell test file is there. Is that because it is an old comment or is there something else I need to add?

gaborbernat:
Can you do a push to trigger the CI without modifying the workflow file?

gaborbernat (commented):
I'd like to see the CI run outcome 🤔

gaborbernat:
I didn't say restore the workflow, I've said add a commit that doesn't contain workflow change. I have to approve every workflow change, but approval only works for follow-up commits.

gaborbernat:
Ok make now an empty commit to trigger the build :+1:

gaborbernat:
Can you push another, not sure why the CI is not running. Do you have a Syntax error in it?

gaborbernat:
Looks like it, see https://github.com/pypa/virtualenv/actions/runs/1177284371, can you please fix it?

gaborbernat (changes requested):
The CI needs to be functional, including Windows.

https://github.com/pypa/virtualenv/actions/runs/1177524887/workflow

elferherrera:
@gaborbernat Hi. for sure Im going to add the installation for windows. I just want to test that it works for linux and mac

elferherrera:
@gaborbernat I simplified the untar operation

gaborbernat:
Seems things fail on macOS now :+1:

elferherrera:
@gaborbernat I was missing a case when there is no prompt. I have added that case as well

elferherrera:
> Seems things fail on macOS now 👍
Unfortunately it is still downloading 0.35 version from brew instead of 0.36. I dont know how to control that

elferherrera:
@gaborbernat It seems that it is passing on linux, that is nice to see.
On mac it is failing because it is getting nu 0.35. I dont know what to do to force it to get 0.36.

The test failure that I dont understand is the `2.7 windows` because that test gets skipped on windows machines now. Do you know what could be?

elferherrera:
@gaborbernat I think I found why it wasnt working for python 2.7.  I added the `ensure_text` function to one variable

elferherrera:
@gaborbernat I removed instructions not needed for nushell test when deactivating

elferherrera:
@gaborbernat Im testing installation with scoop

elferherrera:
@gaborbernat Im trying to install nushell with choco. I hope that works

elferherrera:
@gaborbernat changed the parameter for line ending

elferherrera:
@gaborbernat can you approve again? I changed the path join for python 2.7

gaborbernat:
I can approve, but you can test this on your own under https://github.com/elferherrera/virtualenv/actions 😊 and then I don't need to keep approving 😆

elferherrera:
@gaborbernat really? sorry, didnt know I can do that

elferherrera:
@gaborbernat by any chance do you know why on mac the expected folder for the binary is like this (right hand value in the assert)?
```
assert ('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/activation-tester-env5/e-$ '\n 'èрт🚒♞中片-j/bin/python.exe') == '/Users/runner/hostedtoolcache/Python/3.7.11/x64/bin/python3.7'
```

I dont understand where it is coming from

gaborbernat:
That's the system python path.

elferherrera:
but how come the expected value is the system path and not the virtualenv path?

elferherrera:
but, how come the expected is the system path instead of the env path?

this [line] (https://github.com/pypa/virtualenv/blob/main/tests/unit/activation/conftest.py#L144) is telling me that I need this assert to be true
```
assert ('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/activation-tester-env5/e-$ '\n 'èрт🚒♞中片-j/bin/python.exe') == '/Users/runner/hostedtoolcache/Python/3.7.11/x64/bin/python3.7'
```
which doesnt make sense

gaborbernat:
It could be a bug in either virtualenv or the Github Action python interpreter, can't tell without digging deeper 🤔

elferherrera:
Is the session maker different for Mac?

gaborbernat:
> Is the session maker different for Mac?

It's different for every platform, and even within the platform different per python flavour.

elferherrera:
where do you recommend me to start looking for this bug?

gaborbernat:
No idea 😬 would have to look to tell that 😂

elferherrera:
We are so close. I really want to have this in nushell to start using it as my main shell

gaborbernat:
> We are so close. I really want to have this in nushell to start using it as my main shell

I'm on a holiday now so unavailable for looking into this myself. But if you can figure out the CI, we can push it ahead.

elferherrera:
>I'm on a holiday now so unavailable for looking into this myself. But if you can figure out the CI, we can push it ahead.

the problem is that the mac tests always fail because of that assertion. I dont know what to do about it

gaborbernat:
It passes on main, see https://github.com/pypa/virtualenv/actions/runs/1200507280, so the reason must be with the code you're adding here 😊

lily-mara:
I believe I have resolved the macOS test failures in https://github.com/elferherrera/virtualenv/pull/1

elferherrera:
@gaborbernat thanks to @lily-mara we are passing all virtualenv tests. Do you mind approving the workflow?

src/virtualenv/activation/nushell/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        # deactivate the environment. For that reason a __DEACTIVATE_PATH__
+        # replacement pointing to the deactivate.nu file is created
+
+        return {
+            "__VIRTUAL_PROMPT__": "" if self.flag_prompt is None else self.flag_prompt,
+            "__VIRTUAL_ENV__": ensure_text(str(creator.dest)),
+            "__VIRTUAL_NAME__": creator.env_name,
+            "__BIN_NAME__": ensure_text(str(creator.bin_dir.relative_to(creator.dest))),
+            "__PATH_SEP__": ensure_text(os.pathsep),
+            "__DEACTIVATE_PATH__": ensure_text(str(Path(dest_folder) / Path("deactivate.nu"))),
```
gaborbernat, line 27: I don't think you need to wrap the `deactivate.nu` in path :+1:

tests/unit/activation/test_nushell.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import, unicode_literals
+
+from virtualenv.activation import NushellActivator
+from virtualenv.info import IS_WIN
+
+
+def test_nushell(activation_tester_class, activation_tester):
+    class Nushell(activation_tester_class):
+        def __init__(self, session):
+            cmd = "c:\\program files\\nu\\bin\\nu.exe" if IS_WIN else "nu"
```
gaborbernat, line 10: This might work in the CI but will not work locally as well :+1: Better to use `shutil.which` here and only fallback to this hard coded path if that returns None and we're on windows.

gaborbernat (changes requested):

gaborbernat (approved):

gaborbernat:
@elferherrera seems Python 2 does not like it :worried:

elferherrera:
@gaborbernat let me have a look. I was checking that I was getting an error with the lints

gaborbernat:
Python 2 doesn't support shutil.which so you'd need to add a small backwards compatible method :thinking:

elferherrera:
@gaborbernat I think this time everything passes

gaborbernat:
Released via https://virtualenv.pypa.io/en/latest/changelog.html#v20-8-0-2021-09-16

Issue #2024, opened 2020-11-26
CPython2 suppport on Apple M1


gaborbernat:
👍 I don't have access to such hardware, so PRs are welcome.

gdubicki:
Hi! I also did run into this issue and spent quite a lot of time figuring out what was going on.

How about we start with adding code that will show a friendly error message when trying to create a virtualenv for Python 2 on a Apple Silicon Mac, explaining that it doesn't work as of now.

Would such PR be accepted?

gaborbernat:
Sure.

PR #2179, opened 2021-09-11
Show an error instead of creating non-working env

gdubicki:
for Python 2 on Apple Silicon (M1) Macs.

Related to #2024.

### DIFF
--- a/src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
+++ b/src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
@@ -10,6 +10,7 @@ from textwrap import dedent
 from six import add_metaclass
 
 from virtualenv.create.via_global_ref.builtin.ref import ExePathRefToDest, PathRefToDest, RefMust
+from virtualenv.info import IS_MAC_ARM64
 from virtualenv.util.path import Path
 from virtualenv.util.six import ensure_text
 
@@ -101,6 +102,15 @@ class CPython2macOsFramework(CPythonmacOsFramework, CPython2PosixBase):
             ),
         )
         return result
+
+    @classmethod
+    def can_create(cls, interpreter):
+        if IS_MAC_ARM64:
+            # https://github.com/pypa/virtualenv/issues/2024
+            logging.fatal("Creating virtualenv for Python 2 on Apple Silicon Macs is not supported as of now.")
+            return False
+        else:
+            return super(CPythonmacOsFramework, cls).can_create(interpreter)
 
 
 class CPython3macOsFramework(CPythonmacOsFramework, CPython3, CPythonPosix):
--- a/docs/changelog/2179.misc.rst
+++ b/docs/changelog/2179.misc.rst
@@ -1,0 +1,1 @@
+Show an error instead of creating non-working env for Python 2 on Apple Silicon (M1) Macs.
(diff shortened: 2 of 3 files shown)


gdubicki:
I cannot add a real test that would run on GitHub Actions it doesn't provide Macs on Apple Silicon (yet). But it successfully fails ;P on my machine:

```
(venv) gdubicki@mac ~/git/virtualenv 2024_friendly_error $ .tox/dev/bin/virtualenv -p python2 venv4
Creating virtualenv for Python 2 on Apple Silicon Macs is not supported as of now.
RuntimeError: No virtualenv implementation for PathPythonInfo(spec=CPython2.7.16.final.0-64, system=/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python, exe=/usr/bin/python2, platform=darwin, version='2.7.16 (default, Jun 18 2021, 03:23:52) \n[GCC Apple LLVM 12.0.5 (clang-1205.0.19.59.6) [+internal-os, ptrauth-isa=deploy', encoding_fs_io=utf-8-None)
```

gaborbernat (approved):

src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
```diff
@@ -102,6 +103,15 @@ def reload_code(self):
         )
         return result
 
+    @classmethod
+    def can_create(cls, interpreter):
+        if IS_MAC_ARM64:
+            # https://github.com/pypa/virtualenv/issues/2024
+            logging.fatal("Creating virtualenv for Python 2 on Apple Silicon Macs is not supported as of now.")
```
gaborbernat, line 110: You'll need to remove this, as this would still print if someone adds support via a plugin

gaborbernat (changes requested):

src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
gdubicki: Thanks for a review @gaborbernat !

Unfortunately I am not familiar with this project's structure yet. Can you recommend a better place to put this check then?

Alternatively we could change this to:
```
logging.warning("The built-in creator does not handle creating virtualenvs for Python 2 on Apple Silicon Macs as of now.")
```
- that should not be misleading for a person that adds their own plugin to support this case.

gaborbernat: The location is good. There isn't a place to put the warning message though, so that needs to be removed just.

gdubicki: I am afraid that without this message this PR doesn't make sense as the whole point was to show a friendly error that explains what happened.

The generic "No virtualenv implementation" error would make look for issues in my machine Python setup, not assume that this is actually a missing virtualenv feature...

Maybe it would be easier for you to implement the missing plugin instead? I would love to do it myself but I just don't know enough about Mach-O hacking to it to make sense... I am willing to do all the testing you need and as it seems that we are in similar timezones (BST / CEST) we could sync easily.

gaborbernat: > I would love to do it myself but I just don't know enough about Mach-O hacking to it to make sense

Me neither... And I don't have access to an Apple M1 machine to write an implementation for it or time to do that. So guess we wait until someone comes along who does.

gdubicki:
On second thought it is better to fail with a proper error message than creating a non-working env, so I created #2189.

PR #2199, opened 2021-09-25
Restore unintentionally-modified pre-``2186`` cleanup behaviors

### LINKED ISSUE #2198: Support overriding corrupt virtual environments
The fix for #2185 in #2186 introduced changes to the behavior of Virtualenv that was not the intention of the fix. Such changes may affect existing environments in unexpected ways. Specifically:

1. The uninstall now would uninstall from `platlib`. However Virtualenv never installed (and does not now install) the embedded into the `platlib` and if a distro was installed there, the pre-2186 behavior was to leave `platlib` completely undisturbed. The #2186 modified this behavior without a warning, reason, an ability to test such change (none of the embeds go to `platlib`) while violating [the principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

2. Prior to #2186 during installation if the target directory was present (for whatever reason), the logic would remove such target. Now that target must not exist or the installation would fail. A virtual env may contain extraneous directories/symlinks that are not associated with a distro (that would be handled by an uninstall), if the distro was installed in [development mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html), which deals in egg-links, symlinks and pth-files.

arcivanov:
1. Do not disturb platdir
2. Do delete a target directory on install if it is present and
wasn't a part of the distro that could've been uninstalled

fixes #2198

### DIFF
--- a/src/virtualenv/seed/embed/via_app_data/pip_install/base.py
+++ b/src/virtualenv/seed/embed/via_app_data/pip_install/base.py
@@ -36,6 +36,8 @@ class PipInstall(object):
         # sync image
         for filename in self._image_dir.iterdir():
             into = self._creator.purelib / filename.name
+            # The `into` path may still be dirty due to partial/broken installs - clean it up
+            self._delete_path(into)
             self._sync(filename, into)
         # generate console executables
         consoles = set()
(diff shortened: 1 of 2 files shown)


src/virtualenv/seed/embed/via_app_data/pip_install/base.py
```diff
@@ -149,12 +151,21 @@ def _create_console_entry_point(self, name, value, to_folder, version_info):
 
     def _uninstall_previous_version(self):
         dist_name = self._dist_info.stem.split("-")[0]
-        in_folders = chain.from_iterable([i.iterdir() for i in {self._creator.purelib, self._creator.platlib}])
+        # Only uninstall purelib as it's the only one that VirtualEnv installs into
+        in_folders = chain.from_iterable([i.iterdir() for i in {self._creator.purelib}])
```
gaborbernat, line 155: Please remove this.

gaborbernat (changes requested):
Also, improve the wording of the changelog/PR to describe the change, and not speak about unintentional changes. They were very much intentional.

PR #2204, opened 2021-10-02
Skip directories when symlinking libraries for PyPy3

### LINKED ISSUE #2182: Incorrect (broken) virtualenv layout with pypy3.8's new layout
**Issue**

PyPy3.8 (currently 7.3.6rc1) supports a new install layout that resembles CPython more. That is, `sys.prefix` no longer needs being isolated, and site-packages are found in `$prefix/lib/pypy3.8/site-packages`. However, virtualenv tries to symlink everything from `/usr/lib` including the `pypy3.8` directory. As a result, the user can't write to the site-packages directory in the venv.

I haven't tried running it as root but I can imagine it doing major mess if it virtualenv doesn't take any precautions from writing into system directories.

**Environment**

Provide at least:
- OS: Gentoo Linux
- ``pip list`` of the host python where ``virtualenv`` is installed: [pip-list.txt](https://github.com/pypa/virtualenv/files/7167321/pip-list.txt)

Full output: [output.txt](https://github.com/pypa/virtualenv/files/7167331/output.txt)

tail:
```
1048 create virtualenv import hook file /tmp/z/lib/pypy3.8/site-packages/_virtualenv.pth [DEBUG api:95]
  File "/usr/lib/pypy3.8/runpy.py", line 198, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/pypy3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/lib/pypy3.8/site-packages/virtualenv/__main__.py", line 80, in <module>
    run_with_catch()  # pragma: no cov
  File "/usr/lib/pypy3.8/site-packages/virtualenv/__main__.py", line 65, in run_with_catch
    run(args, options, env)
  File "/usr/lib/pypy3.8/site-packages/virtualenv/__main__.py", line 18, in run
    session = cli_run(args, options, env)
  File "/usr/lib/pypy3.8/site-packages/virtualenv/run/__init__.py", line 32, in cli_run
    of_session.run()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/run/session.py", line 46, in run
    self._create()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/run/session.py", line 53, in _create
    self.creator.run()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/create/creator.py", line 171, in run
    self.create()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/create/via_global_ref/builtin/via_global_self_do.py", line 101, in create
    super(ViaGlobalRefVirtualenvBuiltin, self).create()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/create/via_global_ref/api.py", line 89, in create
    self.install_patch()
  File "/usr/lib/pypy3.8/site-packages/virtualenv/create/via_global_ref/api.py", line 96, in install_patch
    pth.write_text("import _virtualenv")
  File "/usr/lib/pypy3.8/pathlib.py", line 1255, in write_text
    with self.open(mode='w', encoding=encoding, errors=errors) as f:
  File "/usr/lib/pypy3.8/pathlib.py", line 1223, in open
    opener=self._opener)
  File "/usr/lib/pypy3.8/pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
PermissionError: [Errno 13] Permission denied: PosixPath('/tmp/z/lib/pypy3.8/site-packages/_virtualenv.pth')
```

mgorny:
The PyPy3 logic creates symlinks for all files from the library
directory existing alongside the PyPy executable.  This is meant
to ensure that the bundled libraries to which PyPy is linked can also
be found from inside the virtualenv.  However, this logic also symlinks
all directories which is unnecessary and causes library directory
collisions with the new install layout.  Change to logic to symlink
non-directories only.

A similar fix has been applied to the internal venv module in PyPy3.8:
https://foss.heptapod.net/pypy/pypy/-/commit/713b2af9abd2b9453e12c60143e17431a1aefb33

Fixes #2182

- [ ] ensured there are test(s) validating the fix (n/a?)
- [ ] updated/extended the documentation (n/a)

### DIFF
--- a/src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
+++ b/src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
@@ -44,6 +44,8 @@ class PyPy3Posix(PyPy3, PosixSupports):
         host_lib = Path(interpreter.system_prefix) / "lib"
         if host_lib.exists() and host_lib.is_dir():
             for path in host_lib.iterdir():
+                if path.is_dir():
+                    continue
                 yield PathRefToDest(path, dest=cls.to_lib)
 
 
--- a/docs/changelog/2182.bugfix.txt
+++ b/docs/changelog/2182.bugfix.txt
@@ -1,0 +1,2 @@
+Fixed path collision that could lead to a PermissionError or writing to system
+directories when using PyPy3.8 - by :user:`mgorny`.


src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
```diff
@@ -44,6 +44,8 @@ def sources(cls, interpreter):
         host_lib = Path(interpreter.system_prefix) / "lib"
         if host_lib.exists() and host_lib.is_dir():
             for path in host_lib.iterdir():
+                if path.is_dir():
+                    continue
```
mattip, line 48: Once upon  a time we needed to copy any shared object files in `base/lib` but these days we use RPATH in the `libpypy3-c.so` to find the `lib` dir, and that main shared object is now symlinked.

So in short, iterating over `base/lib` and copying the files is no longer needed, so lines 45-48 can be deleted. That means this whole method can be deleted.

mattip:
I validated this (both the minimal change first suggested and my more extensive suggested change) against the upcoming PyPy3.8 rc1 and against PyPy 3.7. I don't think the failures are connected to this change.

src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
mattip: ... as long as `os.symlink` is available where the "portable PyPy" downloads (the ones from https://www.pypy.org/download.html) are used. Linux and brew build PyPy against system libraries, so they do not need the shared objects in the `base/lib` directory .

gaborbernat: @mattip for 3.6 we still need this, not? What about Windows?

mattip: No, 3.6 also uses symlinks and RPATH. Windows never needed this.

@mgorny could you change line 30, 31 to use `pypy` instead of `python` ?

mattip:
I tried to emit a PR to update this from [mattip/virtualenv](https://github.com/mattip/virtualenv/commit/7f15291649fcedaee8854aa0a8b5f492e4c49c3f) but for some reason could not make a PR to mgorny/virtualenv. While this version is fine, I think removing the whole `source` class method is cleaner.

mgorny:
> I tried to emit a PR to update this from [mattip/virtualenv](https://github.com/mattip/virtualenv/commit/7f15291649fcedaee8854aa0a8b5f492e4c49c3f) but for some reason could not make a PR to mgorny/virtualenv. While this version is fine, I think removing the whole `source` class method is cleaner.

I'll just close my PR and please submit yours instead ;-).

PR #2206, opened 2021-10-04
Pypy3 sym

mattip:

This builds on and replaces #2204. In an earlier version of portable PyPy (one built on manylinux 2010 and bundling dependent shared objects), it was necessary to copy the bundled shared objects from `prefix/lib`. This is no longer needed since now the `libpypy3-c.so` is symlinked not copied.

On non-portable PyPy installations, like those in distros or on conda-forge, this code was never used since there the dependent shared objects are provided outside the python tree.

Also fixed the stdlib which had python not pypy as the stem of the directory name.

I tested this locally to make sure it works as advertised.


src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
```diff
@@ -37,15 +35,6 @@ def _shared_libs(cls):
...
         return self.dest / "lib" / src.name
 
-    @classmethod
-    def sources(cls, interpreter):
-        for src in super(PyPy3Posix, cls).sources(interpreter):
-            yield src
-        host_lib = Path(interpreter.system_prefix) / "lib"
-        if host_lib.exists() and host_lib.is_dir():
-            for path in host_lib.iterdir():
-                yield PathRefToDest(path, dest=cls.to_lib)
```
gaborbernat, line 47: I'm fairly sure there was a pypy3 variant that worked like this (over a year ago when I wrote this). Doesn't this drop support for those?

mattip: Ok, let's err on the side of caution and stick with @mgorny 's fix

gaborbernat: I'd actually prefer if we'd not do this :blush: but make it version-specific; if you can dig out from pypy when this change happened?

mattip: Before PyPy's reorganization of the directory layout for python3.8 there were no directories in `prefix/lib`. 

It seems this code was added in #1503 in a blanket change to "support c-extensions inside virtual environments". The issue pointed to is #1448, but that issue only mentions include files, not lib files. It seems to relate to issue #1023, but that one is windows only and deals with the libs directory, not the lib one.

gaborbernat: Hmm, in that case I'm happy to go with your original proposal, sorry for the back and forth.

mattip: cool, updating

gaborbernat (changes requested):
Seems the pypy3 test suite fails :thinking:

mattip:
Hmm. When I run `./pypy3.6-HEAD/bin/pypy -m pytest tests/unit/create` I get an error `error: unrecognized arguments: --no-success-flaky-report`. This is with pytest 6.2.5.

We should add a pypy3.7 run.

gaborbernat:
> We should add a pypy3.7 run.

Yes, we should have both 3.6 and 3.7 there 😊

gaborbernat (approved):
Assuming the CI passes 😊

mattip:
Cool, thanks

.github/workflows/check.yml
```diff
@@ -30,9 +30,11 @@ jobs:
           - 3.7
           - 3.6
           - 3.5
-          - pypy3
+          - pypy-3.6
+          - pypy-3.7
+          - pypy-3.8
```
mattip, line 35:
```suggestion
          - pypy-3.8-v7.3.6rc1
```

gaborbernat (approved):

mattip:
I think the linux pypy3.8 tests will be fixed with rc2 which should be ready this week. I am not sure what is the problem with windows, is it missing python.exe ?

gaborbernat:
> I think the linux pypy3.8 tests will be fixed with rc2 which should be ready this week. I am not sure what is the problem with windows, is it missing python.exe ?

Could be this https://github.com/tox-dev/py-filelock/issues/101? Does pypy support https://docs.python.org/3/library/msvcrt.html#msvcrt.locking? Pypy on Windows is sometimes interesting 🤔 but don't have time to debug it...

mattip:
> Could be this tox-dev/py-filelock#101? 

Maybe

> Does pypy support https://docs.python.org/3/library/msvcrt.html#msvcrt.locking?

Nominally, but there may be bugs since it is implemented via CFFI, so maybe there are threading problems ...

gaborbernat:
Guess this is blocked until RC2 then? Ping me once that happens.

mattip:
ok, but I would like to leave windows out for now, otherwise we won't have a virtualenv story for the 3.8 release.

gaborbernat:
Generally we don't tend to do piecewise support 🤔 so if possible would prefer to fix Windows... 🤔

mattip:
I need some help to debug the create_no_seed failures on windows. Although the source location has `['pypy.exe', `pypy3.exe`, `pypy3w.exe`, pypyw.exe`, `python.exe`, `pythonw.exe`]`, the last two are not copied to the destination, and the [check for `python.exe`] fails. https://github.com/pypa/virtualenv/blob/9cd075ef0b8ead2d8c2a230ab407e59406e4f819/tests/unit/create/test_creator.py#L216. 

But when I stop in a debugger where the exes are listed, `python` appears in the list. https://github.com/pypa/virtualenv/blob/9cd075ef0b8ead2d8c2a230ab407e59406e4f819/src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py#L21-L22.

Why isn't it being copied?

gaborbernat:
@mattip  The tests failing are not the ones testing that our creator is correct, but rather than the built-in venv behaves equivalently to ours. And in this case seems pypy differs, as the python.exe is not created when using pypy builtin venv 🤔 the weird part here though is that neither 3.6 should pass... but for some reason does (only in the CI locally for me fails both). So guess the question is that should the venv module on Windows create python.exe? And if not we should patch the test to express such 😊

mattip:
> the weird part here though is that neither 3.6 should pass

The 3.6 windows downloads were last updated for release 7.3.3, and have only `pypy3.exe` and `pypy3w.exe`. We started adding all the other `exe` variants three months ago, so only the 7.3.6rc1 downloads have `pypy.exe`, `python.exe` and `python3.exe`

mattip:
Ahah. Found the problem with `venv`. In `symlink_or_copy`, it intercepts `python.exe` and replaces it with `venvlauncher.exe`, which does not exists on PyPy.

gaborbernat:
> Ahah. Found the problem with `venv`. In `symlink_or_copy`, it intercepts `python.exe` and replaces it with `venvlauncher.exe`, which does not exists on PyPy.

Is that within pypy venv module?

mattip:
Yes.

mattip:
cool. Fixing venv makes the failures go away locally. I will put that in for rc2 and then we can try it here in CI.

gaborbernat:
👍

mattip:
It seems I should pay better attention to stdlib venv test failures. The problem is there, not with virtualenv

gaborbernat:
virtualenv seems like a good place to test the venv module 😃

mattip:
PyPy3.8 on windows hangs waiting for a lock in a `__del__`. Would it make sense to refactor this code to avoid doing anything in the destructor as long as `_del_lock(lock)` is called with `None`, and then in the `__exit__` set `self.lock` to `None` once it is released?
https://github.com/pypa/virtualenv/blob/a40640c40181625e9fc7addc857da72ae3bb82ea/src/virtualenv/util/lock.py#L92-L95

refactored to
```
    def _del_lock(lock):
        if lock is not None:
            with _store_lock:
                with lock.thread_safe:

```
<details>

```
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\app_data\via_disk_folder.py", line 99, in py_info_clear
    with py_info_folder.lock_for_key(filename.stem):
  File "C:\hostedtoolcache\windows\PyPy\3.8.12\x64\Lib\contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\util\lock.py", line 132, in lock_for_key
    lock = self._create_lock(name)
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\util\lock.py", line 88, in _create_lock
    _lock_store[lock_file] = _CountedFileLock(lock_file)
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\util\lock.py", line 18, in __init__
    parent = os.path.dirname(lock_file)
  File "C:\hostedtoolcache\windows\PyPy\3.8.12\x64\Lib\ntpath.py", line 223, in dirname
    return split(p)[0]
  File "C:\hostedtoolcache\windows\PyPy\3.8.12\x64\Lib\ntpath.py", line 186, in split
    seps = _get_bothseps(p)
  File "C:\hostedtoolcache\windows\PyPy\3.8.12\x64\Lib\ntpath.py", line 38, in _get_bothseps
    return '\\/'
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\coverage\pytracer.py", line 101, in _trace
    if THIS_FILE in frame.f_code.co_filename:
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\util\lock.py", line 100, in __del__
    self._del_lock(self._lock)
  File "D:\a\virtualenv\virtualenv\.tox\pypy3\lib\site-packages\virtualenv\util\lock.py", line 93, in _del_lock
    with _store_lock:

```

</details>

mattip:
Additionally, macos PyPy 3.8 failed `test_app_data_pinning` with
```
ERROR: No matching distribution found for pip==19.3.1
```

A transient network error?

mattip:
Rebased off main to get #2213. CI passed except codecov.

Issue #2236, opened 2021-11-18
DeprecationWarning: use poll_interval instead of poll_intervall


fungi:
**Issue**

The filelock library deprecated `poll_intervall` in favor of `poll_interval` (one `l` instead of two) in its 3.4.0 release earlier this week. Now all tox invocations with latest filelock raise a DeprecationWarning, and terminate in a rather opaque manner if `PYTHONWARNINGS` is set to error (`--with-traceback` helps, but is usually not set when calling virtualenv from, e.g., tox).

**Environment**

Provide at least:
- OS: Debian sid with manually compiled Python 3.10.0
- ``pip list`` of the host python where ``virtualenv`` is installed:

  ```console
  backports.entry-points-selectable 1.1.1
  distlib                           0.3.3
  filelock                          3.4.0
  packaging                         21.2
  pip                               21.3.1
  platformdirs                      2.4.0
  pluggy                            1.0.0
  py                                1.11.0
  pyparsing                         2.4.7
  setuptools                        59.1.1
  six                               1.16.0
  toml                              0.10.2
  tox                               3.24.4
  virtualenv                        20.10.0
  wheel                             0.37.0
  ```

**Output of the virtual environment creation**

Make sure to run the creation with `-vvv --with-traceback`:

```console
$ PYTHONWARNINGS=error lib/tox/bin/virtualenv -vvv --with-traceback foo

498 setup logging to NOTSET [DEBUG report:39]
522 find interpreter for spec PythonSpec(path=/home/fungi/lib/tox/bin/python3) [INFO builtin:62]
523 discover exe for PythonInfo(spec=CPython3.10.0.final.0-64, exe=/home/fungi/lib/tox/bin/python3, platform=linux, version='3.10.0 (tags/v3.10.0:b494f5935c, Oct  4 2021, 22:28:32) [GCC 10.3.0]', encoding_fs_io=utf-8-utf-8) in /home/fungi/lib/cpython/3.10.0 [DEBUG py_info:399]
525 filesystem is case-sensitive [DEBUG info:29]
Traceback (most recent call last):
  File "/home/fungi/lib/tox/bin/virtualenv", line 8, in <module>
    sys.exit(run_with_catch())
  File "/home/fungi/lib/tox/lib/python3.10/site-packages/virtualenv/__main__.py", line 65, in run_with_catch
    run(args, options, env)
  File "/home/fungi/lib/tox/lib/python3.10/site-packages/virtualenv/__main__.py", line 18, in run
[...]

PR #2237, opened 2021-11-20
Avoid a new deprecation warning from filelock dependency

ofek:
PR: https://github.com/tox-dev/py-filelock/pull/119
Version: https://github.com/tox-dev/py-filelock/blob/main/docs/changelog.rst#v340-2021-11-16


setup.cfg
```diff
@@ -42,7 +42,7 @@ packages = find:
 install_requires =
     backports.entry-points-selectable>=1.0.4
     distlib>=0.3.1,<1
-    filelock>=3.2,<4
+    filelock>=3.4,<4
```
XuehaiPan, line 45:
```suggestion
    filelock>=3.2,<4
```

Leave it as it is. `filelock >= 3.4` requires Python >= 3.6:

https://github.com/tox-dev/py-filelock/blob/a6c8fabc4192fa7a4ae19b1875ee842ec5eb4f61/setup.cfg#L34

but `virtualenv` supports:

https://github.com/pypa/virtualenv/blob/623189a00aa96c805990fd12dc883c69e251f4b2/setup.cfg#L21-L29

src/virtualenv/util/lock.py
```diff
@@ -25,10 +25,10 @@ def __init__(self, lock_file):
         self.count = 0
         self.thread_safe = RLock()
 
-    def acquire(self, timeout=None, poll_intervall=0.05):
+    def acquire(self, timeout=None, poll_interval=0.05):
         with self.thread_safe:
             if self.count == 0:
-                super(_CountedFileLock, self).acquire(timeout=timeout, poll_intervall=poll_intervall)
+                super(_CountedFileLock, self).acquire(timeout=timeout, poll_interval=poll_interval)
```
XuehaiPan, line 31:
```suggestion
                super(_CountedFileLock, self).acquire(timeout, poll_interval)
```

Since we cannot use `filelock >= 3.4` (see comment above), use positional arguments.

setup.cfg
ofek: Thanks!

src/virtualenv/util/lock.py
ofek: Thanks!

fungi:
This should fix #2236, thanks!

ofek:
@gaborbernat only Codecov is failing, tests pass

gaborbernat:
I'm off to a holiday for two weeks, will check afterwards.

ofek:
CI is failing

ofek:
Should I revert that commit?

setup.cfg
gaborbernat: Indeed, but then to solve this we would want to have differing version requirements for 3.6+ and earlier; furthermore the code needs to dynamically generate the kwargs, otherwise on older pythons we'd be using kwarg that's not supported.

gaborbernat (changes requested):

ofek:
Does this look fine now?

gaborbernat:
No, see my comment on outdated entry, but still change request:

> Indeed, but then to solve this we would want to have differing version requirements for 3.6+ and earlier; furthermore the code needs to dynamically generate the kwargs, otherwise on older pythons we'd be using the new kwarg that's not supported.

ofek:
So the current positional argument strategy won't suffice?

gaborbernat:
> So the current positional argument strategy won't suffice?

Ah you're right.

gaborbernat (approved):

PR #2238, opened 2021-11-26
Omit backports.entry_points_selectable for Python >=3.10

Chocimier:
If I understood backports.entry_points_selectable documentation correctly, this is equivalent.
Change is proposed to update above 20.4.7 on Void Linux, which is on python 3.10 already, without fuss of adding the backport module to repository.

### DIFF
--- a/docs/changelog/2238.feature.rst
+++ b/docs/changelog/2238.feature.rst
@@ -1,0 +1,1 @@
+Omit backports.entry_points_selectable for Python >=3.10
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,11 +40,11 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    backports.entry-points-selectable>=1.0.4
     distlib>=0.3.1,<1
     filelock>=3.2,<4
     platformdirs>=2,<3
     six>=1.9.0,<2   # keep it >=1.9.0 as it may cause problems on LTS platforms
+    backports.entry-points-selectable>=1.0.4;python_version<"3.10"
     importlib-metadata>=0.12;python_version<"3.8"
     importlib-resources>=1.0;python_version<"3.7"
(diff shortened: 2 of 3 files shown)


hroncok:
Thank you!

gaborbernat (approved):

docs/changelog/2238.misc.rst
```diff
@@ -0,0 +1 @@
+Omit backports.entry_points_selectable for Python >=3.10
```
gaborbernat, line 1: Can you add this as a feature please and not misc; misc basically reserved for project maintenance files only, not that affects code.

gaborbernat (changes requested):

docs/changelog/2238.misc.rst
gaborbernat: Also, can we drop `backports.entry_points_selectable` entirely here?

hroncok: I suppose we can if we require `importlib-metadata>=3.6;python_version<"3.10"` and do:

```python
if sys.version_info >= (3, 10):
    from importlib.metadata import entry_points
else:
    from importlib_metadata import entry_points
```

-- not sure if that is much better.

gaborbernat: Alternatively can we use the old API on old pythons and the new API on new Pythons? Rather than try to use the same?

hroncok: Like this? https://github.com/pypa/virtualenv/pull/2241

Chocimier: Added as feature. The other PR would fit my needs as well.

gaborbernat: I prefer 2241, so will be closing this.

PR #2241, opened 2021-12-06
Use 'selectable' interface for entry points only when available

hroncok:
This reverts commit 17caadc974cb2a24f70374525596cd1160579594.

### DIFF
--- a/src/virtualenv/run/plugin/base.py
+++ b/src/virtualenv/run/plugin/base.py
@@ -1,8 +1,16 @@
 from __future__ import absolute_import, unicode_literals
 
+import sys
 from collections import OrderedDict
 
-from backports.entry_points_selectable import entry_points
+if sys.version_info >= (3, 8):
+    from importlib.metadata import entry_points
+
+    importlib_metadata_version = ()
+else:
+    from importlib_metadata import entry_points, version
+
+    importlib_metadata_version = tuple(int(i) for i in version("importlib_metadata").split(".")[:2])
 
 
 class PluginLoader(object):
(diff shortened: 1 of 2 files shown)


src/virtualenv/run/plugin/base.py
```diff
@@ -11,7 +15,11 @@ class PluginLoader(object):
 
     @classmethod
     def entry_points_for(cls, key):
-        return OrderedDict((e.name, e.load()) for e in cls.entry_points().select(group=key))
+        try:
```
gaborbernat, line 18: Instead of try/catch can we make it explicit based on the version? I want it so that an AttributeError on 3.10 is now silently swallowed.

hroncok: We would need to get the version of importlib_metadata.

gaborbernat: We should only need importlib_metadata where we don't have importlib.metadata

hroncok: Yes that is correct. But we would still need to check the version then.

hroncok: The if would need to be (pseudo-code):

```
if PYTHON_VERSION >= 3.10 or (PYTHON_VERSION < 3.8 and IMPORTLIB_METADATA_VERSION < 3.6):
```

I don't like that very much.

gaborbernat: I prefer it personally

hroncok: I wouldn't know how to get the importlib metadata version.

This works:

```
>>> importlib_metadata.version('importlib_metadata')
'4.8.2'
```

But I would need to parse it with packaging.version to get a comparable object.

gaborbernat: You can just split by . and cast the first two element to int, no?

hroncok: I can do that, but that seems rather fragile. What if the next version is 5b2? I am hesitant to implement our own version parsing logic just so we can do the check. I can certainly try.

hroncok: I am not happy about the code, but I've done it.

hroncok:
I've pushed the changelog fragment to this branch, but github is not showing it here.

hroncok:
> I've pushed the changelog fragment to this branch, but github is not showing it here.

Re-opened as https://github.com/pypa/virtualenv/pull/2246

PR #2253, opened 2021-12-27
chore: test with python 3.10 instead of 3.10.0-rc.1

mayeut:
Test with python 3.10 instead of 3.10.0-rc.1
Quote python versions to prevent them being interpreted as floats (where 3.10 == 3.1)

Should I add a news fragment for such a trivial change ?


gaborbernat:
Thanks, this needs more love so I've chosen to superseed it with #2254.

PR #2273, opened 2021-12-31
feature: store version update source in embed JSON file

mayeut:
Store version update source in embed JSON file.
For now, the source can take the value `periodic` or `manual` (not enforced to allow future sources).
Only `periodic` & `manual` sources are allowed to provide an update path for a `bundle` version.
Running a manual update marks all versions source as `manual`.
Running a periodic update marks all versions source as `periodic` if the source value is not `manual`.

The following issues are fixed:
- fixes #2265
- fixes #2266
- fixes #2267

This will also allow to implement #2268 with a new source (e.g. `download`).

 
### Thanks for contributing, make sure you address all the checklists (for details on how see
[development documentation](https://virtualenv.pypa.io/en/latest/development.html#development))!

- [x] ran the linter to address style issues (``tox -e fix_lint``)
- [x] wrote descriptive pull request text
- [x] ensured there are test(s) validating the fix
- [x] added news fragment in ``docs/changelog`` folder
- [ ] updated/extended the documentation

PR #2276, opened 2022-01-01
feature: cache downloaded wheel information

### LINKED ISSUE #2268: Do not attempt to download previously downloaded wheels
**What's the problem this feature will solve?**
Speed-Up virtualenv creation when using previously downloaded wheels.

**Describe the solution you'd like**
Cache the fact that it was already downloaded in the relevant JSON embed file.

Real world example:
cibuildwheel uses pinned versions of all those packages, this would speed-up the virtual environment creation.

Timings:
Using CPython 3.9 on my macOS laptop, I expect the creation time to drop from around 3 seconds to less than 1 second. 

**Alternative Solutions**
Custom seeder but it wouldn't benefit everyone for what I consider a base feature. Furthermore, the API is marked as unstable in the docs.

**Additional context**
Needed evolution to the JSON embed file format can be taken into consideration when solving #2265, #2266, #2267

mayeut:
Add downloaded wheel information in the relevant JSON embed file to prevent additional downloads of the same wheel.

This requires to bump the JSON embed file version once again as I was lacking some tests cases in #2273.

I will add some inline comments for some choices that I made.

Fixes #2268

### DIFF
--- a/src/virtualenv/seed/wheels/periodic_update.py
+++ b/src/virtualenv/seed/wheels/periodic_update.py
@@ -82,6 +82,19 @@ def handle_auto_update(distribution, for_py_version, wheel, search_dirs, app_dat
         trigger_update(distribution, for_py_version, wheel, search_dirs, app_data, periodic=True, env=env)
 
 
+def add_wheel_to_update_log(wheel, for_py_version, app_data):
+    embed_update_log = app_data.embed_update_log(wheel.distribution, for_py_version)
+    logging.debug("adding %s information to %s", wheel.name, embed_update_log.file)
+    u_log = UpdateLog.from_dict(embed_update_log.read())
+    if any(version.filename == wheel.name for version in u_log.versions):
+        logging.warning("%s already present in %s", wheel.name, embed_update_log.file)
+        return
+    # we don't need a release date for sources other than "periodic"
+    version = NewVersion(wheel.name, datetime.now(), None, "download")
+    u_log.versions.append(version)  # always write at the end for proper updates
+    embed_update_log.write(u_log.to_dict())
+
+
 DATETIME_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 
@@ -248,23 +261,27 @@ def _run_do_update(app_data, distribution, embed_filename, for_py_version, perio
     embed_update_log = app_data.embed_update_log(distribution, for_py_version)
     u_log = UpdateLog.from_dict(embed_update_log.read())
     now = datetime.now()
+
+    update_versions, other_versions = [], []
+    for version in u_log.versions:
+        if version.source in {"periodic", "manual"}:
+            update_versions.append(version)
+        else:
+            other_versions.append(version)
+
     if periodic:
         source = "periodic"
-        # mark everything not updated manually as source "periodic"
-        for version in u_log.versions:
-            if version.source != "manual":
-                version.source = source
     else:
         source = "manual"
-        # mark everything as source "manual"
-        for version in u_log.versions:
-            version.source = source
+        # mark the most recent one as source "manual"
+        if update_versions:
+            update_versions[0].source = source
 
     if wheel_filename is not None:
         dest = wheelhouse / wheel_filename.name
         if not dest.exists():
             copy2(str(wheel_filename), str(wheelhouse))
-    last, last_version, versions = None, None, []
+    last, last_version, versions, filenames = None, None, [], set()
     while last is None or not last.use(now, ignore_grace_period_ci=True):
         download_time = datetime.now()
         dest = acquire.download_wheel(
@@ -276,13 +293,14 @@ def _run_do_update(app_data, distribution, embed_filename, for_py_version, perio
             to_folder=wheelhouse,
             env=os.environ,
         )
-        if dest is None or (u_log.versions and u_log.versions[0].filename == dest.name):
+        if dest is None or (update_versions and update_versions[0].filename == dest.name):
             break
         release_date = release_date_for_wheel_path(dest.path)
         last = NewVersion(filename=dest.path.name, release_date=release_date, found_date=download_time, source=source)
         logging.info("detected %s in %s", last, datetime.now() - download_time)
         versions.append(last)
-        last_wheel = Wheel(Path(last.filename))
+        filenames.add(last.filename)
+        last_wheel = last.wheel
         last_version = last_wheel.version
         if embed_version is not None:
             if embed_version >= last_wheel.version_tuple:  # stop download if we reach the embed version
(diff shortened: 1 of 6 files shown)


src/virtualenv/seed/wheels/periodic_update.py
```diff
@@ -82,6 +82,18 @@ def handle_auto_update(distribution, for_py_version, wheel, search_dirs, app_dat
...
+    logging.debug("adding %s information to %s", wheel.name, embed_update_log.file)
+    u_log = UpdateLog.from_dict(embed_update_log.read())
+    if any(version.filename == wheel.name for version in u_log.versions):
+        logging.warning("%s already present in %s", wheel.name, embed_update_log.file)
+        return
+    version = NewVersion(wheel.name, datetime.now(), release_date_for_wheel_path(wheel.path), "download")
+    u_log.versions.append(version)  # always write at the end for proper updates
+    embed_update_log.write(u_log.to_dict())
+
+
```
mayeut, line 96: This has been added to `periodic_update.py` as it is the only place dealing with the embed_update_log (which is not used only for updates anymore).

src/virtualenv/seed/wheels/periodic_update.py
```diff
@@ -250,16 +265,20 @@ def _run_do_update(app_data, distribution, embed_filename, for_py_version, perio
     now = datetime.now()
     if periodic:
         source = "periodic"
-        # mark everything not updated manually as source "periodic"
```
mayeut, line 253: All the modifications in `_run_do_update` are what lead me to bump the JSON embed format version.

gaborbernat (approved):

src/virtualenv/seed/wheels/periodic_update.py
```diff
@@ -291,6 +318,8 @@ def _run_do_update(app_data, distribution, embed_filename, for_py_version, perio
     if not u_log.periodic:
         u_log.started = now
     u_log.versions = versions + u_log.versions
+    # we need to sort versions for the next update to work properly
+    u_log.versions.sort(reverse=True)
```
mayeut, line 322: The log needs to be sorted. It was sorted by design before. That's not the case anymore.

gaborbernat: Can you explain why is no longer the case?

mayeut:
There's another way to do the update which might be a less bit risky for periodic updates:
- Step A: filter the versions from the UpdateLog to only have `manual`/`periodic` sources on one hand (back to the previous problem) and other sources on the other hand.
- Step B: run the update as was done before
- Step C: filter-out newly found versions from the "other sources" list
- Step D: concatenate the two versions list (the first half being from version update & the 2nd half from other sources)

This would remove the need to sort the list.

src/virtualenv/seed/wheels/periodic_update.py
gaborbernat: > This would remove the need to sort the list. 

I don't mind sorting the log; why would you consider the other solution less risky?

mayeut:
Caveat:
There is a time penalty for first downloads compared to the previous version in order to fill the release_date.
Now that I'm writing this, I now realize it might not be necessary to fill this one which is only ever used by `NewVersion.use` when the source is `periodic`.

mayeut:
@gaborbernat,

Writing things down makes me realize that there are still some things left to do so converting to draft.
Do you have any remarks to share about the comments I left ?

PS: while thinking about the time penalty I mentioned, I thought about caching PyPI requests on disk as well and use the `etag` header to know if actually pulling the data (`setuptools` JSON is about 1MB) was necessary. I deemed this was too heavy to implement for my simple use-case and not worth spending more time on this. I thought of another use-case which might benefit from this: users configuring `--download` as the default setting (checking if there was an update on PyPI should be much faster than starting a pip download. I guess asking PyPI this way is the first thing done by pip)

src/virtualenv/seed/wheels/periodic_update.py
mayeut: > Can you explain why is no longer the case?

because other sources logs are always appended rather than inserted.

> I don't mind sorting the log; why would you consider the other solution less risky?

I'm feeling a bit hesitant to rely on the tuple ordering given:
```
    def as_version_tuple(version):
        result = []
        for part in version.split(".")[0:3]:
            try:
                result.append(int(part))
            except ValueError:
                break
        if not result:
            raise ValueError(version)
        return tuple(result)
```
I feel that the inner try/catch is hiding something I did not think about.

mayeut: I did not think of pre-release versions... In this specific sample, it does not matter much but it's not sorted correctly.
```
{
  "completed": "2022-01-01T21:48:50.557509Z",
  "periodic": false,
  "started": "2022-01-01T21:48:40.230835Z",
  "versions": [
    {
      "filename": "pip-21.3.1-py3-none-any.whl",
      "found_date": "2022-01-01T14:22:37.717326Z",
      "release_date": "2021-10-22T15:57:09.000000Z",
      "source": "manual"
    },
    {
      "filename": "pip-20.3-py2.py3-none-any.whl",
      "found_date": "2022-01-01T21:46:25.286240Z",
      "release_date": "2020-11-30T12:47:49.000000Z",
      "source": "manual"
    },
    {
      "filename": "pip-20.0-py2.py3-none-any.whl",
      "found_date": "2022-01-01T21:48:23.634459Z",
      "release_date": "2020-01-21T11:42:56.000000Z",
      "source": "manual"
    },
    {
      "filename": "pip-20.3b1-py2.py3-none-any.whl",
      "found_date": "2022-01-01T21:44:30.798536Z",
      "release_date": "2020-10-31T19:25:12.000000Z",
      "source": "manual"
    }
  ]
}
```
Further-more, a manual/periodic update should never lead to defaulting to pre-release version. This promise is broken (at least for manual updates).

mayeut: Given an added test with a pre-release version downloaded as "pinned" version indeed broke manual updates, I went with the alternate implementation of _run_do_update which first splits the UpdateLog in 2 depending if it's coming from a previous update or another source.

mayeut:
I rewrote the `_run_do_update` function as mentioned in https://github.com/pypa/virtualenv/pull/2276#issuecomment-1003611763 and removed the time penalty I mentioned in https://github.com/pypa/virtualenv/pull/2276#issuecomment-1003613103.

tests/unit/seed/wheels/test_periodic_update.py
```diff
@@ -559,51 +602,58 @@ def rel_date():
 
     assert download_wheel.call_count == 2
     assert release_date.call_count == 2
-
     assert read_dict.call_count == 1
     assert write.call_count == 1
 
 
-def test_download_manual_stop_at_first_usable(tmp_path, mocker, freezer):
+def test_download_periodic_stop_at_first_usable2(tmp_path, mocker, freezer):
```
gaborbernat, line 609: Why the 2 suffix?

gaborbernat (approved):

tests/unit/seed/wheels/test_periodic_update.py
mayeut: there's already a `test_download_periodic_stop_at_first_usable` test and I didn't see how to do this as a parametrize test easily so adding a suffix `2` on the second one.

gaborbernat: Instead of magic numbers, I generally add to the test name what differentiates them :thinking: for future reference.

Issue #2303, opened 2022-02-15
virtualenv can't be installed as `--user` when an older global installation exists. 


mboisson:
**Issue**

I have the following: 
```
$ virtualenv --version
virtualenv 20.0.18 from /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/Core/python/3.8.10/lib/python3.8/site-packages/virtualenv/__init__.py
$ virtualenv
usage: virtualenv [--version] [--with-traceback] [-v | -q] [--app-data APP_DATA] [--clear-app-data] [--discovery {builtin}] [-p py] [--creator {builtin,cpython3-posix,venv}] [--seeder {app-data,pip}] [--no-seed]
                  [--activators comma_sep_list] [--clear] [--system-site-packages] [--symlinks | --copies] [--download | --no-download] [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--wheel version] [--no-pip]
                  [--no-setuptools] [--no-wheel] [--symlink-app-data] [--prompt prompt] [-h]
                  dest
virtualenv: error: the following arguments are required: dest
``` 
if I now try to upgrade `virtualenv` in my user account: 
```
$ pip install --user --upgrade virtualenv
Ignoring pip: markers 'python_version < "3"' don't match your environment
Looking in links: /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo/avx2, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo/generic, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/generic
Processing /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/generic/virtualenv-20.13.1+computecanada-py2.py3-none-any.whl
Collecting filelock<4,>=3.2
  Using cached filelock-3.5.0-py3-none-any.whl (10.0 kB)
Processing /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/generic/platformdirs-2.5.0+computecanada-py3-none-any.whl
Requirement already satisfied, skipping upgrade: six<2,>=1.9.0 in /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/Core/python/3.8.10/lib/python3.8/site-packages (from virtualenv) (1.14.0)
Collecting distlib<1,>=0.3.1
  Using cached distlib-0.3.4-py2.py3-none-any.whl (461 kB)
[...]

PR #2304, opened 2022-02-16
use importlib_metadata instead of importlib.metadata for python<3.10

mboisson:

This PR is pretty minimal and is meant to address https://github.com/pypa/virtualenv/issues/2303
`importlib.metadata` does not provide unique results for `entry_points` when duplicate packages are found until python >= 3.10. To allow local installation of virtualenv to superseed global installations, we need to use `importlib_metadata` instead. 

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ install_requires =
     filelock>=3.2,<4
     platformdirs>=2,<3
     six>=1.9.0,<2   # keep it >=1.9.0 as it may cause problems on LTS platforms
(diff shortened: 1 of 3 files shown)


gaborbernat:
Can we find out why using the native  `importlib.metadata` does not work?

gaborbernat:
Closing per https://github.com/pypa/virtualenv/issues/2303#issuecomment-1042004390 Please consider using instead https://pypa.github.io/pipx via https://pypi.org/project/pipx-in-pipx/

Issue #2309, opened 2022-03-05
PyPy 3.8 virtualenvs lost system isolation on Debian


stefanor:
**Issue**

Virtualenvs created with (Debian) PyPy 3.8 on Debian have lost system isolation.

```console
root@evident-guppy:/tmp# testve/bin/python -c 'import sys; print(sys.path)'
['', '/usr/lib/pypy3.8', '/tmp/testve/lib/pypy3.8/site-packages', '/tmp/testve/lib/python3/dist-packages']
root@evident-guppy:/tmp# ls /tmp/testve/lib/python3/dist-packages
__pycache__		 pip-22.0.2.dist-info	       six.py
_distutils_hack		 pkg_resources		       virtualenv
distlib			 platformdirs		       virtualenv-20.13.0+ds.dist-info
distlib-0.3.4.egg-info	 platformdirs-2.5.1.dist-info  wheel
filelock		 setuptools		       wheel-0.37.1.egg-info
filelock-3.6.0.egg-info  setuptools-59.6.0.egg-info
pip			 six-1.16.0.egg-info
root@evident-guppy:/tmp# ls -l /tmp/testve/lib/python3
lrwxrwxrwx 1 root root 16 Mar  5 00:58 /tmp/testve/lib/python3 -> /usr/lib/python3
root@evident-guppy:/tmp# ls -l /tmp/testve/lib
total 148
lrwxrwxrwx 1 root root 12 Mar  5 00:58 apt -> /usr/lib/apt
lrwxrwxrwx 1 root root 20 Mar  5 00:58 bfd-plugins -> /usr/lib/bfd-plugins
lrwxrwxrwx 1 root root 17 Mar  5 00:58 binfmt.d -> /usr/lib/binfmt.d
lrwxrwxrwx 1 root root 18 Mar  5 00:58 compat-ld -> /usr/lib/compat-ld
lrwxrwxrwx 1 root root 12 Mar  5 00:58 cpp -> /usr/lib/cpp
lrwxrwxrwx 1 root root 17 Mar  5 00:58 dbus-1.0 -> /usr/lib/dbus-1.0
lrwxrwxrwx 1 root root 13 Mar  5 00:58 dpkg -> /usr/lib/dpkg
lrwxrwxrwx 1 root root 22 Mar  5 00:58 environment.d -> /usr/lib/environment.d
lrwxrwxrwx 1 root root 12 Mar  5 00:58 gcc -> /usr/lib/gcc
[...]

PR #2310, opened 2022-03-05
Avoid symlinking the contents of /usr into PyPy3 virtualenvs

stefanor:
PyPy >= 3.8 supports a standard prefix installation, where older versions always used a portable/developent style installation.
If this is a standard prefix installation, skip the logic that symlinks libraries in a portable PyPy distribution.

Fixes: #2309


src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
```diff
@@ -42,6 +42,11 @@ def to_lib(self, src):
     def sources(cls, interpreter):
         for src in super(PyPy3Posix, cls).sources(interpreter):
             yield src
+        # PyPy >= 3.8 supports a standard prefix installation, where older
+        # versions always used a portable/developent style installation.
+        # If this is a standard prefix installation, skip the below:
+        if interpreter.system_prefix == "/usr":
```
gaborbernat, line 48: Shouldn't this be a sysconfig path rather than a constant? Is it possible to add a test for this?

gaborbernat (changes requested):

src/virtualenv/create/via_global_ref/builtin/pypy/pypy3.py
stefanor: > Shouldn't this be a sysconfig path rather than a constant?

Yeah, I found myself looking for something less static, but I couldn't come up with anything better.
Any sysconfig variables from the virtualenv host wouldn't be relevant here. The ideal thing to check would be the selected sysconfig scheme on the target interpreter, but that isn't available. Do you think we should make it available?

> Is it possible to add a test for this?

What kind of thing are you looking for here? I couldn't find any tests related to any similar functionality, to extend. So it seems like I'd have to build an awful amount of test infrastructure, to test a very minor behavioural change.

gaborbernat: Doesn't sound minor to me 🙂 the hard part in new feature additions is always the test, never the functionality change. And yes that does imply writing new test infrastructure for it.

stefanor: How's this? I had to mock `Path()` but otherwise it isn't *too* ugly.

gaborbernat (approved):

gaborbernat:
This now has been released as https://pypi.org/project/virtualenv/20.13.3/

hroncok:
We just discovered this issue on Fedora 35 and spent some time trying to figure out what's going on. Thanks for fixing it @stefanor!

PR #2325, opened 2022-03-29
Support Python 2.7 framework-style distributions on macOS 12

nickhutchinson:
virtualenv's trick of adding a symlink to the base Python dylib (`/Library/Frameworks/Python.framework/Versions/2.7/Python`) in the virtualenv root, and rewriting the dependent dylib list of the `<virtualenv>/bin/python` binary with the path of this symlink, no longer works in macOS 12.

As of macOS 12, when <https://github.com/python/cpython/blob/v2.7.18/Modules/getpath.c> computes `sys.prefix`, it ends up resolving the symlink. The result is that `sys.prefix` points to the Python framework, e.g.  `/path/to/Python.framework/Versions/2.7` instead of at the virtualenv root.

To mitigate this, add a _copy_ of the Python dylib into the virtualenv instead of a symlink. To placate code signing (at least on Intel machines), add a symlink to the `Resources` dir so the `Info.plist` file is accessible.

Note that Apple's Python 2.7 distribution -- removed in macOS 12.3 but previously located at /System/Library/Frameworks/Python.framework -- behaved differently to the python.org builds. It appears to have incorporated custom patches to getpath.c to support virtualenv. See: <https://github.com/apple-oss-distributions/python/blob/python-170.80.2/2.7/fix/getpath.c.ed>.

Tested on macOS 12.3 on an Intel and M1 mac.

Fixes: #2284

### DIFF
--- a/docs/changelog/2284.feature.rst
+++ b/docs/changelog/2284.feature.rst
@@ -1,0 +1,1 @@
+Support for creating a virtual environment from a Python 2.7 framework on macOS 12 - by :user:`nickhutchinson`.
--- a/src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
+++ b/src/virtualenv/create/via_global_ref/builtin/cpython/mac_os.py
@@ -25,20 +25,11 @@ class CPythonmacOsFramework(CPython):
     def can_describe(cls, interpreter):
         return is_mac_os_framework(interpreter) and super(CPythonmacOsFramework, cls).can_describe(interpreter)
 
-    @classmethod
-    def sources(cls, interpreter):
-        for src in super(CPythonmacOsFramework, cls).sources(interpreter):
-            yield src
-        # add a symlink to the host python image
-        exe = cls.image_ref(interpreter)
-        ref = PathRefToDest(exe, dest=lambda self, _: self.dest / ".Python", must=RefMust.SYMLINK)
-        yield ref
-
     def create(self):
         super(CPythonmacOsFramework, self).create()
 
         # change the install_name of the copied python executables
-        target = "@executable_path/../.Python"
+        target = self.desired_mach_o_image_path()
         current = self.current_mach_o_image_path()
         for src in self._sources:
             if isinstance(src, ExePathRefToDest):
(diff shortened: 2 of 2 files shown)


docs/changelog/2284.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix for creating virtualenvs from a Python 2.7 framework on macOS 12.3 - by :user:`nickhutchinson`.
```
gaborbernat, line 1: I consider this a feature rather than a bugfix, so let's move it to a feature file 👍 something like support for macos 12.3 👍

nickhutchinson:
I want to try and test in a macOS 12.2 VM -- rereading the original thread more carefully, it might be that the behaviour change was actually in macOS 12.0, and Apple's Python 2 (removed in macOS 12.3) had some custom patches to work around the issue.

nickhutchinson:
> I want to try and test in a macOS 12.2 VM -- rereading the original thread more carefully, it might be that the behaviour change was actually in macOS 12.0, and Apple's Python 2 (removed in macOS 12.3) had some custom patches to work around the issue.

This does seem to be the case, and I've updated the PR description and commit message to be more accurate. Virtualenv worked with Apple's Python 2.7 on macOS [12,12.3) -- though _not_ with Python 2.7 distributions from python.org  -- likely because they applied a bunch of custom patches, including this one: <https://github.com/apple-oss-distributions/python/blob/python-170.80.2/2.7/fix/getpath.c.ed>.

hynek:
> I want to try and test in a macOS 12.2 VM -- rereading the original thread more carefully, it might be that the behaviour change was actually in macOS 12.0, and Apple's Python 2 (removed in macOS 12.3) had some custom patches to work around the issue.

This is 100% the case; I remember switching to Apple's 2.7 because "suddenly" virtualenv was broken and forgetting to further investigate.

gaborbernat (approved):

gaborbernat (changes requested):
@nickhutchinson can you make the CI pass? thanks!

evangrim:
@hynek You mentioned switching to Apple's 2.7. How did you manage that? I'm asking as an end user of an old virtualenv project, and I understand little about virtualenv or config related to it.

hynek:
@evangrim What I meant was that I didn't install Python 2 myself anymore & used the one that came with macOS. The point is moot tho, because Apple doesn't ship a Python 2 anymore.

Nowadays, the [pyenv version](https://github.com/pyenv/pyenv) works just fine with current virtualenv releases.

PR #2346, opened 2022-05-24
CI: Update actions and OSes, weekly cron job

EwoutH:
A bit of maintenance on the CI. This PR consists of 3 commits:

- _Update used actions to v3_
   Update the checkout, setup-python and upload-/download-artifact actions to v3

 - _Use the latest stable version of each OS_
   Use the latest stable version of Ubuntu, macOS and Windows. Currently those are Ubuntu 20.04, macOS 11 and Windows Server 2022, which means macOS is directly updated from 10.15 to 11, and all OS will be kept up to date.

   The scheduled cron job will catch any breaking changes due to changing environments (which aren't likely).
 - _Run cron job weekly (instead of daily)_
   Once a week is enough to catch changes in dependencies or the environment. Running this extensive CI daily is quite power intensive and thus has a large costs and environmental impact.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -13,8 +13,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
       - uses: pre-commit/action@v2.0.3
 
   test:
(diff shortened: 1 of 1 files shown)


EwoutH:
Hi @gaborbernat, would you be able to enable the workflow and take a look at this PR?

.github/workflows/check.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: "0 8 * * *"
```
gaborbernat, line 6: I'm -1 on this.

.github/workflows/check.yml
```diff
@@ -13,8 +13,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
       - uses: pre-commit/action@v2.0.3
```
gaborbernat, line 18: For this, a better solution is to set up dependabot to do it automatically, in future too.

.github/workflows/check.yml
```diff
@@ -36,11 +36,11 @@ jobs:
           - "2.7"
           - pypy-2.7
         os:
-          - ubuntu-20.04
-          - macos-10.15
-          - windows-2022
```
gaborbernat, line 41: I like the explicit nature here of the version used then this link. If some explicit version needs bumped we can do that separately.

gaborbernat (changes requested):
All three of these changes deserve their own PR, ideally with an issue discussing the change and approach to tackle it. In the current form I cannot accept this PR.

PR #2351, opened 2022-06-06
Use shlex.quote instead of deprecated pipes.quote

frenzymadness:
pipes module is deprecated in Py 3.11 and will be removed in 3.13.
https://docs.python.org/3.11/whatsnew/3.11.html

This is an implementation detail so I believe that it does not need a changelog entry.

### DIFF
--- a/docs/changelog/2351.bugfix.rst
+++ b/docs/changelog/2351.bugfix.rst
@@ -1,0 +1,1 @@
+Use ``shlex.quote`` instead of deprecated ``pipes.quote`` in Python 3. - by :user:`frenzymadness`.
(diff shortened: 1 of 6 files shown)


hroncok:
I am afraid the code was Python 2 compatible and is no more. The function was moved in 3.3+.

frenzymadness:
So, I can either use `if PY2` to import from the correct module or handle `ImportError` and then import from the deprecated module. @gaborbernat what do you prefer?

frenzymadness:
I've decided to use `if PY2` because conditions are generally faster than exception handling even if it means adding a new import to some of the modules. And it seems that `make_zipapp.py` module cannot import from virtualenv so I have to use a copy of `sys.version_info[0] == 2` there.

gaborbernat (commented):
Let's add a changelog here 👍

gaborbernat (approved):

PR #2363, opened 2022-06-18
Fixes for Windows PyPy 3.6

reksar:
This seems to be a separate issue with Windows PyPy 3.6

I ran `tests/unit/create/test_creator.py` in Windows PyPy 3.6 (v7.3.3) and got errors in the same tests as CI, except `test_create_distutils_cfg[venv]` that passes locally and fails on CI, most likely due to timeout.

Local tests fails when the `test_create_no_seed[venv*]` checks exe files, but CI errors causes when same tests find an exception in `creator.debug`:

```json
{
  "err": "The system cannot find the file specified",
  "exception": "SyntaxError('invalid syntax', ('<unknown>', 0, 1, '\\n', 0))", 
  "out": "",
  "returncode": 2
}
```

The `SyntaxError` was happened in `creator.py` because of `literal_eval(out)` when `out` is empty.

I can't reproduce the exception locally, but I modified the debug info to look something like:

```json
{
  "err": "[Errno 2] The system cannot find the file specified: '<file>'",
  "exception": "OSError<...>", 
  "out": "",
  "returncode": 2
}
```

Windows PyPy <= 3.6 contains only `pypy3.exe` and `pypy3w.exe` and creating a virtual environment with `venv` does not hanlde missing executables, e.g. `python.exe`. Also, the `stdlib` path for old PyPy versions is `<host>\site-packages` instead of `<host>\Lib\site-packages`.

Local tests failures are fixed by creating missing executables.

Test coverage calculation is fixed by adding `stdlib` path for old PyPy versions into `.coveragerc` sources.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -18,7 +18,8 @@ source =
     src
     .tox/*/lib/python*/site-packages
     .tox/pypy*/site-packages
-    .tox\*\Lib\site-packages\
+    .tox\*\Lib\site-packages
+    .tox\py\site-packages
     */src
     *\src
 
--- a/src/virtualenv/create/creator.py
+++ b/src/virtualenv/create/creator.py
@@ -227,7 +227,13 @@ def get_env_debug_info(env_exe, debug_script, app_data, env):
     # noinspection PyBroadException
     try:
         if code != 0:
-            result = literal_eval(out)
+            if out:
+                result = literal_eval(out)
+            else:
+                if code == 2 and "file" in err:
+                    # Re-raise FileNotFoundError from `run_cmd()`
+                    raise OSError(err)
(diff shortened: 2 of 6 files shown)


gaborbernat (approved):

reksar:
😀
I believe we'll catch the bug.

gaborbernat:
> I believe we'll catch the bug.

Catch as in?

reksar:
Windows PyPy 3.6 test may throw an OSError to be fixed further.

gaborbernat:
Can you fix it? 😆

gaborbernat:
Thanks, will cut a release once we fix it 👍

PR #2387, opened 2022-07-25
fix setup-cfg-fmt

asottile:


gaborbernat:
Thanks I've already done this here https://github.com/pypa/virtualenv/pull/2384/commits/c0ba7809fb5d077ca12d4ab0dee4057513f908e5

PR #2394, opened 2022-08-04
Bump dependencies and build tools

gaborbernat:


setup.cfg
```diff
@@ -38,11 +33,11 @@ project_urls =
...
 install_requires =
-    distlib>=0.3.1,<1
-    filelock>=3.2,<4
-    platformdirs>=2,<3
-    importlib-metadata>=0.12;python_version<"3.8"
-    importlib-resources>=1.0;python_version<"3.7"
+    distlib>=0.3.5,<1
+    filelock>=3.4.1,<4
+    platformdirs>=2.4,<3
+    importlib-metadata>=4.8.3;python_version < "3.8"
```
hneiva, line 39: I had issues when requiring `tox` and `flake8` because of this pin.
Shortened error from `pip-compile-multi`:
```
Could not find a version that matches importlib-metadata<4.3,>=0.12,>=0.22,>=1.1.0,>=4.8.3 (from tox==3.25.1->-r requirements/local.in (line 3))
Tried: ............
There are incompatible versions in the resolved dependencies:
  .......
  importlib-metadata<4.3,>=1.1.0 (from flake8==5.0.4->-r requirements/test.in (line 6))
  importlib-metadata>=4.8.3 (from virtualenv==20.16.3->tox==3.25.1->-r requirements/local.in (line 3))
```

In my case, pinning `virtualenv==20.16.2 ; python_version < '3.7'` avoids the problem, but my project will be locked on `virtualenv` version 20.16.2 until this can be resolved.

Is there a reason for `importlib-metadata` to be pinned as `>=4.8.3`? I'm not super familiar with the library, and the problems around it, but unfortunately looks this isn't the only [issue](https://github.com/PyCQA/flake8/commit/975a3f45334861ebd8960c00e881443c23654bca) out there when projects pin it.

gaborbernat: > Is there a reason for `importlib-metadata` to be pinned as `>=4.8.3`?

That's the only version we actually test with in our CI and is the only version I'm comfortable saying it's supported. I recommend reaching out to other projects that upper pin to remove.

PR #2401, opened 2022-08-09
When setting prefixes to empty strings, check the key names in addition to value

hroncok:
In Fedora, we are currently trying to come up with a safe approach that
will make `sudo pip` and similar install to /usr/local rather than /usr.

After years of only patching distutils for this,
we need to adapt our change to use sysconfig.

In the past, we've tried to use a custom installation scheme,
but that has yielded many unexpected side effects for our users.

Now we try to set the defaults of {base} and {platbase} to /usr/local instead.
However, such value is not in the prefixes listed in this check in virtualenv.

Here, I adapted the code to be more robust for such changes:

 - if the value is listed in prefixes, we still set to empty
 - if the key is listed in prefix_vars, we now also set to empty

Let me know if this is acceptable to virtualenv. An alternative would be something like:

```python
if hasattr(sysconfig, 'extra_prefixes'):
    prefixes = prefixes + tuple(sysconfig.extra_prefixes)
```
Or:

```python
if hasattr(sys, 'insatllation_prefix'):
    prefixes = prefixes + (sys.insatllation_prefix,)
```

I plan to work on tests and docs after we agree on the best approach to fix this.

### DIFF
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -136,7 +136,10 @@ class PythonInfo(object):
         if result is None:  # use sysconfig if sysconfig_scheme is set or distutils is unavailable
             # set prefixes to empty => result is relative from cwd
             prefixes = self.prefix, self.exec_prefix, self.base_prefix, self.base_exec_prefix
-            config_var = {k: "" if v in prefixes else v for k, v in self.sysconfig_vars.items()}
+            # Fedora-based distributions mangle the default value of some of those to /usr/local
+            # so we are extra careful here: we set them to empty based on value and key
+            prefix_vars = "prefix", "exec_prefix", "installed_base", "base", "installed_platbase", "platbase"
+            config_var = {k: "" if v in prefixes or k in prefix_vars else v for k, v in self.sysconfig_vars.items()}
             result = self.sysconfig_path(key, config_var=config_var).lstrip(os.sep)
         return result
 


hroncok:
I am unsure if the *check / test brew@3.9 - macos-12* failure is relevant.

gaborbernat:
> I am unsure if the _check / test brew@3.9 - macos-12_ failure is relevant.

It doesn't fail on main branch though :thinking:

src/virtualenv/discovery/py_info.py
```diff
@@ -136,7 +136,10 @@ def install_path(self, key):
         if result is None:  # use sysconfig if sysconfig_scheme is set or distutils is unavailable
             # set prefixes to empty => result is relative from cwd
             prefixes = self.prefix, self.exec_prefix, self.base_prefix, self.base_exec_prefix
-            config_var = {k: "" if v in prefixes else v for k, v in self.sysconfig_vars.items()}
+            # Fedora-based distributions mangle the default value of some of those to /usr/local
+            # so we are extra careful here: we set them to empty based on value and key
+            prefix_vars = "prefix", "exec_prefix", "installed_base", "base", "installed_platbase", "platbase"
+            config_var = {k: "" if v in prefixes or k in prefix_vars else v for k, v in self.sysconfig_vars.items()}
```
gaborbernat, line 142: I'm not on board with this change. Why should we ignore the answer the python interpreter gives back when asked about his layout? This feels wrong.

gaborbernat (changes requested):

src/virtualenv/discovery/py_info.py
hroncok: So, if I get this right, the current code decides to store a dictionary of expanded config vars and it recognizes the following two possibilities:

 - either the returned value is the same as one of the known prefixes from `sys` -> in that case, it stores an empty string
 - or else it stores the returned value

This serves as a (rather complex but quite robust) mechanism to be able to expand *relative* paths in new virtual environments.

--------------------------

The addition here *expands* the first conditional to also check for known prefix vars from `sysconfig`. It does not "ignore the answer the python interpreter gives back" any more than the original solution. So maybe I don't really understand your question.

--------------------------

As said, we'd like to make it so that the `{base}` and `{platbase}` config vars expand to `/usr/local` in Fedora, which *is not* `sys.prefix` (that one is `/usr`). I know this is a bit hackish, but we've tried to roll with the installation scheme layouts and it backfired on us, so we are going back to the previous way of things (but this time in `sysconfig` rather than `distutils`)  -- I can expand the reasoning for that in more detail if you want me to.

virtualenv is one of the things that break with our approach: because it explicitly cleans the vars that would otherwise return a prefix rather than setting the bases vars to empty.

As said, this PR here is one fo the possible solutions. The other I could think of is that we set another property on `syconfig` or `sys` which would list the additional prefix(es) and that could be added to `prefixes` here. Would that be more acceptable?

gaborbernat: > * either the returned value is the same as one of the known prefixes from `sys` -> in that case, it stores an empty string

This is what we do today:

https://github.com/pypa/virtualenv/blob/08b413eca8e6d91cb9b5bfc16a8e14820024a878/src/virtualenv/discovery/py_info.py#L109-L113
https://github.com/pypa/virtualenv/blob/08b413eca8e6d91cb9b5bfc16a8e14820024a878/src/virtualenv/discovery/py_info.py#L220-L222

So if I understand this correctly we do not ignore any variables. We only replace the values that start with the current interpreters prefix with the systems prefix. As far as I remember this was done so that even if we query a virtual environment we get back the information of the system interpreter.

In your new change, we explicitly start ignoring values, effectively lying about what the system variables are. I don't see how this is either robust or correct.

> This serves as a (rather complex but quite robust) mechanism to be able to expand _relative_ paths in new virtual environments.

I do not understand this line.

> I can expand the reasoning for that in more detail if you want me to.
> In Fedora, we are currently trying to come up with a safe approach that
> will make `sudo pip` and similar install to /usr/local rather than /usr.

Please do. This sounds like a pip problem to me, did you talk with the pip people about it, did they recommend this solution or you're trying to roll your own? How about with the `venv` people, how do they handle this use case?

hroncok: Just to be sure, the current code I talk about is:

https://github.com/pypa/virtualenv/blob/08b413eca8e6d91cb9b5bfc16a8e14820024a878/src/virtualenv/discovery/py_info.py#L136-L140

It says "set prefixes to empty". And this is exactly what this is still doing after the change I've proposed here. So I am a bit confused about why what I propose is "laying" but the current code is not. Maybe we understand the purposes differently?

Anyway, as said, would simply expanding the list of `prefixes` in the code above be less controversial?

I plan to expand the Fedora's reasoning more tomorrow because I need to sit down and write it and I imagine it'll take couple of hours. In the meantime:

No, this is not a pip problem, it's a problem of any tool that installs packages (e.g. pypa/installer), so it needs to be tackled in sysconfig (and distutils at least until they die).

The venv solution of what I propose here is:

https://github.com/python/cpython/blob/32ac98e8992caa47ff31e3021444949df058e413/Lib/venv/__init__.py#L96-L103

Inlined here as GitHub does not seem to inline code from different projects:

```python
    def _venv_path(self, env_dir, name):
        vars = {
            'base': env_dir,
            'platbase': env_dir,
            'installed_base': env_dir,
            'installed_platbase': env_dir,
        }
        return sysconfig.get_path(name, scheme='venv', vars=vars)
```

In other words, venv is doing exactly what I propose here (except it uses absolute paths here, not relative).

pfmoore: > The venv solution of what I propose here is

So is the issue that virtualenv does not (yet) respect the (new in Python 3.11) `venv` scheme in sysconfig? How will `venv` work on older versions of Python under the new Fedora approach?

Using the `venv` scheme seems reasonable, but virtualenv has to work on versions of Python that don't have it, and I'm not clear what you'd expect to happen then.

> In other words, venv is doing exactly what I propose here (except it uses absolute paths here, not relative).

I don't see how that's the same as what you're proposing. It's using a different scheme.

hroncok: I see now that I've done a poor job at explaining what this is about. This is not about installation schemes. virtualenv already uses the `venv` scheme when available since https://github.com/pypa/virtualenv/pull/2209

This issue is not about the venv scheme at all. It's about values used for expanding the returned paths.

I plan to write a better rationale, explaining the problem we have and the solution I'm looking for. But this will take time. Sorry about not starting with that in the first place.

gaborbernat: > This issue is not about the venv scheme at all. It's about values used for expanding the returned paths.

Changing the values expanded isn't just another way to change the scheme without changing the scheme? I feel you could solve the problem both with a new scheme and patching the variables used for expansion. The final path is scheme + vars, so changing any part of it will get you there no?

> No, this is not a pip problem, it's a problem of any tool that installs packages (e.g. pypa/installer), so it needs to be tackled in sysconfig (and distutils at least until they die).

I still don't see why it needs to be tackled in syconfig. Couldn't both pip, installer, distutils adopt the solution we come up with? I'd imagine pip will adopt installer at some point for installation either way, and distutils is deprecated since 3.12; so really only need to solve the problem in installer.

hroncok: You actually gave me an idea. Will check it out and report back.

hroncok: OK, so your comments made me realize I can change the value of base and platbase if and only if the posix_prefix scheme is expanded. That makes virtualenv work thanks to https://github.com/pypa/virtualenv/pull/2209

Thank you and sorry for the noise.

I still plan to write an explanation about what is happening, but I will probably post it to https://discuss.python.org/

PR #2412, opened 2022-09-06
Do not turn echo off for subsequent commands in batch activators (#2411)

### LINKED ISSUE #2411: Batch activators unexpectedly turn echo off for subsequent commands
**Issue**

Consider the following batch script:
```batch
python -m virtualenv --quiet venv
CALL venv\Scripts\activate.bat
REM spam
CALL deactivate.bat
REM eggs
```
The expected output is:
```console
W:\>python -m virtualenv --quiet venv

W:\>CALL venv\Scripts\activate.bat

(venv) W:\>REM spam

(venv) W:\>CALL deactivate.bat

W:\>REM eggs

```
The actual output is:
```console
W:\>python -m virtualenv --quiet venv

W:\>CALL venv\Scripts\activate.bat

```
This is caused by `@echo off` at the beginning of both scripts that need to be undone after calling them:
```batch
python -m virtualenv --quiet venv
CALL venv\Scripts\activate.bat
@ECHO ON
REM spam
CALL deactivate.bat
@ECHO ON
REM eggs
```

**Environment**

- OS: Windows 10
- `pip list`: not relevant

Also not really relevant. `virtualenv` version 20.16.4.

pawelszramowski:
Some `@` are certainly unnecessary but I don't feel confident enough to optimize.

Closes #2411

### DIFF
--- a/docs/changelog/2411.bugfix.rst
+++ b/docs/changelog/2411.bugfix.rst
@@ -1,0 +1,2 @@
+Do not turn echo off for subsequent commands in batch activators
+(``activate.bat`` and ``deactivate.bat``) - by :user:`pawelszramowski`.
--- a/src/virtualenv/activation/batch/activate.bat
+++ b/src/virtualenv/activation/batch/activate.bat
@@ -1,39 +1,37 @@
-@echo off
+@set "VIRTUAL_ENV=__VIRTUAL_ENV__"
 
-set "VIRTUAL_ENV=__VIRTUAL_ENV__"
-
-if defined _OLD_VIRTUAL_PROMPT (
-    set "PROMPT=%_OLD_VIRTUAL_PROMPT%"
+@if defined _OLD_VIRTUAL_PROMPT (
+    @set "PROMPT=%_OLD_VIRTUAL_PROMPT%"
 ) else (
-    if not defined PROMPT (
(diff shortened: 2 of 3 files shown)


gaborbernat (changes requested):
Who this does waaaay to much change, that I'm not comfortable with. We should only make it as much as possible. Also did you see how does the venv module handles this and why is not an issue there?

pawelszramowski:
Did you notice that except for removing the first `@echo off` lines all the other changes are just adding a single `@` before each command? [`echo off` turns echo off for the current `cmd.exe` instance, `@` turns it off for one line](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/echo). `@echo off` is convenient and clean but there's no reasonable way to restore the caller's setting when you're done (yeah, batch is crap):
* https://stackoverflow.com/questions/9592732/restore-the-previous-echo-status
* https://stackoverflow.com/questions/48420074/was-echo-on-or-off-when-my-windows-bat-was-called

`venv` has got the same issue.

pfmoore:
I can confirm, while this is pretty messy it is indeed the best way of doing this.

gaborbernat (approved):

gaborbernat:
Can you please fill in a similar PR for venv? thanks!

PR #2418, opened 2022-09-14
Fix test_csh when running with the original csh

### LINKED ISSUE #2409: test_csh fails on Solaris due to unexpected prompt in the output
**Issue**

When running the virtualenv test suite on Solaris, I can see both tests in **tests/unit/activation/test_csh.py** fail.

```
    def test_csh(activation_tester_class, activation_tester):
        class Csh(activation_tester_class):
            def __init__(self, session):
                super().__init__(CShellActivator, session, "csh", "activate.csh", "csh")

            def print_prompt(self):
                return "echo 'source \"$VIRTUAL_ENV/bin/activate.csh\"; echo $prompt' | csh -i"

>       activation_tester(Csh)

.....

    def assert_output(self, out, raw, tmp_path):
        # pre-activation
        assert out[0], raw
        assert out[1] == "None", raw
        # post-activation
        expected = self._creator.exe.parent / os.path.basename(sys.executable)
        assert self.norm_path(out[2]) == self.norm_path(expected), raw
        assert self.norm_path(out[3]) == self.norm_path(self._creator.dest).replace("\\\\", "\\"), raw
        # Some attempts to test the prompt output print more than 1 line.
        # So we need to check if the prompt exists on any of them.
        prompt_text = f"({self._creator.env_name}) "
        assert any(prompt_text in line for line in out[4:-3]), raw

>       assert out[-3] == "wrote pydoc_test.html", raw
E       AssertionError: /..../virtualenv/virtualenv-20.16.3/.tox/py39/bin/python
E         None
E         /tmp/pytest-of-jkulik/pytest-39/activation-tester-env9/e-$ __________-j/bin/python
E         /tmp/pytest-of-jkulik/pytest-39/activation-tester-env9/e-$ __________-j
E         Warning: no access to tty; thus no job control in this shell...
E         myvm-i386-25280% (e-$ __________-j) myvm-i386-25280%
E         (e-$ __________-j) myvm-i386-25280% wrote pydoc_test.html
E         /..../virtualenv/virtualenv-20.16.3/.tox/py39/bin/python
E         None
E
E       assert '(e-$ __________...doc_test.html' == 'wrote pydoc_test.html'
E         - wrote pydoc_test.html
E         + (e-$ __________-j) myvm-i386-25280% wrote pydoc_test.html
```
I executed the test suite simply by downloading the sources (from PyPI) and running `tox` in that directory. I can see the same thing when running `pytest` directly.

It seems that the test suite doesn't expect the prompt to be there and I can easily fix that when I replace `assert out[-3] == "wrote pydoc_test.html", raw` with `assert "wrote pydoc_test.html" in out[-3], raw` though I am not sure whether I can do that.

**Environment**
- OS: Oracle Solaris 11.4
- virtualenv 20.16.3

kulikjak:
This fixes the issue with **test_csh** when running with the original `csh` rather than today widely used `tcsh`.

`tcsh` on Linux (`csh` is just a symlink to `tcsh`) or Solaris:
```shell
$ echo "echo 'test'" | tcsh -i
[me@mymachine ~]$ test
[me@mymachine ~]$ exit
$
```
and `csh` on Solaris:
```shell
$ echo "echo 'test'" | csh -i
mymachine-i386-25280% test
mymachine-i386-25280% $
```
By adding a forced newline to the prompt, we can be sure that at least one will be there, and thus the prompt won't 'leak' to the next line.

I am wondering whether it makes sense to add a similar **test_tcsh** test that would explicitly run `tcsh`. On Linux, it would do the same thing, but on platforms where both versions of the shell are available, it would test both.

Closes #2409 

### DIFF
--- a/docs/changelog/2418.bugfix.rst
+++ b/docs/changelog/2418.bugfix.rst
@@ -1,0 +1,1 @@
+Allow the test suite to pass even with the original C shell (rather than ``tcsh``) - by :user:`kulikjak`.
--- a/tests/unit/activation/test_csh.py
+++ b/tests/unit/activation/test_csh.py
@@ -7,6 +7,8 @@ def test_csh(activation_tester_class, activation_tester):
             super().__init__(CShellActivator, session, "csh", "activate.csh", "csh")
 
         def print_prompt(self):
-            return "echo 'source \"$VIRTUAL_ENV/bin/activate.csh\"; echo $prompt' | csh -i"
+            # Original csh doesn't print the last newline,
+            # breaking the test; hence the trailing echo.
+            return "echo 'source \"$VIRTUAL_ENV/bin/activate.csh\"; echo $prompt' | csh -i ; echo"
 
     activation_tester(Csh)


gaborbernat (changes requested):
You'll need to fix the CI.

kulikjak:
The one failure again doesn't seem to be related to my change.

gaborbernat (changes requested):
Can you add a changelog entry?

kulikjak:
Sure, done.

PR #2422, opened 2022-09-17
Change Nushell activation script to module

kubouch:
### Description

The activation script will be a module that would be added as an overlay which allows for more ergonomic handling of the environment varaibles. 

I had to change the `source` activation command to `overlay use` because that's how you activate overlay in Nushell. However, it required me to also disable quoting of the command. By default, the Python activator would create `'overlay use'` and try to call that as a command but in Nushell, commands are called without quotes. 


tests/unit/activation/test_nushell.py
```diff
@@ -13,9 +13,16 @@ def __init__(self, session):
 
             super().__init__(NushellActivator, session, cmd, "activate.nu", "nu")
 
+            self.activate_cmd = "overlay use"
```
gaborbernat, line 16: Shouldn't we make the activate command a list instead?

kubouch: I tried, but it would then try to run something like `'["overlay", "use"]' ...` literally spelling out the list as a string.

gaborbernat: I meant to rework the base class to allow the use of composed commands and change from single elements to list for the shells. Other shells can use a list of 1 length stuff.

kubouch: OK, as you want. You can do that before merging this PR and I can change it to be a list. This PR is meant for the next release of Nushell 27.9. but the old `source` method will still keep working after the release so there is no need to rush this PR.

gaborbernat: I don't have currently plan or bandwidth to do this myself, so I'm afraid as your feature needs this functionality you'll need to do it yourself here.

kubouch: Is it a problem to accept the current approach? It seems to be working correctly. The quoting change affects only the Nushell activator that affects only Nushell.

gaborbernat: While this approach works I'd prefer my proposal for ease of maintenance which is yes important for me.

kubouch: OK then, I'll see what I can do.

gaborbernat:
@kubouch shall we proceed here?

kubouch:
Yes, this is still intended to be merged, I just didn't find the time to finish the activation command being a list.

gaborbernat (approved):

PR #2424, opened 2022-09-25
Change activation call to be a list of commands

kubouch:
ActivationTester will use a list of strings as the command instead of just a string.

Required to proceed with https://github.com/pypa/virtualenv/pull/2422

### DIFF
--- a/docs/changelog/2424.feature.rst
+++ b/docs/changelog/2424.feature.rst
@@ -1,0 +1,1 @@
+Change shell test activation command to be a list of commands.
--- a/tests/unit/activation/conftest.py
+++ b/tests/unit/activation/conftest.py
@@ -161,7 +161,7 @@ class ActivationTester:
         return NotImplemented
 
     def activate_call(self, script):
-        cmd = " ".join([str(cmd) for cmd in self.activate_cmd])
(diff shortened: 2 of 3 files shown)


kubouch:
@gaborbernat Is something like this what you had in mind?

kubouch:
@gaborbernat Could you please comment whether this approach is what you had in mind when making the activation command a list of strings? Should I continue debugging it to make it work on Windows or is it something else you meant? Reference conversation: https://github.com/pypa/virtualenv/pull/2422#discussion_r973625771

gaborbernat:
Yeah you are on the right path.

gaborbernat:
@kubouch updates on this?

kubouch:
I didn't have time to investigate the Windows issue because I had to focus on my actual work but I'd like this to be finished. If you have an idea why it's failing on Windows, that would help a lot.

gaborbernat:
Not sure sadly.

gaborbernat:
Seems no longer needed.

kubouch:
OK, so the activation commands being a list is no longer necessary?

gaborbernat:
Would still be nice to make the change, but seems we can't get it working 👍

kubouch:
Yeah, it's something related to the Windows paths but I couldn't locate the issue.

Issue #2427, opened 2022-10-05
GraalPython support


steve-s:
GraalPython (https://github.com/oracle/graalpython) is an alternative Python 3 implementation. As it becomes more mature, it is useful to better integrate it into the Python ecosystem. Virtualenv plugin for GraalPy would be an important step, because it, among other things, enables testing on GraalPython via tox.

We would like to contribute GraalPy Creator plugin (based on `ViaGlobalRefVirtualenvBuiltin`) to virtualenv core.

One issue that GraalPython is facing are changes in pip and setuptools that break those packages on GraalPython. This is a chicken-egg problem. We need better integration in Python tools in order to gain wider recognition and to eventually enable continuous testing of some of those packages on GraalPython. Without that the GraalPy virtualenv plugin may become unusable when pip/setuptools release new version. There are several options how to approach this:

* ignore: users of the GraalPython plugin would have to know to pass extra options to virtualenv to pin the versions of pip/setuptools
* new Seeder "ensurepip", which would use target Python's `ensurepip` module to seed the new virtual environment. GraalPython's `ensurepip` takes care of installing the right versions. Users of the GraalPython plugin would have to know to use this seeder.
* building on the previous option: set the "ensurepip" seeder as the default when GraalPython is detected -- GraalPython plugin would work out of the box without any extra options

What do the maintainers of virtualenv think about this issue and the possible solutions?

**Alternative Solutions**

Alternatively, the GraalPython plugin can be external. There are some disadvantages:

* testing: if the plugin is in virtualenv core, it will be tested in virtualenv CI, issues will be detected earlier
* APIs: we would need to iron out what is virtualenv public API that the plugin can use and rely on being stable
   * if classes such as `ViaGlobalRefVirtualenvBuiltin` are not exposed, it seems that that would mean significant code duplication in the external plugin
* we cannot do the suggested change of default seeder for GraalPy unless virtualenv exposes some new hook for that
* we need to capture one GraalPy specific property in `py_info.py`
* ease of use for the end users: it is not a major obstacle, but included out of the box is always better experience than having to install a separate plugin

Another alternative is to use the Venv creator for GraalPython. The issue is that the GraalPython implementation is not recognized by virtualenv and at that point it refuses to continue to invoke the `venv` module. There does not seem to be any way to add a "recognized implementation" without implementing a Creator service for it.

gaborbernat:
[...]

PR #2428, opened 2022-10-05
Add GraalPython support

steve-s:
Adds Posix creator for GraalPython and `EnsurePipSeeder`, which is used as the default if GraalPython is detected. Using the `ensurepip` module is preferred on GraalPython, because:

* new releases of pip and setuptools tend to break on GraalPython
* GraalPython patches compatibility issues in sources of some well known packages installed via pip. This is done by monkey patching pip internals and GraalPy needs predictable pip version for this to work.

The addition of `EnsurePipSeeder` and its default usage if GraalPython is detected are up for a discussion. More details are in the issue #2427. I will update the documentation once this open question is settled.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
 Copyright (c) 2020-202x The virtualenv developers
+Copyright (c) 2022, Oracle and/or its affiliates
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
--- a/docs/changelog/2427.feature.rst
+++ b/docs/changelog/2427.feature.rst
@@ -1,0 +1,1 @@
+Add builtin support for GraalPython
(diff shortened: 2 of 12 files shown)


gaborbernat:
Hey, as said in the issue I'd like this to first get battle tested as a plugin. If as a plugin looks stable for a few months I'm happy to accept it.

LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2020-202x The virtualenv developers
+Copyright (c) 2022, Oracle and/or its affiliates
```
gaborbernat, line 2: You can't change the license, sorry.

gaborbernat (changes requested):

LICENSE
steve-s: Yes, this is meant to be just a copyright, no license change. Related discussion: https://github.com/pyenv/pyenv/pull/1594#discussion_r414980992. In any case, if this is going to be a separate plugin we can defer this to later time

gaborbernat: You can't change the copyright either.

PR #2442, opened 2022-11-07
Try alternate filenames for system_executable

vfazio:
The value of `sys._base_executable` may not be a real file due to changes made in CPython 3.11. The value is derived from the current executable name and the "home" key from pyvenv.cfg.

On POSIX systems, virtual environments deploy "python" for use within the venv however CPython's `make install` and a number of distributions do not provide a system "python" in part because of PEP 394.

Virtualenv exposes this via `PythonInfo.system_executable` and can encounter issues when attempting to execute a non-existent file.

Attempt to fallback to "python<MAJOR>" and "python<MAJOR>.<MINOR>" if "python" does not exist.

Signed-off-by: Vincent Fazio <vfazio@xes-inc.com>


vfazio:
I'll add a news entry later but wanted to get this up to get feedback

gaborbernat:
test too please

vfazio:
> test too please

I haven't quite figured out how i'm going to be able to test this.

tests/unit/discovery/py_info/test_py_info.py
```diff
@@ -378,6 +378,29 @@ def test_custom_venv_install_scheme_is_prefered(mocker):
...
 
+@pytest.mark.skipif(not (os.name == "posix" and sys.version_info[:2] >= (3, 11)), reason="POSIX 3.11+ specific")
+def test_fallback_existent_system_executable(mocker):
+    current = PythonInfo()
+    # Posix may execute a "python" out of a venv but try to set the base_executable
+    # to "python" out of the system installation path. PEP 394 informs distributions
+    # that "python" is not required and the standard `make install` does not provide one
+
+    # Falsify some data to look like we're in a venv
+    current.prefix = current.exec_prefix = "/tmp/tmp.izZNCyINRj/venv"
```
gaborbernat, line 389: will this test fail if upstream accepts your PR?

vfazio: It shouldn't. I mock `sys. _base_executable` to return a nonexistent file to mimic the "broken" behavior so it should continue to work as expected even if the value `sys. _base_executable` would return is correct. Once upstream decides on the PR and whether to backport it, we can change the max version to which this quirk applies.

src/virtualenv/discovery/py_info.py
```diff
@@ -138,6 +138,21 @@ def _fast_get_system_executable(self):
                 base_executable = getattr(sys, "_base_executable", None)  # some platforms may set this to help us
                 if base_executable is not None:  # use the saved system executable if present
                     if sys.executable != base_executable:  # we know we're in a virtual environment, cannot be us
+                        # Do light validation for non-Windows virtual environments
+                        # Some installs/distributions do not provide a version-less "python" binary (PEP 394).
+                        # Try to fallback to an alternative based on version number
+                        # This should still be relatively fast since it's just a couple of `stat` calls
+                        if self.os == "posix" and (self.version_info.major, self.version_info.minor) >= (3, 11):
+                            if not os.path.exists(base_executable):
```
vfazio, line 146: One thing i'm not sure of is if we should ever return a file that doesn't actually exist. Is virtualenv better off returning `None` for `PythonInfo.system_executable`? if so, i'd probably reorder these lines

gaborbernat: Yes if the file does not exist we should return nothing.

vfazio: Addressed this in the latest version

src/virtualenv/discovery/py_info.py
```diff
@@ -138,7 +138,25 @@ def _fast_get_system_executable(self):
...
+                        # the system install location (see PEP 394) so try to fallback to a versioned binary.
+                        #
+                        # Gate this to Python 3.11 as `sys._base_executable` path resolution is now relative to
+                        # the 'home' key from pyvenv.cfg which often points to the system install location.
+                        major, minor = self.version_info.major, self.version_info.minor
+                        if self.os == "posix" and (major, minor) >= (3, 11):
+                            # search relative to the directory of sys._base_executable
+                            base_dir = os.path.dirname(base_executable)
+                            for base_executable in [
+                                os.path.join(base_dir, "{}".format(exe))
```
gaborbernat, line 154: exe is already a string we don't need this format, no?

src/virtualenv/discovery/py_info.py
```diff
@@ -138,7 +138,25 @@ def _fast_get_system_executable(self):
...
+                                for exe in ("python{}".format(major), "python{}.{}".format(major, minor))
+                            ]:
+                                if os.path.exists(base_executable):
+                                    return base_executable
+                        return None  # File doesn't exist and no alternate was available
```
gaborbernat, line 159: I don't think we need this line as would then go to next line that already does this, no?

src/virtualenv/discovery/py_info.py
vfazio: @gaborbernat you're absolutely correct and this is a good example of why i shouldn't be making commits late at night.

Do you prefer subsequent commits to address comments or can i amend and force push this branch?

vfazio:
The ubuntu pipelines are also failing here due to `nushell`

vfazio:
> > The ubuntu pipelines are also failing here due to `nushell`
> 
> Reached out to them separately, hopefully they fix it soon.

@gaborbernat 
It looks like as part of 0.71 they're now including a subdirectory in their releases archives which is actually pretty typical for most releases i've seen from other projects. 0.70 did not have a subdirectory

```
vfazio@vfazio2 ~/Downloads :( $ tar -tvf nu-0.70.0-x86_64-unknown-linux-gnu.tar.gz
-rw-r--r-- runner/docker  1094 2022-10-18 14:11 LICENSE
-rw-r--r-- runner/docker   124 2022-10-18 14:11 README.txt
-rwxr-xr-x runner/docker 70621816 2022-10-18 14:11 nu
-rwxr-xr-x runner/docker  2752208 2022-10-18 14:11 nu_plugin_custom_values
-rwxr-xr-x runner/docker  2699672 2022-10-18 14:11 nu_plugin_example
-rwxr-xr-x runner/docker  6006152 2022-10-18 14:11 nu_plugin_gstat
-rwxr-xr-x runner/docker  2130760 2022-10-18 14:11 nu_plugin_inc
-rwxr-xr-x runner/docker  4950368 2022-10-18 14:11 nu_plugin_query

vfazio@vfazio2 ~/Downloads $ tar -tvf nu-0.71.0-x86_64-unknown-linux-gnu.tar.gz 
drwxr-xr-x runner/docker     0 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/
-rw-r--r-- runner/docker  1094 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/LICENSE
-rwxr-xr-x runner/docker 2050928 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu_plugin_inc
-rw-r--r-- runner/docker     124 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/README.txt
-rwxr-xr-x runner/docker 65975456 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu
-rwxr-xr-x runner/docker  2655216 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu_plugin_example
-rwxr-xr-x runner/docker  4827000 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu_plugin_query
-rwxr-xr-x runner/docker  2708240 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu_plugin_custom_values
-rwxr-xr-x runner/docker  5961760 2022-11-08 12:11 nu-0.71.0-x86_64-unknown-linux-gnu/nu_plugin_gstat

```

So i think the options are:
* update the pipeline to pull 0.70 to abide by the old release archive structure
* modify the current pipeline script to use `--strip-components=1` here https://github.com/pypa/virtualenv/blob/main/.github/workflows/check.yml#L48

src/virtualenv/discovery/py_info.py
vfazio: fixed in the latest push

```diff
diff --git a/src/virtualenv/discovery/py_info.py b/src/virtualenv/discovery/py_info.py
index a2eb633..2a39e50 100644
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -151,12 +151,11 @@ class PythonInfo(object):
                             # search relative to the directory of sys._base_executable
                             base_dir = os.path.dirname(base_executable)
                             for base_executable in [
-                                os.path.join(base_dir, "{}".format(exe))
+                                os.path.join(base_dir, exe)
                                 for exe in ("python{}".format(major), "python{}.{}".format(major, minor))
                             ]:
                                 if os.path.exists(base_executable):
                                     return base_executable
-                        return None  # File doesn't exist and no alternate was available
             return None  # in this case we just can't tell easily without poking around FS and calling them, bail
         # if we're not in a virtual environment, this is already a system python, so return the original executable
         # note we must choose the original and not the pure executable as shim scripts might throw us off
```

src/virtualenv/discovery/py_info.py
vfazio: fixed in the latest push

```diff
diff --git a/src/virtualenv/discovery/py_info.py b/src/virtualenv/discovery/py_info.py
index a2eb633..2a39e50 100644
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -151,12 +151,11 @@ class PythonInfo(object):
                             # search relative to the directory of sys._base_executable
                             base_dir = os.path.dirname(base_executable)
                             for base_executable in [
-                                os.path.join(base_dir, "{}".format(exe))
+                                os.path.join(base_dir, exe)
                                 for exe in ("python{}".format(major), "python{}.{}".format(major, minor))
                             ]:
                                 if os.path.exists(base_executable):
                                     return base_executable
-                        return None  # File doesn't exist and no alternate was available
             return None  # in this case we just can't tell easily without poking around FS and calling them, bail
         # if we're not in a virtual environment, this is already a system python, so return the original executable
         # note we must choose the original and not the pure executable as shim scripts might throw us off
```

vfazio:
Surprisingly,  my cpython fix got merged and was backported into 3.11 so this likely won't be as big of an issue in starting in 3.11.1 and 3.12. 

It's still worthwhile to perform this check because cpython will still return a base executable that may not exist. The primary fix for that would be #2441 to ensure it searches in the right path. The code introduced here still only applies to 3.11 going forward and based on our discussion should still return `None` if the file does not exist.

vfazio:
Surprisingly,  my cpython fix got merged and was backported into 3.11 so this likely won't be as big of an issue in starting in 3.11.1 and 3.12. 

It's still worthwhile to perform this check because cpython will still return a base executable that may not exist. The primary fix for that would be #2441 to ensure it searches in the right path. The code introduced here still only applies to 3.11 going forward and based on our discussion should still return `None` if the file does not exist.

PR #2444, opened 2022-11-10
Try to fix Nushell install

kubouch:
Because https://github.com/pypa/virtualenv/commit/e4a42c6eeb48004461bb8711fa06e830bc40eb37 didn't seem to solve the issue, this is an alternative solution.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -45,8 +45,8 @@ jobs:
             ${{ runner.os == 'Linux' && 'sudo apt-get install curl wget -y' || true }} && \
             ${{ runner.os == 'Linux' && 'nushell_url=$(curl -s https://api.github.com/repos/nushell/nushell/releases/latest | grep "browser_" | grep "x86_64" | grep "linux" | grep "gnu" | cut -d\" -f4 )' || true }} && \
             ${{ runner.os == 'Linux' && 'wget -O nushell.tar.gz $nushell_url' || true }} && \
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -45,8 +45,8 @@ jobs:
             ${{ runner.os == 'Linux' && 'sudo apt-get install curl wget -y' || true }} && \
             ${{ runner.os == 'Linux' && 'nushell_url=$(curl -s https://api.github.com/repos/nushell/nushell/releases/latest | grep "browser_" | grep "x86_64" | grep "linux" | grep "gnu" | cut -d\" -f4 )' || true }} && \
             ${{ runner.os == 'Linux' && 'wget -O nushell.tar.gz $nushell_url' || true }} && \
-            ${{ runner.os == 'Linux' && 'tar -zxf --strip-components=1 nushell.tar.gz' || true }} && \
```
vfazio, line 48: Seems easier to just make this 
```
tar -zxf nushell.tar.gz --strip-components=1
```
instead of trying to glob the file path

kubouch: That was tried before but didn't work

vfazio: There's a very important order of arguments. GNU Tar requires a filename immediately after `-f`. Currently `--strip-components=1` is being interpreted as a filename and is invalid. it needs to be placed after the filename or before the other arguments.

vfazio: The problem with pattern matching is that if the triplet changes, this will break again. using `--strip-components` obviates that.

just my $.02

PR #2458, opened 2022-12-04
Respect `VIRTUAL_ENV_DISABLE_PROMPT` in nushell activation script

m-lima:
The `activate.nu` script was not respecting the `VIRTUAL_ENV_DISABLE_PROMPT` like other scripts. This PR addresses that.

The changes can be summarized as:
- Only add the prompt-related env vars if `$env.VIRTUAL_ENV_DISABLE_PROMPT` evaluates to `false`
- Like with other activation scripts, it will evaluate to false if the env var is missing, it is empty, it is '0'. Also, because of nushell type system and the use of the `into bool` built-in, anything that nushell can covert into a `false` will also work, e.g. the strings `false` and `0000` or `0.0`.

There are no changes to the documentation since this is already an expected behavior.
Also, I did not add an entry to the changelog as the developer documentation states that "non trivial" changes should be added. I considered this change trivial, but please let me know if an entry is required.

Fixes: #2461

### DIFF
--- a/src/virtualenv/activation/nushell/activate.nu
+++ b/src/virtualenv/activation/nushell/activate.nu
@@ -101,7 +101,7 @@ export-env {
           { $'($virtual_prompt)' }
       }
 
-      $new_env | insert {
+      $new_env | merge {
         _OLD_VIRTUAL_PATH   : ($old_path | str collect $path_sep)
         _OLD_PROMPT_COMMAND : $old_prompt_command
         PROMPT_COMMAND      : $new_prompt
--- a/src/virtualenv/activation/nushell/activate.nu
+++ b/src/virtualenv/activation/nushell/activate.nu
@@ -15,6 +15,21 @@ export-env {
         $name in (env).name
     }
 
+    # Emulates a `test -z`, but btter as it handles e.g 'false'
+    def is-env-true [name: string] {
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
This change is 52 new lines, and 28 removals, this is not trivial so let's add the changelog.

m-lima:
Interesting error. Looks like libc was not found to execute nushell? Or maybe nushell requires a specific version of glibc that is not present in the ubuntu vms running the test?
Or am I reading this wrong?

src/virtualenv/activation/nushell/activate.nu
```diff
@@ -49,44 +64,53 @@ export-env {
...
+      $new_env
     } else {
-        if (has-env 'PROMPT_COMMAND') {
-            $env.PROMPT_COMMAND
-        } else {
-            ''
-        }
-    }
+      # Creating the new prompt for the session
+      let virtual_prompt = if ('__VIRTUAL_PROMPT__' == '') {
```
flying-sheep, line 76: `'__VIRTUAL_PROMPT__' == ''` seems like it has an extra pair of `'`s.

Should probably be `__VIRTUAL_PROMPT__ == ''`

PR #2475, opened 2022-12-29
Fix Nushell installation failures

kubouch:
This changes the Nushell installation method to use https://github.com/marketplace/actions/install-development-tools which in turn uses [cargo binstall](https://github.com/cargo-bins/cargo-binstall) to install Nushell binary directly from GitHub releases. Unlike the previous method which points at https://api.github.com/repos/nushell/nushell/releases/latest this one should point at the build artifacts directly and hopefully avoids the 403 problems.

Also makes the CI script simpler.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -36,18 +36,20 @@ jobs:
           - { os: macos-12, py: "brew@3.9" }
           - { os: macos-12, py: "brew@3.8" }
     steps:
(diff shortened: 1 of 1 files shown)


kubouch:
There is a failure on MacOS but it seems unrelated to the changes I made.

kubouch:
OK, I'm falling back to using choco and brew on Windows & Mac so this PR affects only the Linux build.

gaborbernat (approved):

PR #2477, opened 2022-12-30
Fix coverage reporting

gaborbernat:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
 ]
 optional-dependencies.test = [
   "covdefaults>=2.2.2",
-  "coverage>=7.0.1",
+  "coverage>=7.0.2",
   "coverage-enable-subprocess>=1",
   "flaky>=3.7",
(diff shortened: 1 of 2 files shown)


MeCorey (commented):
# virtualenv activation module
# Activate with `overlay use activate.nu`
# Deactivate with `deactivate`, as usual
#
# To customize the overlay name, you can call `overlay use activate.nu as foo`,
# but then simply `deactivate` won't work because it is just an alias to hide
# the "activate" overlay. You'd need to call `overlay hide foo` manually.

export-env {
    def is-string [x] {
        ($x | describe) == 'string'
    }

    def has-env [name: string] {
        $name in (env).name
    }

    # Emulates a `test -z`, but btter as it handles e.g 'false'
    def is-env-true [name: string] {
      if (has-env $name) {
        # Try to parse 'true', '0', '1', and fail if not convertible
        let parsed = do -i { $env | get $name | into bool }
        if ($parsed | describe) == 'bool' {
          $parsed

PR #2478, opened 2023-01-02
Fix coverage reporting

nedbat:
This sets `relative_files=true`, and also uses a fix in coverage to make it all work.  I'll be releasing a new coverage version today, so take the parts of this you want, when you want them.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,7 @@ extras =
     test
 commands =
     coverage erase
-    coverage run -m pytest {posargs:--junitxml {toxworkdir}/junit.{envname}.xml tests --int}
+    coverage run -m pytest {posargs:--junitxml {toxworkdir}/junit.{envname}.xml tests/unit/config --int}
     coverage combine
     coverage report --skip-covered --show-missing
     coverage xml -o {toxworkdir}/coverage.{envname}.xml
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -51,8 +51,9 @@ description = [run locally after tests]: combine coverage data and create report
 skip_install = true
 deps =
     covdefaults>=2.2.2
-    coverage[toml]>=7.0.1
     diff_cover>=7.3
+    git+https://github.com/nedbat/coveragepy@nedbat/fix-virtualenv#egg=coverage==0.0
+    tomli
 extras =
```
gaborbernat, line 57: Why do we need this, isn't the `toml` extra already covering this?

nedbat: That was just for my debugging, I wasn't sure of the extra syntax with a git+ url, so I just stuck it in there like this.

nedbat: You can revert these lines. Coverage.py 7.0.2 is now out, with the fix.

gaborbernat:
Moved to https://github.com/pypa/virtualenv/pull/2478

gaborbernat:
Seems the wheel files are no longer excluded from coverage; see https://github.com/pypa/virtualenv/actions/runs/3824948711/jobs/6507540653. Do relative paths not work with `report.omit`?

Issue #2491, opened 2023-01-30
Windows registry Python discovery (PEP 514) not working for "cpython" specs for distributions by unknown orgs


faph:
**Issue**

On Windows:

```shell
$ python --version
Python 3.11.1
$ python -m virtualenv --version
virtualenv 20.17.1 from C:\<some path>\tools\python3.11\latest\Lib\site-packages\virtualenv\__init__.py
$ python -m virtualenv .venv -p cpython3.10 -vv
```

Output:

```
180 setup logging to DEBUG [DEBUG report:35]
195 find interpreter for spec PythonSpec(implementation=cpython, major=3, minor=10) [INFO builtin:56]
195 proposed PythonInfo(spec=CPython3.11.1.final.0-64, exe=C:\<some path>\tools\python3.11\latest\python.exe, platform=win32, version='3.11.1 (tags/v3.11.1:a7a450f, Dec
6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]', encoding_fs_io=utf-8-utf-8) [INFO builtin:63]
202 discover PATH[0]=C:\<some path>\tools\zip\latest\bin\ [DEBUG builtin:108]
205 filesystem is not case-sensitive [DEBUG info:24]
206 discover PATH[1]=C:\<some path>\tools\yq\latest [DEBUG builtin:108]
...
249 RuntimeError: failed to find interpreter for Builtin discover of python_spec='cpython3.10' [ERROR __main__:61]
```

Seems this is due to logic here: https://github.com/pypa/virtualenv/blob/main/src/virtualenv/discovery/windows/__init__.py#L23

It lists `PythonCore` and `ContinuumAnalytics` as the 2 **organizations** mapping to the `CPython` **implementation**. That logic is unfortunate, since many more organizations can decide to distribute CPython. For example, my firm uses a different organization name in the Windows registry, compliant with PEP 514 <https://peps.python.org/pep-0514/>.

I guess the solution should be similar to how the implementation is looked up for `PATH` OS environment variable entries: execute the Python binary and inspect the "pyinfo" output.

faph:
For the avoidance of doubt, none of my Python interpreters are available on `PATH`. 

If I **do** add the relevant interpreters to `PATH`, then the discover works as expected.

Also, if I change the Windows Registry keys, and use the `PythonCore` organization name, it also works.

faph:
[...]

PR #2504, opened 2023-02-14
[WIP] Support Windows CPython interpreters distributed by non-standard orgs

faph:
See issue #2491 

### DIFF
--- a/src/virtualenv/discovery/windows/__init__.py
+++ b/src/virtualenv/discovery/windows/__init__.py
@@ -19,15 +19,10 @@ def propose_interpreters(spec, cache_dir, env):
     )
 
     for name, major, minor, arch, exe, _ in existing:
-        # pre-filter
-        if name in ("PythonCore", "ContinuumAnalytics"):
-            name = "CPython"
-        registry_spec = PythonSpec(None, name, major, minor, None, arch, exe)
-        if registry_spec.satisfies(spec):
-            interpreter = Pep514PythonInfo.from_exe(exe, cache_dir, env=env, raise_on_error=False)
-            if interpreter is not None:
-                if interpreter.satisfies(spec, impl_must_match=True):
-                    yield interpreter
+        interpreter = Pep514PythonInfo.from_exe(exe, cache_dir, env=env, raise_on_error=False)
+        if interpreter is not None:
+            if interpreter.satisfies(spec, impl_must_match=True):
+                yield interpreter
 
 
 __all__ = [


faph:
Testing here a straightforward fix whereby the Windows Registry-based pre-filtering is removed altogether. I assume this pre-filtering is there for performance reasons?

The issue with the Window Registry-based pre-filtering is that implementation meta data is not available. Previously a hack was implemented to map organizations "PythonCore" and "Continuum" to the "CPython" implementation. That obviously does not work for anyone else!

In any case, the existing logic applied full matching based on meta-data provided by the interpreter executable itself. That logic is cached in any case, so normally we expect this to be evaluated once per interpreter. That seems to suggest the Windows Registry-based optimization is largely redundant?

This would also bring it in line with the Linux implementation which simply relies on interpreter paths on the `PATH` OS environment variable. Again, that would delegate the spec matching fully to the interpreter executable itself.

faph:
Local test runs seem to suggest existing tests are fully passing with this change.

@gaborbernat Are you able to approve the GH Actions workflow to confirm the tests are green?

gaborbernat (commented):
Test and changelog needs to be added.

faph:
Understood - have some tests to replicate the issue that I need to tidy up first.

Just looking for guidance whether this is at all a solution you're willing to consider @gaborbernat

src/virtualenv/discovery/windows/__init__.py
```diff
@@ -19,15 +18,10 @@ def propose_interpreters(spec, cache_dir, env):
     )
 
     for name, major, minor, arch, exe, _ in existing:
-        # pre-filter
-        if name in ("PythonCore", "ContinuumAnalytics"):
```
gaborbernat, line 23: For these two can we still match without the from exe and only for other do what you're doing now?

faph: Yeah we could. I was wondering whether it's worth it to make this exceptional logic for these 2 orgs? What would the additional complexity achieve?

gaborbernat: It is worth it. They are the most popular and will be used 99% of the time.

faph: OK. So the logic would be:

1. if name in ("PythonCore", "ContinuumAnalytics"), implementation becomes "CPython"
2. if implementation == "CPython", apply registry-based pre-filtering
3. in other cases, do not apply pre-filtering at all

gaborbernat: Yes.

src/virtualenv/discovery/windows/__init__.py
```diff
@@ -19,13 +26,17 @@ def propose_interpreters(spec, cache_dir, env):
...
-        registry_spec = PythonSpec(None, name, major, minor, None, arch, exe)
-        if registry_spec.satisfies(spec):
+        # Map well-known/most common organizations to a Python implementation, use the org name as a fallback for
+        # backwards compatibility.
+        implementation = _IMPLEMENTATION_BY_ORG.get(name, name)
+
+        # Pre-filtering based on Windows Registry metadata, for CPython only
+        skip_pre_filter = implementation != "CPython"
+        registry_spec = PythonSpec(None, implementation, major, minor, None, arch, exe)
+        if skip_pre_filter or registry_spec.satisfies(spec):
```
faph, line 35: Alternatively, we add a `impl_must_match` argument to this method and set it to `False` in this case instead of skipping pre-filtering for non-CPython. But that would be a bigger change.

faph:
For info, I am still trying to find a good testing strategy that covers the `Pep514PythonInfo.from_exe(...)` logic.

src/virtualenv/discovery/windows/__init__.py
```diff
@@ -19,13 +26,17 @@ def propose_interpreters(spec, cache_dir, env):
...
-        if name in ("PythonCore", "ContinuumAnalytics"):
-            name = "CPython"
-        registry_spec = PythonSpec(None, name, major, minor, None, arch, exe)
-        if registry_spec.satisfies(spec):
+        # Map well-known/most common organizations to a Python implementation, use the org name as a fallback for
+        # backwards compatibility.
+        implementation = _IMPLEMENTATION_BY_ORG.get(name, name)
+
+        # Pre-filtering based on Windows Registry metadata, for CPython only
+        skip_pre_filter = implementation != "CPython"
```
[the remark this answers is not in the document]
faph: `PythonSpec.satisfies()` uses case-insensitive matching actually.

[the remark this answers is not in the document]
faph: Changed the implementation to match that.

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+            return result
+        except Exception as exception:
+            result = exception
+            raise result
+        finally:
+            enum_collect[key_id].append(result)
+
+    value_collect = defaultdict(dict)
+
+    def _v(key, value_name):
```
gaborbernat, line 110: a meaningful name for this method please

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    mocker.patch.object(winreg, "QueryValueEx", side_effect=_v)
+    mocker.patch.object(winreg, "OpenKeyEx", side_effect=_c)
+
+    yield
+
+    print("")
+    print(f"hive_open = {hive_open}")
+    print(f"key_open = {dict(key_open.items())}")
+    print(f"value_collect = {dict(value_collect.items())}")
+    print(f"enum_collect = {dict(enum_collect.items())}")
```
gaborbernat, line 132: Remove prints please, I think were forgotten here 🤔

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    # noinspection PyUnresolvedReferences
+    from winreg import EnumKey, OpenKeyEx, QueryValueEx
+
+    from virtualenv.discovery.windows.pep514 import winreg
+
+    hive_open = {}
+    key_open = defaultdict(dict)
+
+    @contextmanager
+    def _c(*args):
```
gaborbernat, line 78: a meaningful name for this method please

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+            raise exception
+        finally:
+            if len(args) == 4:
+                hive_open[args] = res
+            elif len(args) == 2:
+                key_open[key_id][args[1]] = res
+
+    enum_collect = defaultdict(list)
+
+    def _e(key, at):
```
gaborbernat, line 96: a meaningful name for this method please

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+            raise value
+        yield result
+
+    mocker.patch.object(winreg, "OpenKeyEx", side_effect=_o)
+    mocker.patch("os.path.exists", return_value=True)
+
+
+@pytest.fixture()
+def _collect_winreg_access(mocker):
+    # noinspection PyUnresolvedReferences
```
gaborbernat, line 69: remove this pycharm specific comment

tests/unit/discovery/windows/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+        ("ContinuumAnalytics", 3, 7, 64, "C:\\Users\\user\\Miniconda3-64\\python.exe", None),
+        ("PythonCore", 3, 6, 64, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python36\\python.exe", None),
+        ("PythonCore", 3, 6, 64, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python36\\python.exe", None),
+        ("PythonCore", 3, 5, 64, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python35\\python.exe", None),
+        ("PythonCore", 3, 6, 64, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python36\\python.exe", None),
+        ("PythonCore", 3, 7, 32, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python37-32\\python.exe", None),
+        ("PythonCore", 3, 9, 64, "C:\\Users\\user\\AppData\\Local\\Programs\\Python\\Python36\\python.exe", None),
+        ("PythonCore", 2, 7, 64, "C:\\Python27\\python.exe", None),
+        ("PythonCore", 3, 4, 64, "C:\\Python34\\python.exe", None),
+        ("CompanyA", 3, 3, 64, "Z:\\CompanyA\\Python\\3.3\\python.exe", None),
```
gaborbernat, line 165: Can we bump here the version numbers to more recent releases while we're here.

tests/unit/discovery/windows/test_windows.py
```diff
@@ -0,0 +1,33 @@
+import sys
+
+import pytest
+
+from virtualenv.discovery.py_spec import PythonSpec
+
+
+@pytest.mark.skipif(sys.platform != "win32", reason="no Windows registry")
```
gaborbernat, line 8: Why we need to skip if we're going to mock the registry?

gaborbernat (commented):
Loooks ok in general 👍

tests/unit/discovery/windows/conftest.py
faph: Done

tests/unit/discovery/windows/conftest.py
faph: This entire fixture function is unused, as far as I can see. Function removed.

tests/unit/discovery/windows/conftest.py
faph: Done

tests/unit/discovery/windows/conftest.py
faph: Done

tests/unit/discovery/windows/conftest.py
faph: Entire function removed.

tests/unit/discovery/windows/conftest.py
faph: Sure. It will obviously require quite a few changes in the test asserts which makes me a little nervous given we're also making changes to the code subject to test. 

Let me do it as one of the last commits.

tests/unit/discovery/windows/test_windows.py
faph: Because before the mocks get applied, we will need to import `winreg` first (through `virtualenv.discover.windows.pep514`). This import itself will fail on non-Windows.

tests/unit/discovery/windows/conftest.py
faph: Done

faph: Done

faph:
@gaborbernat I think that's all feedback addressed. Thanks!

faph:
Appreciate your status atm @gaborbernat. Any indication when we might be able to merge and release this?

MeCorey:
beautiful!

PR #2548, opened 2023-04-19
Drop Python 2 support

gaborbernat:


EKTehtris:
Hi would have been better to release 21.X.X for this as this is a breaking change, but thx for mentionning it in the changelog

encukou:
FWIW, there are still environments that need to support a rather wide range of Python versions. To use virtualenv (and tox) with all of 3.6-3.12, we'll need to patch and diverge. Hopefully we don't mess up too much.

gaborbernat:
Just pin to an older version of virtualenv. No need to patch.

zzzeek:
just agreeing with @EKTehtris that if virtualenv is trying to do semver, this really should have been 21.0.0.    "version 20.22.0" seems pretty "semver" already (i.e. it's not like "20" or "21" is that special of a number).

as far as realistic downstream impact, seeing a major number jump would have made it more obvious why tox stopped working for our CI builds that are still running py2.

gaborbernat:
I understand what you're saying but if every library would bump a major version whenever they are dropping support for a very old and long time not supported interpreter numbers would get very high. tox would have been just as much broken with a new  or minor. Python 2 is end of life for 2 and 1/2 years. So if you still have CI running for it, those projects are probably doomed no matter what.

Issue #1873, opened 2020-06-24
No module named 'virtualenv.seed.via_app_data'


make-ing:
**Issue**

On calling ``virtualenv test`` i expect it to create a folder named test with all the virtual environment files in there. The observed behaviour is just the following message with no folder test created:
``ModuleNotFoundError: No module named 'virtualenv.seed.via_app_data'``

**Environment**

Provide at least:
- OS: Ubuntu 20.04
- ``pip list`` of the host python where ``virtualenv`` is installed:

```console
Package                Version
---------------------- -------------
appdirs                1.4.4
apturl                 0.5.2
awscli                 1.17.14
blinker                1.4
botocore               1.14.14
Brlapi                 0.7.0
certifi                2019.11.28
chardet                3.0.4
chrome-gnome-shell     0.0.0
colorama               0.4.3
command-not-found      0.3
cryptography           2.8
cupshelpers            1.0
dbus-python            1.2.16
defer                  1.0.6
distlib                0.3.0
distro                 1.4.0
distro-info            0.23ubuntu1
docutils               0.16
entrypoints            0.3
filelock               3.0.12
httpie                 1.0.3
httplib2               0.14.0
idna                   2.8
importlib-metadata     1.5.0
jmespath               0.9.4
keyring                18.0.1
language-selector      0.1
launchpadlib           1.10.13
lazr.restfulclient     0.14.2
lazr.uri               1.0.3
louis                  3.12.0
macaroonbakery         1.3.1
more-itertools         4.2.0
netifaces              0.10.4
oauthlib               3.1.0
olefile                0.46
pexpect                4.6.0
Pillow                 7.0.0
pip                    20.1.1
protobuf               3.6.1
pyasn1                 0.4.2
pycairo                1.16.2
pycups                 1.9.73
Pygments               2.6.1
PyGObject              3.36.0
PyJWT                  1.7.1
pymacaroons            0.13.0
PyNaCl                 1.3.0
pyRFC3339              1.1
python-apt             2.0.0
python-dateutil        2.7.3
python-debian          0.1.36ubuntu1
pytz                   2019.3
pyxdg                  0.26
PyYAML                 5.3.1
reportlab              3.5.34
requests               2.22.0
requests-unixsocket    0.2.0
roman                  2.0.0
rsa                    4.0
s3transfer             0.3.3
SecretStorage          2.3.1
setuptools             45.2.0
simplejson             3.16.0
six                    1.14.0
systemd-python         234
[...]

PR #2555, opened 2023-04-22
[docs] Added note how to install tox on Debian

chrysle:

Just stumbled over #1873 again and thought it would be helpful to add a note about that to the docs.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -31,6 +31,10 @@ virtualenv is a command line application written in Python. To work on it, you'l
 - :pypi:`tox`: to automatically get the projects development dependencies and run the test suite. We recommend
   installing it using `pipx <https://pipxproject.github.io/pipx/>`_.
 
+.. note::
+
+    Debian/Ubuntu users are required to install :pypi:`tox` using the alternative installation methods and not the version provided by the package manager. Otherwise ``tox`` will break. See `this issue <https://github.com/pypa/virtualenv/issues/1873>`_ for more information.
+
 Running from source tree
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/changelog/2555.doc.rst
+++ b/docs/changelog/2555.doc.rst
@@ -1,0 +1,2 @@
+Add note how to install ``tox`` on Debian/Ubuntu systems to development guide - by :user:`chrysle`.
+
--- a/docs/changelog/2555.doc.rst
+++ b/docs/changelog/2555.doc.rst
@@ -1,2 +1,1 @@
 Add note how to install ``tox`` on Debian/Ubuntu systems to development guide - by :user:`chrysle`.
-


docs/development.rst
```diff
@@ -31,6 +31,10 @@ virtualenv is a command line application written in Python. To work on it, you'l
 - :pypi:`tox`: to automatically get the projects development dependencies and run the test suite. We recommend
   installing it using `pipx <https://pipxproject.github.io/pipx/>`_.
 
+.. note::
+
```
gaborbernat, line 35: I don't think this is the case and we already voted against adding os special installation. pipx should work

gaborbernat (changes requested):

docs/development.rst
chrysle: My intention was to warn Debian users **not** to use the os specific installation but to use `pip` and `pipx` (or zipapp and other). I mean that with _alternative installation methods_.

I had installed the debian package `tox` (which installs `python3-virtualenv` as a dependency) and tried to invoke the development environment (`tox -e dev`) with the following traceback: 

```
.tox create: /home/user/repos/virtualenv/.tox/.tox
ERROR: invocation failed (exit code 1), logfile: /home/user/repos/virtualenv/.tox/.tox/log/.tox-0.log
================================================================================= log start ==================================================================================
ModuleNotFoundError: No module named 'virtualenv.activation.xonsh'

================================================================================== log end ===================================================================================
ERROR: InvocationError for command /usr/bin/python3 -m virtualenv --no-download --python /usr/bin/python3 .tox (exited with code 1)
``` 
Don't you think a warning would be helpful if other developers on Debian run into that issue?

gaborbernat: That like could be useful. But

chrysle: > That like could be useful. But

But? Would you suggest any changes?

gaborbernat: I meant not documentation but runtime error to be better 🥰

chrysle: Hm, but while this issue is still present (and I don't think I'll be able to fix it), wouldn't it be good to spread the information? 

Maybe someone will be inspired by the note to fix it!  😉

gaborbernat: They can get inspired by reading the bug tracker? 🙂

chrysle: Well, just trying to help (and doing some persuasive work 😉).

Many developers may run into that issue and not know what to do. If they find out what cause this has in the main development docs (which surely almost everyone will read), this'll bear no problem. I personally feel reading docs more comfortable than searching for old, stale (and potentially locked) issues.

Maybe we could add the note anyway **and** transer the problem to a new issue, which we could link to?

gaborbernat: I'd rather fix the underlying issue than patch the documentation like this 🤷‍♂️

chrysle: OK, I give up. Will open an issue.

chrysle:
I have opened #2557

Issue #2560, opened 2023-04-27
pip error in virtualenv with python 3.9 and base python 3.10


szleb:
**Issue**

When creating a virtualenv with version 20.21.1 or higher with python 3.9, calling pip in the virtualenv raises the following error:

```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

The error does not appear when using python 3.10 in the virtualenv; thus, when removing the `-p py39` in the example below.

**Example**
```console

>python -m virtualenv -p py39 test
created virtual environment CPython3.9.13.final.0-64 in 2877ms
  creator Venv(dest=C:\DEV\test, clear=False, no_vcs_ignore=False, global=False, describe=CPython3Windows)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:\Users\BEL14466\AppData\Local\pypa\virtualenv)
    added seed packages: pip==23.1, setuptools==67.6.1, wheel==0.40.0
  activators BashActivator,BatchActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

>test\Scripts\activate

(test)>pip install numpy
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.

[notice] A new release of pip is available: 23.1 -> 23.1.2
[notice] To update, run: python.exe -m pip install --upgrade pip
```

**Environment**

- OS: Windows 10
- base python 3.10.6
- venv python 3.9.13
- virtuelnv >= 20.21.1

*outside virtualenv*
```console

>>> pip list
Package      Version
------------ -------
build        0.9.0
click        8.1.3
colorama     0.4.6
distlib      0.3.6
filelock     3.12.0
packaging    23.1
pep517       0.13.0
pip          23.1.2
pip-tools    6.13.0
platformdirs 3.4.0
setuptools   67.7.2
tomli        2.0.1
virtualenv   20.22.0
wheel        0.40.0 
 
```

pfmoore:
This was an issue with pip 23.1. If you upgrade to pip 23.1.2 as the message suggests, the error will be fixed. (You may need to recreate the environment as the bug means that upgrading probably won't work 🙁)

jugmac00:
@szleb you should be able to set the required packages and versions via https://tox.wiki/en/latest/config.html#requires

szleb:
I create a PR for updating the embedded wheels.

PR #2561, opened 2023-04-27
upgrade pip and setuptools (#2560)

szleb:


gaborbernat:
Replaced by https://github.com/pypa/virtualenv/pull/2562

PR #2588, opened 2023-06-13
Add ruff, test latest OS and bump setuptools

gaborbernat:


src/virtualenv/discovery/py_info.py
```diff
@@ -21,17 +21,17 @@
 
 
 def _get_path_extensions():
-    return list(OrderedDict.fromkeys([""] + os.environ.get("PATHEXT", "").lower().split(os.pathsep)))
+    return list(OrderedDict.fromkeys(["", *os.environ.get("PATHEXT", "").lower().split(os.pathsep)]))
```
ylecuyer, line 24: Looks like this syntax isn't compatible with all python version:

```
root@4ac062d1f6e7:~# python3 --version
Python 3.9.2
root@4ac062d1f6e7:~# virtualenv --version
virtualenv 20.23.1 from /usr/local/lib/python3.9/dist-packages/virtualenv/__init__.py
root@4ac062d1f6e7:~# virtualenv -p $(which python2) venv
RuntimeError: failed to query /usr/bin/python2 with code 1 err: '  File "/usr/local/lib/python3.9/dist-packages/virtualenv/discovery/py_info.py", line 24\n    return list(OrderedDict.fromkeys(["", *os.environ.get("PATHEXT", "").lower().split(os.pathsep)]))\n                                          ^\nSyntaxError: invalid syntax\n'
```

and the project is supposed to be compatible with

````
requires-python = ">=3.7"
```

gaborbernat: You're trying to create a Python 2 environment there, see https://virtualenv.pypa.io/en/latest/changelog.html#features-20-22-0 that's no longert supported. Use older version of the project, but really Python 2 is way past its EOL now.

Issue #2194, opened 2021-09-22
Make `activate` et al. export custom prompt prefix as an envvar


jwodder:
**What's the problem this feature will solve?**

I use [a custom script](https://github.com/jwodder/ps1.py) (and I'm sure many others have similar scripts as well) for setting my prompt in Bash.  It shows the name of the current virtualenv (if any) by querying the `VIRTUAL_ENV` environment variable, but if the virtualenv was created with a custom `--prompt`, it is unable to use this prompt prefix, as the `activate` script does not make this information available.

**Describe the solution you'd like**

The `activate` et al. scripts should set and export an environment variable named something like `VIRTUAL_ENV_PROMPT_PREFIX` that contains the prompt prefix (either custom or default) that virtualenv would prepend to the prompt.  Ideally, this should be set even when `VIRTUAL_ENV_DISABLE_PROMPT` is set in case the user wants total control over their prompt.

gaborbernat:
I'd be interested what core (venv) things about it, can you please create a ticket to https://bugs.python.org/

jwodder:
@gaborbernat Ticket created: https://bugs.python.org/issue45264

Tadaboody:
While no environment variable is created there is a `$VIRTUAL_ENV/pyvenv.cfg` file created with the `prompt` attribute.
You can see an example usage in https://github.com/starship/starship/pull/1747/files#diff-6a82ae9de73a3d1ea14bd519050b16337b03b9022d8d5dcc5e95db7c08f9db28R120

MegaBluejay:
Since 2020 venv does indeed set the `VIRTUAL_ENV_PROMPT` envar on activation: [commit](https://github.com/python/cpython/commit/c82dda1e08c4b74ca24f88d6a549d93108c319cf)

jimporter:
> @gaborbernat Ticket created: https://bugs.python.org/issue45264

I filed a [PR](https://github.com/python/cpython/pull/106643) against this issue ([GH version of the issue](https://github.com/python/cpython/issues/89427)) which ensures that `VIRTUAL_ENV_PROMPT` gets exported no matter the value of `VIRTUAL_ENV_DISABLE_PROMPT`. Once that PR is resolved (either by merging or closing), I'll file a PR for this repo to sync the behavior with the built-in `venv` package.

jimporter:
@gaborbernat I think we can close this issue now? Thanks for merging!

PR #2606, opened 2023-07-13
Export `VIRTUAL_ENV_PROMPT` in activation scripts (#2194)

jimporter:
See issue #2194. This change exports a `VIRTUAL_ENV_PROMPT` environment variable during venv activation so that users can use this in their own custom shell prompts if they like. ~I didn't add a test for this since I don't see any similar tests, and I'm not quite sure the bet way to go about writing one...~

I manually tested these changes in Bash and the results look good to me, though my changes for the other shell scripts are speculative. I did my best to guess how they should work, but there may be bugs!

---

### DIFF
--- a/docs/changelog/2194.feature.rst
+++ b/docs/changelog/2194.feature.rst
@@ -1,0 +1,2 @@
+Export the prompt prefix as ``VIRTUAL_ENV_PROMPT`` when activating a virtual
+environment - by :user:`jimporter`.
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -242,8 +242,10 @@ Note that you don't have to activate a virtual environment to use it. You can in
 executables, rather than relying on your shell to resolve them to your virtual environment.
 
 Activator scripts also modify your shell prompt to indicate which environment is currently active, by prepending the
-environment name in brackets, like ``(venv)``. You can disable this behaviour by setting the environment variable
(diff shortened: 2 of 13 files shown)


gaborbernat (approved):

gaborbernat (changes requested):
Please add extend the test suite to test for this.

jimporter:
> Please add extend the test suite to test for this.

I'll take a look. I'm going to mark this as a draft for now until I get tests resolved (and ideally for my second `venv` PR to merge so that I can be sure the behavior is the same between `venv` and `virtualenv`).

jimporter:
@gaborbernat Ok, hopefully these tests pass. I had to make some educated guesses for a few of the shells.

gaborbernat (changes requested):
You'll need to make the CI pass 😊

jimporter:
@gaborbernat Ok, I think these tests should be good now. One small note is that I renamed `VIRTUAL_PROMPT` to `VIRTUAL_PREFIX` in Nushell to reduce ambiguity, though I could revert that. As far as I can tell, that variable is only there to make it easier to write a unit test (this is the variable that holds `(<env-name>) `, as opposed to `VIRTUAL_ENV_PROMPT` which holds `<env-name>`).

gaborbernat (approved):

PR #2613, opened 2023-07-20
Fix #1949: zipapp virtual environment creation fails if zipapp path is symlinked

### LINKED ISSUE #1949: zipapp virtual environment creation fails if zipapp path is symlinked
# Issue

On a fresh new system, trying to use the zipapp to setup a virtualenv when the pyz files is referenced by a symlink, it fails saying `KeyError: "There is no item named 'nv/create/via_global_ref/_virtualenv.py' in the archive"`

## But it works on my machine?

Yes, if you ever used the virtualenv (zipapp?) successfully, it creates files in `~/Library/Application Support/virtualenv` (on macOS) or `~/.local/share/virtualenv/` (on Linux) and the existence of these files (unzip and wheel) makes the zipapp work. By deleting that directory, you should be able to reproduce the problem. This was first discovered on travis-ci, where the machine is new every time.

# Environment

- OS: macOS 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64.
- Python 3.7.7. (from brew), also tried 3.7.9

Also reproduced on
- OS: fedora:31
- Python 3.7.9

# Output of the virtual environment creation

```console
bash-3.2$ (cd /var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/foo; python3 /var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz -vvv --with-traceback /var/folders/g0/944401t11gz1jl7vxt
nzqs_m0001yj/T/foo/pipenv-2020.6.2)
206 setup logging to NOTSET [DEBUG report:43]
219 find interpreter for spec PythonSpec(path=/usr/local/opt/python/bin/python3.7) [INFO builtin:44]
219 proposed PythonInfo(spec=CPython3.7.7.final.0-64, exe=/usr/local/opt/python/bin/python3.7, platform=darwin, version='3.7.7 (default, Mar 10 2020, 15:43:33) \n[Clang 11.0.0 (clang-1100.0.33.17)]', encoding_fs_io=utf-8-utf-8) [INFO builtin:50]
219 accepted PythonInfo(spec=CPython3.7.7.final.0-64, exe=/usr/local/opt/python/bin/python3.7, platform=darwin, version='3.7.7 (default, Mar 10 2020, 15:43:33) \n[Clang 11.0.0 (clang-1100.0.33.17)]', encoding_fs_io=utf-8-utf-8) [DEBUG builtin:52]
224 filesystem is not case-sensitive [DEBUG info:28]
263 create virtual environment via CPython3Posix(dest=/private/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/foo/pipenv-2020.6.2, clear=False, global=False) [INFO session:52]
263 write /private/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/foo/pipenv-2020.6.2/pyvenv.cfg [DEBUG pyenv_cfg:34]
263     home = /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7 [DEBUG pyenv_cfg:38]
263     implementation = CPython [DEBUG pyenv_cfg:38]
263     version_info = 3.7.7.final.0 [DEBUG pyenv_cfg:38]
263     virtualenv = 20.0.31 [DEBUG pyenv_cfg:38]
263     include-system-site-packages = false [DEBUG pyenv_cfg:38]
264     base-prefix = /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7 [DEBUG pyenv_cfg:38]
264     base-exec-prefix = /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7 [DEBUG pyenv_cfg:38]
264     base-executable = /usr/local/opt/python/bin/python3.7 [DEBUG pyenv_cfg:38]
264 remove file /private/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/foo/pipenv-2020.6.2/bin/python [DEBUG _sync:38]
264 symlink /usr/local/opt/python/bin/python3.7 to /private/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/foo/pipenv-2020.6.2/bin/python [DEBUG _sync:44]
265 Attempting to acquire lock 4396469136 on /Users/a-andy/Library/Application Support/virtualenv/unzip/20.0.31/_virtualenv.py.lock [DEBUG filelock:270]
265 Lock 4396469136 acquired on /Users/a-andy/Library/Application Support/virtualenv/unzip/20.0.31/_virtualenv.py.lock [INFO filelock:274]
265 extract /var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/_virtualenv.py to /Users/a-andy/Library/Application Support/virtualenv/unzip/20.0.31/_virtualenv.py [DEBUG zipapp:19]
267 Attempting to release lock 4396469136 on /Users/a-andy/Library/Application Support/virtualenv/unzip/20.0.31/_virtualenv.py.lock [DEBUG filelock:315]
267 Lock 4396469136 released on /Users/a-andy/Library/Application Support/virtualenv/unzip/20.0.31/_virtualenv.py.lock [INFO filelock:318]
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/__main__.py", line 168, in <module>
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/__main__.py", line 164, in run
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/__main__.py", line 16, in run
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/run/__init__.py", line 28, in cli_run
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/run/session.py", line 46, in run
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/run/session.py", line 53, in _create
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/creator.py", line 157, in run
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/builtin/via_global_self_do.py", line 86, in create
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 88, in create
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 91, in install_patch
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/builtin/cpython/cpython3.py", line 27, in env_patch_text
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 102, in env_patch_text
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/app_data/base.py", line 54, in ensure_extracted
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/app_data/via_disk_folder.py", line 81, in extract
  File "/var/folders/g0/944401t11gz1jl7vxtnzqs_m0001yj/T/tmp.7cHdIsKw/virtualenv.pyz/virtualenv/util/zipapp.py", line 22, in extract

  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/zipfile.py", line 1431, in getinfo
    'There is no item named %r in the archive' % name)
KeyError: "There is no item named 'nv/create/via_global_ref/_virtualenv.py' in the archive"
```

**Note:** Notice `/virtuale` is missing from the KeyError, the exact length `/private` which is the difference between the TMPDIR and the real path of the TMPDIR on macOS

On Fedora, the last line becomes:

```console
KeyError: "There is no item named '/virtualenv/create/via_global_ref/_virtualenv.py' in the archive"
```

# Steps to reproduce

1. `docker run -it --rm fedora:31`
1. `mkdir -p /tmp/foo/bar`
1. `ln -s /tmp/foo/bar /tmp/foo/bar2`
1. `curl -L -o /tmp/foo/bar/virtualenv.pyz https://bootstrap.pypa.io/virtualenv/3.7/virtualenv.pyz`
1. `python3 /tmp/foo/bar2/virtualenv.pyz /tmp/foo/blah`

Here are some one-liners to reproduce the error

```
docker run -it --rm fedora:32 bash -c 'REAL=/tmp/foo/car; SYM=/tmp/foo/bbb2; mkdir -p "${REAL}"; ln -s "${REAL}" "${SYM}"; curl -L -o "${REAL}/virtualenv.pyz" https://bootstrap.pypa.io/virtualenv/3.8/virtualenv.pyz; python3 "${SYM}/virtualenv.pyz" /tmp/blah'
```

```
docker run -it --rm fedora:32 bash -c 'REAL=/tmp/foo/car; SYM=/tmp/foo/bbb2; mkdir -p "${REAL}"; ln -s "${REAL}" "${SYM}"; curl -L -o "${REAL}/virtualenv.pyz" https://bootstrap.pypa.io/virtualenv/3.8/virtualenv.pyz; python3 "${SYM}/virtualenv.pyz" /tmp/blah'
```

Not all values of REAL/SYM fail. It appears if the length of the symlink path and realpath differ, it fails. If they are the exact same length, no error occurs

**Fail**

REAL | SYM
--|--
/tmp/foo/bar|/tmp/bar/bar2
/tmp/foo/bar|/tmp/foo/bc
/tmp/foo/car|/tmp/foo/bb
/tmp/bb|/tmp/aaa

**Does not fail**

REAL | SYM
--|--
/tmp/foo|/tmp/bar
/tmp/foo/bar|/tmp/foo/abc
/tmp/foo/car|/tmp/foo/bbb

petamas:
Original issue: https://github.com/pypa/virtualenv/issues/1949

The function in question assumes that `ROOT` is a prefix of `full_path`, and tries to remove it. However, it is not necessarily a prefix:
- when `ROOT` is determined, both `os.path.abspath()` and `os.path.realpath()` gets called
- when `full_path` is determined, only `os.path.abspath()` is called

This means that if the path contains a symlink, it will show up as resolved in `ROOT`, but not `full_path`, so simply removing the first X characters of `full_path` is not sufficient.

This PR adds an assertion to catch the bad situation early, then adds the `os.path.realpath` call to `full_path`.

I did not add a test because I was unable to figure out how to add it to the existing structure.

### DIFF
--- a/src/virtualenv/util/zipapp.py
+++ b/src/virtualenv/util/zipapp.py
@@ -24,7 +24,9 @@ def extract(full_path, dest):
 
 def _get_path_within_zip(full_path):
     full_path = os.path.abspath(str(full_path))
-    sub_file = full_path[len(ROOT) + 1 :]
+    prefix = ROOT + os.sep
+    assert full_path.startswith(prefix), f"full_path={full_path} should start with prefix={prefix}"
+    sub_file = full_path[len(prefix) :]
     if IS_WIN:
         # paths are always UNIX separators, even on Windows, though __file__ still follows platform default
         sub_file = sub_file.replace(os.sep, "/")
--- a/src/virtualenv/util/zipapp.py
+++ b/src/virtualenv/util/zipapp.py
@@ -23,7 +23,7 @@ def extract(full_path, dest):
 
 
 def _get_path_within_zip(full_path):
-    full_path = os.path.abspath(str(full_path))
+    full_path = os.path.realpath(os.path.abspath(str(full_path)))
     prefix = ROOT + os.sep
     assert full_path.startswith(prefix), f"full_path={full_path} should start with prefix={prefix}"
     sub_file = full_path[len(prefix) :]
--- a/docs/changelog/1949.bugfix.rst
+++ b/docs/changelog/1949.bugfix.rst
@@ -1,0 +1,1 @@
+``virtualenv.pyz`` no longer fails when zipapp path contains a symlink - by :user:`petamas`.


petamas:
@gaborbernat `ruff` is failing this PR because it considers using `assert`s harmful. (https://beta.ruff.rs/docs/rules/assert/) I strongly believe that an assert is the right way to address this error case, because it can only happen by a programmer error (eg. if someone changes the definition of `ROOT`), but naturally I'm willing to adapt to maintainers' taste.

I see the following possible solutions:
1. Disable S101 for this line specifically
2. Disable S101 globally (if you also happen to disagree with the rationale)
3. Remove the assert, only fix the single line
4. Replace the assert with throwing an exception (I have no idea what kind, possibly `ValueError`?)

How do you want me to proceed?

gaborbernat:
> 1. Disable S101 for this line specifically

src/virtualenv/util/zipapp.py
```diff
@@ -23,8 +23,10 @@ def extract(full_path, dest):
 
 
 def _get_path_within_zip(full_path):
-    full_path = os.path.abspath(str(full_path))
-    sub_file = full_path[len(ROOT) + 1 :]
+    full_path = os.path.realpath(os.path.abspath(str(full_path)))
+    prefix = ROOT + os.sep
```
gaborbernat, line 27: Can we use here the pathlib variant instead?

gaborbernat (changes requested):
Please add tests validating your change fixes the problem.

src/virtualenv/util/zipapp.py
petamas: Sure. I saw most of the other code uses pathlib, but I wasn't sure whether it's intentional or not, and wanted to keep the size of the diff minimal. It's way simpler to implement the whole function using `PurePath.relative_to` anyway.

Question: if I migrate to pathlib, the trivial implementation will throw ValueError in case `ROOT` and `full_path` diverge, i.e. when my assert fails in the current version. Is that acceptable, or should I turn it back to an assert?

gaborbernat: The ValueError is fine considering we're not expecting it to happen.

petamas:
> Please add tests validating your change fixes the problem.

Can you point me in the direction of where it should be added? I'm not that familiar with either pytest or tox. I also don't know how I could make sure the test passes because the issue is really fixed, or because of the caching behaviour described in the original report. Or should I add a unit test for the zipapp functionality specifically? Would the place for that be in `test_util.py` or somewhere else?

gaborbernat:
You want to add it here https://github.com/pypa/virtualenv/blob/main/tests/integration/test_zipapp.py

gaborbernat:
Seems stalled, closing until you pick it up again.

petamas:
Yeah, sorry, I'm swamped, and I'll be AFK next week. I hope to eventually get back to this and reopen...

Issue #2472, opened 2022-12-26
ResourceWarning on env's setup


stanislavlevin:
**Issue**
There are `ResourceWarning`s on virtual env creation.

**Environment**

- `pip list` of the host python where `virtualenv` is installed:

```console
(integration) [builder@localhost pyproject_installer]$ pip list
Package        Version
-------------- -------
attrs          22.2.0
distlib        0.3.6
exceptiongroup 1.1.0
filelock       3.8.2
iniconfig      1.1.1
packaging      22.0
pip            22.3.1
platformdirs   2.6.0
pluggy         1.0.0
pytest         7.2.0
setuptools     65.6.3
tomli          2.0.1
virtualenv     20.17.1
wheel          0.38.4
  ```

**Output of the virtual environment creation**

```console
(integration) [builder@localhost pyproject_installer]$ python -W error::ResourceWarning -m virtualenv --no-setuptools --no-wheel -vvv --with-traceback test
280 setup logging to NOTSET [DEBUG report:35]
280 created app data folder /usr/src/.local/share/virtualenv [DEBUG __init__:40]
284 find interpreter for spec PythonSpec(path=/usr/src/RPM/BUILD/pyproject_installer/.tox/integration/bin/python) [INFO builtin:56]
284 discover exe for PythonInfo(spec=CPython3.10.8.final.0-64, exe=/usr/src/RPM/BUILD/pyproject_installer/.tox/integration/bin/python, platform=linux, version='3.10.8 (main, Dec 17 2022, 11:34:59) [GCC 12.1.1 20220518 (ALT Sisyphus 12.1.1-alt2)]', encoding_fs_io=utf-8-utf-8) in /usr [DEBUG py_info:437]
284 filesystem is case-sensitive [DEBUG info:24]
285 Attempting to acquire lock 140183178084464 on /usr/src/.local/share/virtualenv/py_info/1/8a94588eda9d64d9e9a351ab8144e55b1fabf5113b54e67dd26a8c27df0381b3.lock [DEBUG _api:172]
285 Lock 140183178084464 acquired on /usr/src/.local/share/virtualenv/py_info/1/8a94588eda9d64d9e9a351ab8144e55b1fabf5113b54e67dd26a8c27df0381b3.lock [DEBUG _api:176]
[...]

PR #2617, opened 2023-08-04
fix ResourceWarning on process exit (#2472)

sigma67:

- [x] wrote descriptive pull request text - see #2472
- [] ensured there are test(s) validating the fix

### DIFF
--- a/src/virtualenv/seed/wheels/periodic_update.py
+++ b/src/virtualenv/seed/wheels/periodic_update.py
@@ -12,7 +12,7 @@ from datetime import datetime, timedelta, timezone
 from itertools import groupby
 from pathlib import Path
 from shutil import copy2
-from subprocess import PIPE, Popen
+from subprocess import DEVNULL, Popen
 from textwrap import dedent
 from threading import Thread
 from urllib.error import URLError
@@ -216,7 +216,7 @@ def trigger_update(distribution, for_py_version, wheel, search_dirs, app_data, e
         .format(distribution, for_py_version, wheel_path, str(app_data), [str(p) for p in search_dirs], periodic),
     ]
     debug = env.get("_VIRTUALENV_PERIODIC_UPDATE_INLINE") == "1"
-    pipe = None if debug else PIPE
+    pipe = None if debug else DEVNULL
     kwargs = {"stdout": pipe, "stderr": pipe}
     if not debug and sys.platform == "win32":
         kwargs["creationflags"] = CREATE_NO_WINDOW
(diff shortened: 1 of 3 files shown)


sigma67:
@gaborbernat I decided on the simpler solution with the context manager to let the stdlib do the heavy lifting

src/virtualenv/seed/wheels/periodic_update.py
```diff
@@ -220,16 +220,17 @@ def trigger_update(distribution, for_py_version, wheel, search_dirs, app_data, e
...
-        "triggered periodic upgrade of %s%s (for python %s) via background process having PID %d",
-        distribution,
-        "" if wheel is None else f"=={wheel.version}",
-        for_py_version,
-        process.pid,
-    )
-    if debug:
-        process.communicate()  # on purpose not called to make it a background process
+
+    with Popen(cmd, **kwargs) as process:  # noqa: S603
```
gaborbernat, line 224: What happens here? Wouldn't this make it blocking call?

sigma67: Yes, that's the point. You can't keep a subprocess open while your main process exits without getting a `ResourceWarning`, afaik. Is the intention to keep this process running until it terminates even if the main process finishes?

If you want to continue the rest of the program execution and only wait for the subprocess to finish at the end you'd need to register the process somewhere central and ensure termination.

Another option maybe is to use [`close_fds` param of `Popen`](https://docs.python.org/3/library/subprocess.html#popen-constructor), but I'm not sure if that fixes the issue.

gaborbernat: Ideally, I'd like to just ignore the resource warning. There's no need for this process to finish when the main process finishes. It's more of a fire and forget kinda thing.

sigma67: Nevermind, it doesn't. I was running with `--no-periodic-update`.

sigma67: Yes sorry, after a bunch of research: the proper way to do this is to wait until your subprocess finishes before your application exits. You don't want a zombie process hanging around. Even solutions using asyncio recommend to wait for async tasks to finish before main process exit, else you'll get a warning.

gaborbernat: > You don't want a zombie process hanging around.

That's exactly what I want here though.

sigma67: So, the unclosed file warning stems from setting `pipe = PIPE` in the non-debug case. It disappears if we set it to `pipe=subprocess.DEVNULL` instead. 

But then we get a different `ResourceWarning`: `subprocess is still running`

The warning was added in 3.6, here: https://bugs.python.org/issue26741
Pretty much everyone saying keeping a zombie process is bad practice, but here we go.

The followup issue mentions some potential solutions to hide the warning, but `detach` didn't work: https://bugs.python.org/issue27068

So I had a look at the CPython source, turns out if we set the returncode manually (since we don't care about it anyway), the warning will not be raised: https://github.com/python/cpython/blob/85e5b1f5b806289744ef9a5a13dabfb23044f713/Lib/subprocess.py#L1127 (from main, just to make it a permalink)

Tests passed locally.

gaborbernat (approved):

PR #2638, opened 2023-09-07
GitHub Actions: Add PyPy 3.10 to testing

cclauss:
A subset of
* #2634

- [x] wrote descriptive pull request title


tests/unit/config/test_ini.py
```diff
@@ -10,7 +10,7 @@
 
 
 @pytest.mark.skipif(not fs_supports_symlink(), reason="symlink is not supported")
-@pytest.mark.xfail(IS_PYPY and IS_WIN and sys.version_info[0:2] == (3, 9), reason="symlink is not supported")
+@pytest.mark.xfail(IS_PYPY and IS_WIN and sys.version_info[0:2] >= (3, 9), reason="symlink is not supported")
```
gaborbernat, line 13: Rather than skip the test can we understand why they don't support symlink.

gaborbernat (changes requested):

.github/workflows/check.yml
```diff
@@ -56,7 +57,7 @@ jobs:
         with:
           python-version: "3.11"
       - name: Install tox
-        run: python -m pip install tox pip -U
+        run: python -m pip install --upgrade pip tox
```
gaborbernat, line 60: Revert this orthogonal change.

cclauss: Pip has at least five different options that start with `--u` so I always think it is lazy and not self-documenting to use `-U` unless one it typing on the command line.

Keeping lists sorted makes it easy to spot missing items and nearly impossible insert duplicates.

tests/unit/config/test_ini.py
cclauss: As [discussed in the PyPy](https://doc.pypy.org/en/latest/install.html?highlight=symlink) modern versions want users who want a system-wide PyPy to create a symlink in their bin dir so that PyPy can still find its library.

.github/workflows/check.yml
cclauss: Six...
```
  --user                      Install to the Python user install directory for
  -U, --upgrade               Upgrade all specified packages to the newest
  --upgrade-strategy <upgrade_strategy>
  --use-pep517                Use PEP 517 for building source distributions
  --use-feature <feature>     Enable new functionality, that may be backward
  --use-deprecated <feature>  Enable deprecated functionality, that will be
```

gaborbernat: Be that so, but is an orthogonal change to this PR.

tests/unit/config/test_ini.py
gaborbernat: Let's add link to that discussion.

gaborbernat:
Please rebasa against main 👍

gaborbernat:
Also update https://github.com/pypa/virtualenv/blob/main/docs/installation.rst#python-and-os-compatibility

gaborbernat (approved):

gaborbernat (approved):

gaborbernat (approved):

cclauss:
Nice!!

PR #2647, opened 2023-09-23
Apply a few suggestions from refurb

DimitriPapadopoulos:
These [refurb](https://github.com/dosisod/refurb) suggestions appear to improve readability.

The changes seem trivial to me, hence the lack of news fragment.

### DIFF
--- a/src/virtualenv/app_data/na.py
+++ b/src/virtualenv/app_data/na.py
@@ -53,7 +53,6 @@ class ContentStoreNA(ContentStore):
 
     def read(self):
         """Nothing to read."""
-        return
 
     def write(self, content):
         """Nothing to write."""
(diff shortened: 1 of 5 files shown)


gaborbernat (changes requested):
I don't agree with most of these .

DimitriPapadopoulos:
> I don't agree with most of these .

Why?

* The return is missing from the next function. Either add it to `write()`, or remove it from `read()`:
  https://github.com/pypa/virtualenv/blob/390ed2a5f6963f8867a916c161fc2c1dc04eda96/src/virtualenv/app_data/na.py#L54-L60
* What's the point of converting a `str` literal to `str`?
  https://github.com/pypa/virtualenv/blob/390ed2a5f6963f8867a916c161fc2c1dc04eda96/tests/conftest.py#L189-L190
* The `or` instead of `if ... else` is used in tox, so it cannot be that bad. See for example:
  https://github.com/tox-dev/tox/blob/85ce052d60aadb4fce779c404c052074bc530fbc/src/tox/config/loader/api.py#L63

PR #2653, opened 2023-10-16
`source venv/bin/activate` now works between different shells (fish/zsh/ect...)

TheRook:

Hey everyone I noticed `source venv/bin/activate` wasn't working on fish, and that is when I found a better way of passing environment variables around.  If you execute a bash script with `#!/usr/bin/env bash` you will pass in environment variables, and features like EXPORT and SET will also work. 

There is no reason not to have this feature, as it is helpful for anyone who cannot use bash, but would like to use the default venv files produced.


gaborbernat:
On fish you need to use activate.fish not that script.

PR #2687, opened 2024-02-05
Fix windows utf8 encoding issue

PzaThief:
Batch and Powershell do not work properly if non-ASCII text (e.g. "аạąäàáą") is used in the folder name or prompt name.
For Batch case, there is a codepage issue under the problem and [cpython's venv](https://github.com/python/cpython/blob/main/Lib/venv/scripts/nt/activate.bat) has the solution already.
And Powershell assumes Windows 1252 encoding when reading files without BOM. So I added Bom using utf-8-sig encoding.

Batch error screenshot (see broken prompt name after activate)
Batch error screenshot

Powershell error screenshot
Powershell error screenshot


gaborbernat (changes requested):
CI is failing.

gaborbernat (changes requested):
Tests please.

PzaThief:
> Tests please.

Sorry for late.
I fixed test failing now. could you check CI again?

PR #2688, opened 2024-02-06
Avoid importing contextlib in `_virtualenv`

hauntsaninja:
This appears to be imported in a `.pth` file, so this file runs in basically every Python program in my environment. contextlib takes about 3ms to import and I'd prefer to not pay that cost if not strictly necessary.

- [] ensured there are test(s) validating the fix

### DIFF
--- a/docs/changelog/2688.misc.rst
+++ b/docs/changelog/2688.misc.rst
@@ -1,0 +1,1 @@
+Avoid importing :mod:`contextlib` in ``_virtualenv.py``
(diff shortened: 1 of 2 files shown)


hauntsaninja:
This was introduced in https://github.com/pypa/virtualenv/pull/2588 , presumably as part of a mass application of SIM105

hauntsaninja:
Looks like tests are failing on #2685 as well...

hauntsaninja:
(Also noob question: what is the `os.path.join(__file__)` idiom meant to do?)

gaborbernat:
> (Also noob question: what is the `os.path.join(__file__)` idiom meant to do?)

Probably a no-op, could try removing it.

Issue #2708, opened 2024-04-23
Discovery with a generic spec fails to find specific python


flying-sheep:
**Issue**

See https://github.com/pypa/hatch/issues/1395

`virtualenv.discovery.builtin.get_interpreter('3')` fails to find e.g. `/usr/bin/python3.12`.

**Environment**

All environments: Windows, Linux, macOS.

Tested with virtualenv master, currently after 20.25.3

PR #2709, opened 2024-04-23
Allow builtin interpreter discovery to find specific Python versions given a general spec

flying-sheep:

Fixes https://github.com/pypa/virtualenv/issues/2708, fixes https://github.com/pypa/hatch/issues/1395

This PR fixes the discovery in `virtualenv.discovery.builtin`. 

It works by removing the method `PythonSpec.generate_names` and replacing it with `generate_re`:

A static name list can’t represent all Python executables that are compatible with a given spec, i.e. `PythonSpec.from_string_spec('3')` should match a executable called `/usr/bin/python3.12`. I therefore inverted the way candidates are found. Instead of generating all possible candidates, we simply loop and use a regex to find candidates.

Pro:
- Fixes the bug
- Instead of just trying all-lowercase and all-uppercase variants of candidates, we’re now truly case-insensitive

Con:
- Might be slower if a lot of executables hang out on PATH. Since Python’s `glob` just traverses directories in Python code as well, it wouldn‘t help here.

## How to review

The main commits are
- f29be4a71d09afef4aece4adef89a2e47dd4086f: Fixes the tox configuration so it can cope with spaces. Allows me to run tests on my system. Humanity needs to stop interpolating strings into shell scripts.
- 6f9f270ebfa491cdf5d9c3a3bd97c9647d7a8a74: Adds typing to the file I’m going to edit so I can make sense of what’s going on
- 16a6e1d5a0ccf1aaf9f44400f4f8bd0eb0ced9c6: Switch that file to pathlib, since you seem to be doing that elsewhere and it makes things more readable
- d9b52ada0f67e8c85ce58bd6f46c5aa37e1201bd: The actual changes, described above

### DIFF
--- a/docs/changelog/2709.bugfix.rst
+++ b/docs/changelog/2709.bugfix.rst
@@ -1,0 +1,1 @@
+allow builtin discovery to discover specific interpreters (e.g. ``python3.12``) given an unspecific spec (e.g. ``python3``) - by :user:`flying-sheep`.
--- a/src/virtualenv/discovery/builtin.py
+++ b/src/virtualenv/discovery/builtin.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import os
 import sys
+from typing import TYPE_CHECKING
 
 from virtualenv.info import IS_WIN
 
@@ -10,8 +11,18 @@ from .discover import Discover
 from .py_info import PythonInfo
 from .py_spec import PythonSpec
 
+if TYPE_CHECKING:
+    from argparse import ArgumentParser
+    from collections.abc import Generator, Iterable, Mapping, Sequence
+
+    from virtualenv.app_data.base import AppData
+
 
 class Builtin(Discover):
+    python_spec: Sequence[str]
+    app_data: AppData
+    try_first_with: Sequence[str]
+
     def __init__(self, options) -> None:
         super().__init__(options)
         self.python_spec = options.python or [sys.executable]
@@ -19,7 +30,7 @@ class Builtin(Discover):
         self.try_first_with = options.try_first_with
 
     @classmethod
-    def add_parser_arguments(cls, parser):
+    def add_parser_arguments(cls, parser: ArgumentParser) -> None:
         parser.add_argument(
             "-p",
             "--python",
@@ -41,7 +52,7 @@ class Builtin(Discover):
             help="try first these interpreters before starting the discovery",
         )
 
-    def run(self):
+    def run(self) -> PythonInfo | None:
         for python_spec in self.python_spec:
             result = get_interpreter(python_spec, self.try_first_with, self.app_data, self._env)
             if result is not None:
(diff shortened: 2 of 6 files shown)


gaborbernat (changes requested):
Tests please 🥺

flying-sheep:
On it! It’s a draft so you can look at it if you want while I check those last boxes!

flying-sheep:
Now I’m done!

Seems like this change snuck into one of the tests:

```diff
-    pyvenv_cfg = Path(sys.executable).parents[1] / "pyvenv.cfg"
-    if pyvenv_cfg.exists():
-        (target / pyvenv_cfg.name).write_bytes(pyvenv_cfg.read_bytes())
```

that fragment doesn’t seem to be necessary to run the test locally?

gaborbernat (approved):

gaborbernat:
Seems you've upset the CI 😅

flying-sheep:
Huh, tests run fine locally … let’s see.

Seems like it works in practice though!

1. create project requiring Python 3.12 on a system where `/usr/bin/python` is 3.11:

	```console
	$ cd some-project
	$ rg requires-python pyproject.toml 
	4:requires-python = ">=3.12"
	```

2. try finding an interpreter with hatch:

	```console
	$ pipx install hatch
	$ hatch env create
	Environment `default` is incompatible: no compatible Python distribution available
	```

2. try my patch:

	```console
	$ ~/.local/pipx/venvs/hatch/bin/python -m pip install 'virtualenv@ git+https://github.com/flying-sheep/virtualenv.git@discover-more'
	[…]
	ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
	hatch 1.9.4 requires virtualenv>=20.16.2, but you have virtualenv 0.1.dev788+g66d9408 which is incompatible.
	Successfully installed virtualenv-0.1.dev788+g66d9408
	$ hatch env create
	```

	no output means success!

gaborbernat:
> Huh, tests run fine locally … let’s see.

I cannot meege the pull request until the CIA is green so it doesn't really matter that "it works on my machine" 😅. Sorry about the CI gods are absolute. 😭

flying-sheep:
Ah, the regex always needs to be case insensitive to get the intended effect. I tested locally on macOS and Linux now, so I hope windows doesn’t throw a wrench.

flying-sheep:
OK, I can only speculate that the `pyvenv.cfg` stuff was load bearing on windows?

If that’s not it, someone with a windows machine needs to help.

gaborbernat (dismissed):

flying-sheep:
Looks good! I don’t think this timeout is caused by my code

flying-sheep:
OK, pypy on windows fails with

```
125251 get interpreter info via cmd: 'C:\Users\runneradmin\AppData\Local\Temp\pytest-of-unknown\pytest-0\test_discovery_via_path_specif0\Scripts\somethingVeryCryptic3.10.exe' 'D:\a\virtualenv\virtualenv\.tox\pypy310\lib\site-packages\virtualenv\discovery\py_info.py' 2o7zKhwQ7RmUZlP1L7qYsRzWDWJd2Uc0 266eD4ebldJtJex0LCwbXGC040ZIuScj [DEBUG cached_py_info:112]
125280 failed to query C:\Users\runneradmin\AppData\Local\Temp\pytest-of-unknown\pytest-0\test_discovery_via_path_specif0\Scripts\somethingVeryCryptic3.10.exe with code 3221225781 [INFO cached_py_info:34]
```

I have absolutely no idea what “125280 failed to query …exe with code 3221225781” means, so I don’t think I can proceed here.

PR #2716, opened 2024-05-08
Update activate_this.py documentation to use runpy instead of exec

FredStober:
runpy.run_path was added in python 2.7 and 3.2 - and every python that is not EOL supports it.

It is arguably nicer to read and the path is only given once in the command.

At least right now, runpy - unlike exec with S102 - is not flagged by bandit or bandit-derived (eg. ruff) checks.
(I guess because it loads from a file instead of a simple string...)

Because of the import, it is also not a one-liner anymore. 

- [X] ensured there are test(s) validating the fix - the change only updates documentation
- [X] added news fragment in `docs/changelog` folder: This is a tiny change, so I'm not sure it warrants a news fragment

### DIFF
--- a/src/virtualenv/activation/python/activate_this.py
+++ b/src/virtualenv/activation/python/activate_this.py
@@ -1,7 +1,8 @@
 """
 Activate virtualenv for current interpreter:
 
-Use exec(open(this_file).read(), {'__file__': this_file}).
+import runpy
+runpy.run_path(this_file)
 
 This can be used when you must use an existing Python interpreter, not the virtualenv bin/python.
 """  # noqa: D415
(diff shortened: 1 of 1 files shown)


tasks/make_zipapp.py
```diff
@@ -180,7 +180,7 @@ def get_dependencies(whl, version):
             platforms = []
             platform_positions = WheelDownloader._marker_at(markers, "sys_platform")
             deleted = 0
-            for pos in platform_positions:  # can only be ore meaningfully
+            for pos in platform_positions:  # can only be more meaningfully
```
pfmoore, line 183: "more" doesn't make sense here. Was the typo meant to be "one"? I'm not sure if that makes any more sense, but at least it's a grammatically complete phrase...

FredStober: tbh - that whole comment didn't make sense to me either - but pre-commit wanted to change it. I'm happy to leave it out, since it's in a file I was not really interested in :)

gaborbernat (approved):

PR #2717, opened 2024-05-08
Fix bad return code in bash activation if hashing is disabled

fenkes-ibm:
If hashing is disabled in bash, the `hash` command will return a nonzero return code. Since it is the last command in the script that return code will also be the "return code" of the source command, so if anyone uses

source activate.sh || die horribly

having hashing disabled will ruin their day.

Fix this by overriding the return code of `hash` if it's bad. Add tests to verify a good return code and try with and without hashing.

### DIFF
--- a/docs/changelog/2717.bugfix.rst
+++ b/docs/changelog/2717.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bad return code from activate.sh if hashing is disabled - by :user:'fenkes-ibm'.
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
CI failing.

fenkes-ibm:
Ah, looks like I wedged in before #2721. Rebased, let's try again.

gaborbernat (approved):

fenkes-ibm:
brew@3.11 on macos timed out? :raised_eyebrow: maybe it just had a bad day and needs a rerun?

PR #2726, opened 2024-05-20
Update pyproject.toml so it validates.

hexmode:
While working with the Poetry's [proposed support for the `project` section](https://github.com/python-poetry/poetry/pull/9135) in pyproject.toml, it identified the following problems:

``` console
$ poetry check

The Poetry configuration is invalid:
  - project.license must be valid exactly by one definition (0 matches found)
  - Either [project.version] or [tool.poetry.version] is required in package mode.
```

Since the [license field is awkward](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license), [is not required](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license), and is specified in the classifiers, it could be removed.

The second issue from `poetry check` is probably a non-issue since the version is dynamic.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ keywords = [
   "isolated",
   "virtual",
 ]
-license = "MIT"
 maintainers = [
   { name = "Bernat Gabor", email = "gaborjbernat@gmail.com" },
 ]


pyproject.toml
```diff
@@ -14,7 +14,6 @@ keywords = [
   "isolated",
   "virtual",
 ]
-license = "MIT"
```
gaborbernat, line 17: This is correct according to the newer standards. The version Field is optional because can be dynamic as is here. It sounds to me like poetry has some broken checks.

gaborbernat (changes requested):

gaborbernat:
Please refer to https://peps.python.org/pep-0639/

pyproject.toml
radoering: @gaborbernat Are you saying that https://packaging.python.org/en/latest/specifications/pyproject-toml/#license is not up to date?

gaborbernat: Yes, and no. This is a specification that is in draft, so not yet accepted however our build back end already uses it.

PR #2736, opened 2024-06-12
limit filelock to highest compatible version

emmettbutler:
filelock 3.15.0 was just released and causes the following crash when used with the latest version of virtualenv

```
  File "/root/.pyenv/versions/3.10.11/lib/python3.10/site-packages/filelock/_api.py", line 365, in __del__
    self.release(force=True)
  File "/root/.pyenv/versions/3.10.11/lib/python3.10/site-packages/virtualenv/util/lock.py", line 34, in release
    with self.thread_safe:
AttributeError: '_CountedFileLock' object has no attribute 'thread_safe'
```

This change avoids depending on this new incompatible version of filelock.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dynamic = [
 ]
 dependencies = [
   "distlib<1,>=0.3.7",
-  "filelock<4,>=3.12.2",
+  "filelock<3.15,>=3.12.2",
   "importlib-metadata>=6.6; python_version<'3.8'",
   "platformdirs<5,>=3.9.1",
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -44,7 +44,7 @@ dynamic = [
 ]
 dependencies = [
   "distlib<1,>=0.3.7",
-  "filelock<4,>=3.12.2",
+  "filelock<3.15,>=3.12.2",
```
gaborbernat, line 47: I'd rather just disallow that version, not limit all future versions.

gaborbernat (changes requested):

gaborbernat (approved):

PR #2783, opened 2024-10-17
Fix broken Windows zipapp and drop 3.7 support

gaborbernat:
Resolves https://github.com/pypa/distlib/issues/235
Resolves https://github.com/pypa/virtualenv/issues/2784

Both change sets are needed to make the CI pass.


henryiii:
Not sure it was expected, but this change also breaks making a 3.7 venv, not just running on 3.7. Seen in https://github.com/pypa/pyproject-metadata/pull/207; nox can no longer target 3.7.

gaborbernat:
That is very much expected. tox does https://tox.wiki/en/4.23.0/faq.html#testing-end-of-life-python-versions NOx should do similar

Issue #2728, opened 2024-05-27
Activating venv create unwanted console output 


ShootGan:
**Issue**
Activating venv shouldn't generate any output. In versions above 20.25.1 it prints `()` on activation.

20.25.1:
```console
C:\projects\pythontests>venv\Scripts\activate.bat

(venv) C:\projects\pythontests>
```
20.25.2 (same effect in the latest version)
```console
C:\projects\pythontests>venv\Scripts\activate.bat

C:\projects\pythontests>()
```

**Environment**

Provide at least:

- OS: Windows 10, Windows 11
- `pip list` of the host python where `virtualenv` is installed:

```console
Package      Version
distlib      0.3.8
filelock     3.14.0
pip          22.0.4
platformdirs 4.2.2
setuptools   58.1.0
virtualenv   20.25.2
```

**Output of the virtual environment creation**

Make sure to run the creation with `-vvv --with-traceback`:
20.25.1
```console
C:\projects\pythontests>python -m virtualenv -vvv --with-traceback env
417 setup logging to NOTSET [DEBUG report:37]
565 find interpreter for spec PythonSpec(path=C:\Users\PNalepka\AppData\Local\Programs\Python\Python39\python.exe) [INFO builtin:58]
565 proposed PythonInfo(spec=CPython3.9.12.final.0-64, exe=C:\Users\PNalepka\AppData\Local\Programs\Python\Python39\python.exe, platform=win32, version='3.9.12 (tags/v3.9.12:b28265d, Mar 23 2022, 23:52:46) [MSC v.1929 64 bit (AMD64)]', encoding_fs_io=utf-8-utf-8) [INFO builtin:65]
568 accepted PythonInfo(spec=CPython3.9.12.final.0-64, exe=C:\Users\PNalepka\AppData\Local\Programs\Python\Python39\python.exe, platform=win32, version='3.9.12 (tags/v3.9.12:b28265d, Mar 23 2022, 23:52:46) [MSC v.1929 64 bit (AMD64)]', encoding_fs_io=utf-8-utf-8) [DEBUG builtin:67]
580 symlink on filesystem does not work [DEBUG info:44]
580 filesystem is not case-sensitive [DEBUG info:25]
702 create virtual environment via CPython3Windows(dest=C:\projects\pythontests\env, clear=False, no_vcs_ignore=False, global=False) [INFO session:50]
702 create folder C:\projects\pythontests\env\Lib\site-packages [DEBUG _sync:12]
[...]

PR #2801, opened 2024-11-22
No longer forcibly echo off during windows batch activation

wiktorinox:
During fixing of https://github.com/pypa/virtualenv/issues/2728 new issue has been introduced. 
Adding `echo off` at the beginning of the activate.bat cause this setting to persist after activation. 
It doesn't occur when activated in the interactive shell, but it does when activating inside of the Windows batch script - incidentaly that's how Jenkins `bat` step executes it's contents. 
This behaviour is unexpected and activation should not force specific `echo` on/off setting.

This change removes that `echo off` while still fixing https://github.com/pypa/virtualenv/issues/2728 - no additional parentheses are displayed when activating in windows batch.

I would love to write tests for that but I have no idea how - the way current tests are written is pretty convoluted and I don't understand them. However, I hope that this change is small enough that tests won't be necessary.


gaborbernat (changes requested):
Test.

wiktorinox:
@gaborbernat Like I wrote in the PR - I don't understand how to write the tests in this repository. If you could advise what's the execution flow and what parts are necessary for the test and where the assertions should live, then I would be able to write the tests.

For example for the test for this functionality I need to be able to create a bat file that will contain invocation of `activate.bat` and then observe output - is that even possible in your framework?

wiktorinox:
I've added a test for this case.

gaborbernat (approved):

Issue #2776, opened 2024-10-04
virtualenv is not able to detect free-threading python3.13t


ssbarnea:
**Issue**

The free-threading python 3.13 has a new executable `python3.13t` and virtualenv is unable to detect it.

**Environment**

Any

**Output of the virtual environment creation**

Make sure to run the creation with `-vvv --with-traceback`:

```console
$ virtualenv -p 3.13t
RuntimeError: failed to find interpreter for Builtin discover of python_spec='3.13t'
FAIL: 1


$ virtualenv -p 3.13
usage: virtualenv [--version] [--with-traceback] [-v | -q] [--read-only-app-data] [--app-data APP_DATA] [--reset-app-data] [--upgrade-embed-wheels] [--discovery {builtin}] [-p py] [--try-first-with py_exe]
                  [--creator {builtin,cpython3-posix,venv}] [--seeder {app-data,pip}] [--no-seed] [--activators comma_sep_list] [--clear] [--no-vcs-ignore] [--system-site-packages] [--symlinks | --copies] [--no-download | --download]
                  [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--wheel version] [--no-pip] [--no-setuptools] [--no-wheel] [--no-periodic-update] [--symlink-app-data] [--prompt prompt] [-h]
                  dest
virtualenv: error: the following arguments are required: dest
SystemExit: 2
FAIL: 2
```

PR #2809, opened 2024-11-29
Add free-threaded Python support

robsdedude:
# Add free-threaded Python support

This PR adds the possibility to specify a free threaded Python.
Examples:
 * `virtualenv -p 3.13t .venv`
 * `virtualenv -p 3t .venv`
 * `virtualenv -p cpython3.13t .venv`

To be able to change `PythonInfo` to contain a flag whether the python version is a free-threaded one or not, I've introduced versioning to the py info cache files as I couldn't find another way to invalidate old cached infos.

Closes: #2776

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,6 @@ skip_missing_interpreters = true
 
 [testenv]
 description = run tests with {basepython}
-base_python = {env:TOX_BASEPYTHON:{envname}}
 package = wheel
 wheel_build_env = .pkg
 extras =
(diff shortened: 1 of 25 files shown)


tox.ini
```diff
@@ -13,6 +13,7 @@ env_list =
     coverage
     readme
     docs
+    3.13t
```
gaborbernat, line 16: Would we not need support in tox for this?

gaborbernat (approved):

gaborbernat (approved):

gaborbernat (changes requested):
Seems the CI is failing, can you fix it? @robsdedude Thanks

tox.ini
robsdedude: I think I found a way by using an env var that points to the python executable and using that to explicitly configure tox with `base_python` to use that Python installation. I'm working on it.

robsdedude: @gaborbernat maybe I've got something working now. Could you please re-trigger the CI?

gaborbernat:
@robsdedude seems it is still failing 🤔

robsdedude:
Now CI should pass. I made the CI run on my fork to increase the iteration velocity :upside_down_face:

gaborbernat (approved):

ngoldbaum:
Thanks so much for fixing this and seeing the PR through @robsdedude. Our team at Quansight Labs is trying to spot ecosystem issues like this around free-threaded Python. If you are aware of any other issues, we should chat!

PR #2832, opened 2025-01-24
Add GraalPy support

msimacek:
Add support for creating GraalPy virtualenvs and add GraalPy to the CI. The added code mostly mirrors PyPy3 implementation.

CC @timfel

### DIFF
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -11,6 +11,7 @@ import logging
 import os
 import platform
 import re
+import struct
 import sys
 import sysconfig
 import warnings
@@ -44,7 +45,8 @@ class PythonInfo:  # noqa: PLR0904
 
         # this is a tuple in earlier, struct later, unify to our own named tuple
         self.version_info = VersionInfo(*sys.version_info)
-        self.architecture = 64 if sys.maxsize > 2**32 else 32
+        # see https://github.com/pypa/packaging/pull/711 why we calculate architecture this way
+        self.architecture = 32 if struct.calcsize("P") == 4 else 64  # noqa: PLR2004
 
         # Used to determine some file names.
         # See `CPython3Windows.python_zip()`.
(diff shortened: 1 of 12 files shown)


src/virtualenv/discovery/py_info.py
```diff
@@ -44,7 +45,8 @@ def abs_path(v):
 
         # this is a tuple in earlier, struct later, unify to our own named tuple
         self.version_info = VersionInfo(*sys.version_info)
-        self.architecture = 64 if sys.maxsize > 2**32 else 32
+        # see https://github.com/pypa/packaging/pull/711 why we calculate architecture this way
```
gaborbernat, line 48: Explain it here rather than link it.

tests/conftest.py
```diff
@@ -355,7 +355,9 @@ def _skip_if_test_in_system(session_app_data):
         pytest.skip("test not valid if run under system")
 
 
-if IS_PYPY or (IS_WIN and sys.version_info[0:2] >= (3, 13)):  # https://github.com/adamchainz/time-machine/issues/456
+if (
+    IS_PYPY or IS_GRAALPY or (IS_WIN and sys.version_info[0:2] >= (3, 13))
+):  # https://github.com/adamchainz/time-machine/issues/456
```
gaborbernat, line 360: I think this can be now changed to just `IS_PYPY or IS_GRAALPY` as the issue has been fixed for Windows in the meantime.

tox.ini
```diff
@@ -10,6 +10,7 @@ env_list =
     3.10
     3.9
     3.8
+    graalpy
```
gaborbernat, line 13: No version constraint here? graalpy-3.13, graalpy-3.12?

gaborbernat (changes requested):
Seems the CI is failing, you'll need to address those, can check that on your forks actions, will mark it as draft until then.

tox.ini
timfel: It's just like `pypy3` (except that graalpy only targets Python 3, we don't have a Python 2 variant like PyPy does so no `3` suffix)

gaborbernat:
seems it's stalled we can reopen when you restart work on it.

PR #2848, opened 2025-02-24
Fix debug logging interpolation

tipabu:
Previously, message and args would each get interpolated separately, with a result like

```
wrote python info of %s at (PosixPath('/usr/bin/python3.12'), PosixPath('/home/...f265.json')) [DEBUG via_disk_folder:153]
```

Now, interpolate the message (which has its own formatting placeholder) first, then the args; the result is more like

```
wrote python info of /usr/bin/python3.12 at /home/...f265.json [DEBUG via_disk_folder:153]
```


gaborbernat (changes requested):
We should use logging formatting and not f-string formatting.

PR #2859, opened 2025-03-27
Add GraalPy support

timfel:
This adds support for creating GraalPy virtualenvs and add GraalPy to the CI. The added code mostly mirrors PyPy3 implementation.

This PR is picking up from https://github.com/pypa/virtualenv/pull/2832

Addressed https://github.com/pypa/virtualenv/pull/2832#discussion_r1928919497
Addressed https://github.com/pypa/virtualenv/pull/2832#discussion_r1928921433

The main addition is that we skip coverage reporting and slower tests on GraalPy, because that pushes the CI very close to the limit (had it passing in just seconds shy of the 20min limit on Ubuntu). Skipping a few slower tests and coverage reporting should make it fit comfortably inside the timeout.

/cc @msimacek

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -32,6 +32,7 @@ jobs:
           - pypy-3.10
           - pypy-3.9
           - pypy-3.8
+          - graalpy-24.1
         os:
           - ubuntu-latest
           - macos-latest
@@ -41,6 +42,7 @@ jobs:
           - { os: macos-latest, py: "brew@3.10" }
           - { os: macos-latest, py: "brew@3.9" }
         exclude:
+          - { os: windows-latest, py: "graalpy-24.1" }
           - { os: windows-latest, py: "pypy-3.10" }
           - { os: windows-latest, py: "pypy-3.9" }
           - { os: windows-latest, py: "pypy-3.8" }
--- a/docs/changelog/2832.feature.rst
+++ b/docs/changelog/2832.feature.rst
@@ -1,0 +1,1 @@
+Add support for `GraalPy <https://github.com/oracle/graalpython>`_.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ optional-dependencies.test = [
   "packaging>=23.1",
   "pytest>=7.4",
   "pytest-env>=0.8.2",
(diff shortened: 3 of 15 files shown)


gaborbernat (commented):
Have you tried implementing and using this as a plugin first?

msimacek:
Yes, we were shipping [a plugin as a part of GraalPy's distribution](https://github.com/oracle/graalpython/blob/master/graalpy_virtualenv_seeder/graalpy_virtualenv_seeder/graalpy.py#L52) for several years now. The first part of the plugin source is now exactly the same as the code submitted here. The second part adds an alternative seeder, which we're not submitting here (that part would stay as a plugin, it's not essential for basic operation).

The problems with having a plugin are:
1. If you're running virtualenv with CPython and want to install GraalPy (typical usage pattern with tox), you first need to manually install the plugin into your CPython. Without it, virtualenv just fails with `TypeError: 'NoneType' object is not callable`. A GraalPy user who has seen this for the first time would probably not think of looking for any plugin and it's not like asking google/chatgpt about such a generic error message can point them to it. And a lot of tools use virtualenv behind the scenes, so users might not even know the error comes from virtualenv. TLDR: users need to do an extra step and that step is hard to discover
2. Some other python projects are interested in adding support for GraalPy. Not having upstream support in virtualenv complicates things needlessly for them. Notably, cibuildwheel wants to support GraalPy, but integrating the plugin installation into the virtualenv backend there proved to be too complicated.

Note that uv already supports creating GraalPy virtualenvs out of the box.

gaborbernat (changes requested):
You'll need to fix the CI before we can merge 👍

henryiii:
Cross-referencing cibuildwheel PR: https://github.com/pypa/cibuildwheel/pull/1538

timfel:
> You'll need to fix the CI before we can merge 👍

I don't understand why those two builds timed out - they seem to just have hung? I ran the check in our fork and it just passed (as I would expect): https://github.com/msimacek/virtualenv/actions/runs/14131798860

Can you just restart the two jobs to double check?

timfel:
> You'll need to fix the CI before we can merge 👍

After the restart it passed :-)

gaborbernat (approved):

PR #2868, opened 2025-04-10
Stop including 'wheel', setuptools 70.1 has native bdist_wheel support

stefanor:
setuptools has native `bdist_wheel` support these days, so I don't think it's necessary to bundle the `wheel` library any more. Or am I missing something?

Possibly we could include a noop `--no-wheel` argument (but I'm not sure where that would be best defined).


tests/unit/create/test_creator.py
```diff
@@ -398,6 +398,7 @@ def test_create_long_path(tmp_path):
     subprocess.check_call([str(result.creator.script("pip")), "--version"])
 
 
+@pytest.mark.skip("Depends on https://github.com/pypa/pip/pull/13330")
```
stefanor, line 401: We probably don't want to land this PR until pip has released with this change?

tests/unit/seed/embed/test_bootstrap_link_via_app_data.py
```diff
@@ -224,6 +224,7 @@ def test_app_data_parallel_ok(tmp_path):
     assert not exceptions, "\n".join(exceptions)
 
 
+@pytest.mark.skip("Fails for unknown reasons, without wheel")
```
stefanor, line 227: I could use help understanding this failure. (Or really what the test was testing before the change)

gaborbernat: Yeah, we need to figure our this before we can merge.

stefanor:
The two blockers are resolved. Note that this PR now bumps the bundled pip (except for Python 3.8, which I assume we'll remove shortly).

src/virtualenv/seed/wheels/embed/__init__.py
```diff
@@ -9,37 +9,30 @@
     "3.8": {
         "pip": "pip-25.0.1-py3-none-any.whl",
         "setuptools": "setuptools-75.3.2-py3-none-any.whl",
-        "wheel": "wheel-0.45.1-py3-none-any.whl",
     },
```
stefanor, line 12: Will go away (and the old pip can be deleted) with #2874.

gaborbernat (changes requested):
PR still failing.

tests/unit/create/test_creator.py
```diff
@@ -398,6 +398,10 @@ def test_create_long_path(tmp_path):
     subprocess.check_call([str(result.creator.script("pip")), "--version"])
 
 
+@pytest.mark.skipif(
+    sys.version_info[:2] == (3, 8),
+    reason="Disable on Python 3.8, which still uses pip 25.0.1 (without https://github.com/pypa/pip/pull/13330)",
+)
```
stefanor, line 404: Tests are passing due to skipping this on 3.8.

We could continue to bundle `wheel` only on 3.8, instead? That would defer some of the cleanup here until after 3.8 goes out of support in virtualenv.

gaborbernat: Yes.

gaborbernat (approved):

src/virtualenv/seed/embed/base_embed.py
```diff
@@ -18,14 +20,21 @@ def __init__(self, options) -> None:
...
         self.no_setuptools = options.no_setuptools
-        self.no_wheel = options.no_wheel
         self.app_data = options.app_data
         self.periodic_update = not options.no_periodic_update
 
+        if options.no_wheel:
+            warn(
+                "The --no-wheel option is deprecated. "
+                "It has no effect, wheel is no longer bundled in virtualenv. "
+                "This option will be removed in pip 26.",
```
hroncok, line 33: in pip 26? should this be a virtualenv version?

gaborbernat: Yeah, it should 🤔 @stefanor

gaborbernat: But I think here we're conflating virtualenv wheel support and pip wheel support 🤔

stefanor: #2876

henryiii:
~~By the way, I'm now getting dozens of errors like this (I have warnings as errors turned on in tests):~~

```
DeprecationWarning: The --no-wheel option is deprecated. It has no effect for Python >= 3.8 as wheel is no longer bundled in virtualenv.
```

Ahh, I'm actually adding `--no-wheel`, that's why! Nevermind. Is there an ETA when this will be removed? Just trying to decide if I should conditionally add it or ignore the error for now. I test the lower bound of dependencies.

gaborbernat:
April next year.

hroncok:
May I suggest an alternative approach? What if it was never removed?

Keeping a dull no-action option is not a maintenance burden.

Removing its usage from countless places on the other end seems not worth it.

JuroOravec:
Hi, I have a suspicion this broke the [airspeed-velocity](https://github.com/airspeed-velocity/asv) library, because there they used the `wheel=bundle` CLI argument.

See https://github.com/airspeed-velocity/asv/issues/1484

How to fix this? I skimmed through the changelog, but not 100% sure what to do to make the fix backwards compatible:
- Should I replace `wheel=bundle` with `VIRTUALENV_WHEEL=bundle`?
- Or should I remove `wheel=bundle` when on Python >=3.12?

hroncok:
I think you should remove wheel=bundle on Python 3.8+. What is your motivation to use it? The general idea is that it should not be needed with the updated setuptools and pip.

(However I relize that it's a distruption, perhaps it should warn but not fail?)

allanlewis:
> (However I relize that it's a distruption, perhaps it should warn but not fail?)

This is breaking our builds using Poetry 1.8.5, which I don't expect the Poetry team to update for this breaking change. We'll update to Poetry 2.x at some point but I'd rather not be forced to by this change.

JuroOravec:
I managed to fix it by omitting the `wheel=bundle` for Py3.12+. But @hroncok after seeing your comment I will update it to use py3.8 as cutoff.

The problematic code is here: https://github.com/airspeed-velocity/asv/blob/01f7cfa3da86f6d98ef68984ecc663ac6ff7dfaa/asv/plugins/virtualenv.py#L136

And here is how I fixed it:

```py
util.check_call([
    sys.executable,
    "-mvirtualenv",
    # Omit `--wheel=bundle` from py3.9+
    # See https://github.com/airspeed-velocity/asv/issues/1484
    *(["--wheel=bundle"] if sys.version_info < (3, 9) else []),
    "--setuptools=bundle",
    "-p",
    self._executable,
    self._path], env=env)
```

As for *why* that command is called with those arguments, I don't know. I'm maintainer at [django-components](https://github.com/django-components/django-components) and went to report / complain at [airspeed-velocity](https://github.com/airspeed-velocity/asv), which brought me here :smile:

allanlewis:
I think this is breaking any Poetry project when at least one dependency needs a wheel to be built from source: see [python-poetry/poetry#10378 (comment)](https://github.com/python-poetry/poetry/issues/10378#issuecomment-2853932239)
I expect the same to be the case with other package managers that use `virtualenv` under-the-hood.

PidgeyBE:
It's indeed a breaking change for any project (poetry and others) having the `--wheel` argument still around in older versions.
I think removing the `--wheel` argument (even if it's a noop) justifies a major bump :grimacing:

hroncok:
> I think a solution for the poetry problem (and similar) might be: On Python 3.8, `--wheel=bundle` works as before. On newer Pythons, `--wheel=bundle` does nothing and DeprecateWarns.

I was looking into this, but fundamentally at:

https://github.com/pypa/virtualenv/blob/e9b6ba728f44ffb84ca44a40b58508c3dd5f7b29/src/virtualenv/seed/embed/base_embed.py#L23-L34

We would have all the options, and I am not entirely sure how to check `interpreter.version_info` from this method.

hroncok:
This is my draft PR (which doesn't work, as I am stuck). https://github.com/pypa/virtualenv/pull/2884

stefanor:
>  I think removing the --wheel argument (even if it's a noop) justifies a major bump 😬

Note that there have been no major bumps since the rewrite. I guess because behavioural changes have been keyed to Python versions.

hroncok:
> What got us into this pickle was `wheel` no longer being bundle-able, so we had to remove it immediately without a deprecation period.

Well, the other option was to keep the last known good version of wheel. (The new one was yanked BTW, so we might just as well do that for now.)

src/virtualenv/seed/embed/base_embed.py
```diff
@@ -18,14 +20,21 @@ def __init__(self, options) -> None:
...
-        self.wheel_version = options.wheel
 
         self.no_pip = options.no_pip
         self.no_setuptools = options.no_setuptools
-        self.no_wheel = options.no_wheel
         self.app_data = options.app_data
         self.periodic_update = not options.no_periodic_update
 
+        if options.no_wheel:
+            warn(
```
hroncok, line 30: @stefanor I would also like to point out that as an user, I did not see this warning when I used the option.

black-snow:
What's the conclusion here? Can I just wait for my poetry 1.x pipelines to suddenly work again or do I need to take action?

PR #2876, opened 2025-04-29
Revert a large part of the wheel removal, to support Python 3.8

stefanor:
This started as a cleanup to a whoopsie in #2868 (thanks @hroncok).

Revert many parts of our wheel removal to support Python 3.8. Pip 25.1 (which understands that wheel is not needed) is not supported on Python 3.8, so we continue to ship pip 25.0.1 for Python 3.8, and provide wheel.

This contains a partial revert of commit db1d9fc02739d6388192eefb6befa9e4eccbe630. Once support for Python 3.8 has been removed, it can be reverted again.


src/virtualenv/seed/embed/base_embed.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, options) -> None:
             warn(
                 "The --no-wheel option is deprecated. "
                 "It has no effect, wheel is no longer bundled in virtualenv. "
-                "This option will be removed in pip 26.",
+                "This option will be removed in virtualenv 26.",
```
gaborbernat, line 33: Do we ever need to remove it? There's no such thing as virtualenv 26. What are you trying to do here? Should we just remove this flag?

stefanor: Oh 🤦

Yeah, we don't need to ever remove it.

src/virtualenv/seed/embed/base_embed.py
```diff
@@ -28,9 +28,7 @@ def __init__(self, options) -> None:
 
         if options.no_wheel:
             warn(
-                "The --no-wheel option is deprecated. "
-                "It has no effect, wheel is no longer bundled in virtualenv. "
-                "This option will be removed in pip 26.",
+                "The --no-wheel option is deprecated. It has no effect, wheel is no longer bundled in virtualenv.",
```
gaborbernat, line 31: When the user is running with 3.8 it should still work though no?

stefanor: It has no effect with 3.8 too. We removed all the code to ever bundle `wheel`.

stefanor: We removed `wheel` from `src/virtualenv/seed/wheels/embed/`, so it won't be bundled.

And setuptools doesn't need it, but `pip --no-use-pep517` won't work, because it requires `wheel` to enable legacy mode.

hroncok: So, does this mean we either:

 - add wheel back to 3.8
 - remove support for 3.8 envs

?

gaborbernat: yeah we need to add it back because we didn't drop the support for 3.8 it just

stefanor: OK, let me do that.

stefanor: Done. It brings a bit of unavoidable mess, but trivial to revert later.

gaborbernat (approved):

PR #2884, opened 2025-05-06
Reintroduce the --wheel CLI option, even though it has no effect on Python > 3.8

hroncok:


src/virtualenv/seed/embed/base_embed.py
```diff
@@ -20,21 +20,23 @@ def __init__(self, options) -> None:
...
-            self.wheel_version = options.wheel
-            self.no_wheel = options.no_wheel
-        elif options.no_wheel:
+        self.wheel_version = options.wheel
+
+        self.no_pip = options.no_pip
+        self.no_setuptools = options.no_setuptools
+        self.no_wheel = options.no_wheel
+
+        if options.no_wheel or options.wheel:  # and interpreter.version_info[:2] >= (3, 9)
```
hroncok, line 29: This is where I am stuck. I want to check the interpreter version, but this method does not know the interpreter.

The possible but very hacky thing is to add a hidden parser argument `--wheel-is-deprecated` with a default value based on interpreter version and read it here. But that's not nice.

stefanor: Yeah, I hit the same problem when I implemented this. I was rather pleased with the `hasattr` hack, but indeed the whole API needs a refactor to make this work...

gaborbernat: Yeah we need to pass it through, however maybe this should not be inside the init 😊 but rather inside the run within https://github.com/hroncok/virtualenv/blob/wheel-option-back/src/virtualenv/seed/seeder.py#L31

hroncok: Does the creatoer there hold a reference to interpreter?

stefanor: Or we support all these options here (altering CLI options based on another CLI option (`-p`) is already weird), and make the decision on whether to do something or nothing later in the stack.

gaborbernat: > Does the creatoer there hold a reference to interpreter?

Yes.

src/virtualenv/seed/embed/pip_invoke.py
```diff
@@ -18,8 +18,8 @@ def __init__(self, options) -> None:
     def run(self, creator):
         if not self.enabled:
             return
-        for_py_version = creator.interpreter.version_release_str
-        with self.get_pip_install_cmd(creator.exe, for_py_version) as cmd:
+        self.warn_wheel_options(creator.interpreter)
```
gaborbernat, line 21: Hmm, I don't like how this ended up. Perhpas better to attach the interpreter version information onto the `options` [here](https://github.com/pypa/virtualenv/blob/main/src/virtualenv/run/__init__.py#L51):

```python
options.py_version = parser._interpreter.version_info  # noqa: SLF001
```

And keep all the logic inside the `__init__` as was before this PR 🤔

gaborbernat (changes requested):

src/virtualenv/seed/embed/pip_invoke.py
hroncok: Ok then, will do.

hroncok: See what I pushed. It does not work:

```
AttributeError: 'VirtualEnvOptions' object has no attribute 'py_version'
```

hroncok: Never mind, I got it.

hroncok:
For the record, based on https://github.com/tox-dev/tox/pull/3528 I now see that this also warns when $VIRTUALENV_WHEEL is set.

This was not my intention. Should we keep it that way? And if so, shoudl the warning be adjusted to mention it?

PR #2911, opened 2025-07-26
test: Use `@pytest.mark.flaky` instead of `@flaky.flaky`

mgorny:
Use the more modern `@pytest.mark.flaky` decorator to mark tests as flaky, rather than `@flaky.flaky`.  This avoids type checking issues, and improves compatibility with other implementations of the marker, such as `pytest-rerunfailures`.

-----

I think this doesn't justify a news item, since it won't affect end users — but I can write one if you wish.

### DIFF
--- a/tests/integration/test_zipapp.py
+++ b/tests/integration/test_zipapp.py
@@ -6,7 +6,6 @@ from contextlib import suppress
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

mgorny:
Thanks a lot!

PR #2920, opened 2025-08-01
fix: Update venv redirector detection for Python 3.13 on Windows

### LINKED ISSUE #2851: venv redirector failing to discover on Python 3.13
To support free-threaded builds, the Windows venv redirector executables have been renamed in Python 3.13.
The executables were under `Lib/venv/scripts/nt` and were named `python.exe` and `pythonw.exe`. They have been renamed to `venvlauncher.exe` and `venvwlauncher.exe` respectively. This causes virtualenv not to find the shim and to copy DLLs, etc., into the virtual environment.
There may be other issues because of this change.
In the meantime, a possible workaround could be to create file symbolic links.

References:
- https://github.com/python/cpython/issues/112984
- https://github.com/python/cpython/pull/113129
- https://discuss.python.org/t/when-should-venv-scripts-nt-python-exe-be-present/47620

esafak:
* Detect the new `venvlauncher.exe` and `venvwlauncher.exe` on Windows for Python 3.13 and later.
* Previously, virtualenv expected `python.exe` which is no longer the case on newer Python versions.
* This ensures that virtualenv can correctly create environments using the system Python interpreter on Windows for Python 3.13+.

Fixes #2851

### DIFF
--- a/docs/changelog/2851.bugfix.rst
+++ b/docs/changelog/2851.bugfix.rst
@@ -1,0 +1,2 @@
+Support renamed Windows venv redirector (`venvlauncher.exe` and `venvwlauncher.exe`) on Python 3.13
+Contributed by :user:`esafak`.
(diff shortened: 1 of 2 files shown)


gaborbernat:
Marked this as a draft for now 🚧—let’s get the CI sorted, and then feel free to mark it ready for review! 😊

PR #2925, opened 2025-08-01
fix cache invalidation for PythonInfo

esafak:
The cache for PythonInfo is now invalidated if the PythonInfo class changes. I achieved this by hashing the py_info.py file and storing the hash in the cache. When loading from the cache, the hash is recomputed and compared with the stored hash. If they don't match, the cache is invalidated.

Fixes #2647

### DIFF
--- a/docs/changelog/2467.bugfix.rst
+++ b/docs/changelog/2467.bugfix.rst
@@ -1,0 +1,2 @@
+Fix cache invalidation for PythonInfo by hashing `py_info.py`.
+Contributed by :user:`esafak`.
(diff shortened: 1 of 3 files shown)


gaborbernat (dismissed):

gaborbernat:
```
            # 6. Assert that _run_subprocess was called again
>           assert spy.call_count == 2
E           assert 3 == 2
E            +  where 3 = <function _run_subprocess at 0x1031e74c0>.call_count

```

Seems this is a valid error, you'll need to fix it.

esafak:
That linter rule is dumb. It will be happier if I completely remove the comment leaving everybody in the dark. Should I remove it?

gaborbernat:
@esafak you need to fix the pre commit check 😊

gaborbernat (approved):

PR #2926, opened 2025-08-01
fix(lint): Disable FIX002 to allow TODO comments

esafak:
The linter was configured to flag all TODO comments, which can discourage developers from leaving them. This change disables the FIX002 rule in ruff to allow TODO comments without triggering a linting error.


esafak:
OK I won't leave them, but I won't open tickets for every little issue either!

gaborbernat:
That's fine, if something is not important, to do it right there and is not big enough to have an issue, it's likely a case of you ain't gonna need it. People don't tend to go back and clean up all this todo comments, so hence why I think they're a bit of a noise rather than value.

esafak:
People don't tend to go and fix things full stop. We should only be concerned with people like me who are. 

I hopefully fixed a flaky test in #2923 today simply because I came across it. It was not even a TODO, let alone an issue.

gaborbernat:
That's great and much appreciated. However, if you go and fix it, then there's no need for either to do or an issue.😁

Issue #2074, opened 2021-03-01
Extract discovery to it's own package


gaborbernat:
From https://github.com/pipxproject/pipx/issues/638 @uranusjr
 
> I love virtualenv.discovery, but virtualenv is a big thing to pull in. I would reach for it every time if the functionality could be extracted into its own distribution.

gaborbernat:
@pfmoore may we get the ``py-discovery`` package under PyPa? If so can you create a repo for it and make me admin of it?

pfmoore:
The usual process is that you create a separate package under your own name, then propose it for PyPA inclusion on Discourse, and then it gets moved to the PyPA namespace. See [the process here](https://www.pypa.io/en/latest/members/#project-membership).

I don't see any reason why this shouldn't be straightforward - it sounds like a good idea to me.

gaborbernat:
That's usually the case with new projects, but in the case of splitting the existing ones, I've thought we don't need the intermediate step.

pfmoore:
Honestly, I don't know - the process in PEP 609 was decided by the PyPA with @di as the PEP author (the discussion was [here](https://discuss.python.org/t/pep-609-pypa-governance/2619)). I don't think I have the authority to modify it - my role is as PEP delegate for interoperability PEPs, not PyPA BDFL.

I agree that splitting out part of an existing project should be straightforward - the process should be just a formality.

uranusjr:
Bringing this to the technical side of things, what dependencies are needed for the discovery logic?

gaborbernat:
Can be made no dependency at all :-) Depends on how much we extract, and how much we leave behind.

gaborbernat:
Seemns there's wasn't much request for this, so closing as not planned.

dukecat0:
Maybe I can try to extract the logic into a separate package first?

gaborbernat:
Sure 👍 feel free to add me as a reviewer.

dukecat0:
Should the cache part be removed? I personally think that it can be implemented in virtualenv.

gaborbernat:
Removed, not, but doesn't need to be part of the extracted package 👍 Though the cache is very useful could be an extra at a future time.

gaborbernat:
Started at https://github.com/tox-dev/py-discovery, will try to get something working by Monday 😊

gaborbernat:
[...]

PR #2949, opened 2025-08-11
refactor: Decouple discovery from creator plugins

esafak:
The PythonInfo class had a method called creators() that directly imported and used CreatorSelector from `virtualenv.run.plugin.creators`. This created a tight coupling between the discovery and creator plugin systems.

This commit removes the creators() method from PythonInfo, making it a pure data class. The logic for selecting a creator has been moved to the call sites, which now directly use `CreatorSelector.for_interpreter(python_info)`.

This change prepares the discovery code to be moved into a new, standalone package (#2074).

### DIFF
--- a/src/virtualenv/discovery/py_info.py
+++ b/src/virtualenv/discovery/py_info.py
@@ -30,7 +30,7 @@ EXTENSIONS = _get_path_extensions()
 _CONF_VAR_RE = re.compile(r"\{\w+\}")
 
 
-class PythonInfo:  # noqa: PLR0904
+class PythonInfo:
     """Contains information for a Python interpreter."""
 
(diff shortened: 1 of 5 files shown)


src/virtualenv/discovery/py_info.py
```diff
@@ -467,8 +459,7 @@ def _to_json(self):
         return json.dumps(self._to_dict(), indent=2)
 
     def _to_dict(self):
-        data = {var: (getattr(self, var) if var != "_creators" else None) for var in vars(self)}
-
+        data = {var: getattr(self, var) for var in vars(self)}
```
gaborbernat, line 462: This now is just vars(self)?

PR #2956, opened 2025-08-13
refactor: Decouple discovery module

esafak:
After this refactor it should be possible to extract `discovery` into its own package (#2074). There is some coupling in tests, but this can be considered integration testing.

The primary dependencies on `virtualenv.info`, `virtualenv.app_data`, `virtualenv.cache`, and `virtualenv.util` have been removed. Instead of direct imports, these dependencies are now injected into the discovery components.

This was achieved by:
* Introducing `AppData` and `Cache` protocols within the `discovery` module to act as abstract interfaces.
* Modifying the functions in `cached_py_info.py` and `builtin.py` to accept `app_data` and `cache` objects as arguments.
* Plumbing these new arguments down from the main entry points (e.g., `virtualenv.run`).
* Refactoring the test suite extensively to provide the necessary mock or real objects to the refactored functions, ensuring all tests pass.


src/virtualenv/discovery/cache.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from typing_extensions import Protocol
+else:
+    Protocol = object
```
gaborbernat, line 10: Note like this; this should be something like:

```
if sys.version_info >= (3, 10):
   from typing import Protocol
else:
   from typing_extensions import Protocol
```

with appropiate covdefaults pragma no covers.

gaborbernat (changes requested):

src/virtualenv/discovery/cache.py
esafak: actually I'll remove all that since Protocol shipped in 3.8, ok?

gaborbernat (approved):

PR #2958, opened 2025-08-14
feat: added support for `pkg-config`

esafak:
To do this, I modified the activation scripts for all supported shells. Now, when you activate a virtual environment, the `lib/pkgconfig` directory will be prepended to the `PKG_CONFIG_PATH` environment variable. The original value is restored upon deactivation.

I also added unit tests for each shell to verify that the activation scripts are generated correctly and that `PKG_CONFIG_PATH` is set and unset as expected.


gaborbernat:
Seems stalled we can reopen if you pick it up again.

PR #2996, opened 2025-11-25
fix: wrong path on migrated venv

sk1234567891:


pfmoore:
Is there an issue for this? It's hard to see how `__VIRTUAL_ENV__` could be incorrect here so an issue with a demonstration of how to reproduce the problem would be helpful.

Also, this would presumably need to be fixed the same way in *all* the activate scripts, not just the bash one?

gaborbernat:
Released via https://github.com/pypa/virtualenv/releases/tag/20.36.0

PR #3006, opened 2026-01-02
Fix Interpreter discovery bug wrt. Microsoft Store shortcut using Latin-1

rahuldevikar:
When virtualenv performs interpreter discovery on Windows, it spawns a subprocess to run the py_info.py script for each discovered Python interpreter. On systems where the Microsoft Store Python shortcut exists (but Python wasn't installed via MS Store), invoking this shortcut returns an error message in CP1252 encoding. On systems with non-English locales (e.g., German), these error messages contain characters like umlauts (ü) that can't be decoded as UTF-8, causing UnicodeDecodeError exceptions to appear in the terminal.

Use Latin-1 instead

update:
Forcing subprocess to use utf-8 mode

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -55,6 +55,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: 🏷️ Fetch upstream tags for versioning
+        shell: bash
+        run: |
+          git fetch --force --tags https://github.com/pypa/virtualenv.git
       - name: 🐍 Setup Python for tox
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


src/virtualenv/discovery/cached_py_info.py
```diff
@@ -134,7 +134,7 @@ def _run_subprocess(cls, exe, app_data, env):
                 stderr=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 env=env,
-                encoding="utf-8",
+                encoding="latin-1",
```
gaborbernat: Isn't a better solution to force it to run in utf-8 mode than doing this?

rahuldevikar:
All tests are succeeding here: https://github.com/rahuldevikar/virtualenv/actions/runs/20727523037
Note, for some reason, the pipeline is using tox 1.1(and failing) and I have to hard push to use >= 4.32 as shown in the check.yml. It seems that the original repo is fetching the tox >= 4.32 somehow. I think the changes to check.yml should be safe enough but I need feedback on how can I make sure to keep the forked repo pipeline run in sync with original repo's pipeline.

Cc: @gaborbernat (Sorry for tagging, do let me know if there is a better way to communicate on such issues)

gaborbernat:
Released via https://github.com/pypa/virtualenv/releases/tag/20.36.0

repairman29:
🎯 **PING RECEIVED!**

**@repairman29** has claimed this bounty via [Echeo](https://echeo.io).

**Match Score:** 85%  
**Status:** Ready to start working!

[View conversation in Echeo](https://echeo.io/messages) to coordinate.

Issue #3037, opened 2026-02-16
Incorrect dependency range specified for python 3.9 (e.g., macOS xcode python3.9)


reactive-firewall:
**Issue**

Describe what's the expected behavior and what you're observing.

> With Python 3.9, it attempts to install the latest version of virtualenv (20.36.1), and in turn to install filelock 3.20.1, which fails of course, because filelock ≥ 3.20.1 is not compatible with Python 3.9.

_Originally posted by @DimitriPapadopoulos in https://github.com/raimon49/pip-licenses/issues/299#issuecomment-3904899000_

> And of course the bug is in `virtualenv` as you mentioned:
> 
>  ```toml
>    "filelock>=3.16.1,<4; python_version<'3.10'",
>    "filelock>=3.20.1,<4; python_version>='3.10'",
>   ```
>   
>   is obviously a mistake as it should be (see [PEP-440](https://peps.python.org/pep-0440/))
> 
>   ```
>    "filelock>=3.16.1,<3.20.1; python_version<'3.10'",
>    "filelock>=3.20.1,<4; python_version>='3.10'",
>   ```

Originally posted by @reactive-firewall in https://github.com/raimon49/pip-licenses/pull/299#issuecomment-3905815689_

**Environment**

Provide at least:

- OS: MacOS (probably others)
- Shell: `bash`
- Python version and path: `python3.9.6` `/Applications/Xcode.app/Contents/Developer/usr/bin/python3` (probably others)
- `pip list` of the host python where `virtualenv` is installed:

```console
$ python3 -B -m pip list -vvv
Package           Version Location                                                                                                                 Installer
----------------- ------- ------------------------------------------------------------------------------------------------------------------------ ---------
altgraph          0.17.2  /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/site-packages pip
future            0.18.2  /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/site-packages pip
macholib          1.15.2  /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/site-packages pip
pip               26.0.1  /Users/user/Library/Python/3.9/lib/python/site-packages                                                                  pip
[...]

PR #3038, opened 2026-02-16
Python3.9 dependency range correction

reactive-firewall:
This change is to fix GHI pypa/virtualenv#3037

Just corrects the upper limit of pre-`3.10` Python `filelock` to the version `3.19.1`

Should be transparent to users of Python `3.10`+

Refs:
* https://github.com/pypa/virtualenv/issues/3037
* https://github.com/pypa/virtualenv/blob/9719376addaa710b61d9ed013774fa26f6224b4e/pyproject.toml#L48-L49
* [development documentation](https://virtualenv.pypa.io/en/latest/development.html#development))
* https://pypi.org/project/filelock/3.19.1/ (see trove classifier `python :: 3.9` present)
* https://pypi.org/project/filelock/3.20.0/ (see trove classifier `python :: 3.9` removed)

### DIFF
--- a/src/virtualenv/activation/bash/activate.sh
+++ b/src/virtualenv/activation/bash/activate.sh
@@ -11,29 +11,28 @@ deactivate () {
     unset -f pydoc >/dev/null 2>&1 || true
 
     # reset old environment variables
-    # ! [ -z ${VAR+_} ] returns true if VAR is declared at all
-    if ! [ -z "${_OLD_VIRTUAL_PATH:+_}" ] ; then
+    if [ -n "${_OLD_VIRTUAL_PATH:-}" ] ; then
         PATH="$_OLD_VIRTUAL_PATH"
         export PATH
         unset _OLD_VIRTUAL_PATH
     fi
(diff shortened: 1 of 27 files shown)


pyproject.toml
```diff
@@ -45,7 +45,7 @@ dynamic = [
 ]
 dependencies = [
   "distlib>=0.3.7,<1",
-  "filelock>=3.16.1,<4; python_version<'3.10'",
+  "filelock>=3.16.1,<=3.19.1; python_version<'3.10'",
```
rahuldevikar: There is a filelock version == 3.20.**0** Should we include that too in one or the other condition?
https://pypi.org/project/filelock/#history

gaborbernat (approved):

pyproject.toml
rahuldevikar: Make sense. Sounds good. Thanks

reactive-firewall:
@gaborbernat I saw your merge on main and rebased this PR so both this fix and the discussed version bump for python `3.10`+ would be based off the same point along `main`, yet can be tested, merged separately.

:bow: hopefully this is helpful, and not an additional maintenance burden.

PR #3057, opened 2026-02-19
Add auto-upgrade workflow for embedded dependencies

rahuldevikar:
Add a daily GitHub Actions workflow that:
- Runs tox -e upgrade to check for new pip/setuptools/wheel releases
- Automatically creates a PR with updated embedded wheels when new versions are detected
- Includes a job summary showing what changed
- Restricted to pypa org to avoid running on forks

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -42,6 +42,9 @@ jobs:
           - macos-15
           - windows-2025
         include:
+          - { os: macos-15, py: "brew@3.14" }
+          - { os: macos-15, py: "brew@3.13" }
+          - { os: macos-15, py: "brew@3.12" }
           - { os: macos-15, py: "brew@3.11" }
           - { os: macos-15, py: "brew@3.10" }
           - { os: macos-15, py: "brew@3.9" }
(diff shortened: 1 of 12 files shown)


.github/workflows/upgrade.yaml
```diff
@@ -0,0 +1,61 @@
...
+          commit-message: "Upgrade embedded dependencies"
+          branch: auto/upgrade-embedded-deps
+          delete-branch: true
+          title: "Upgrade embedded pip/setuptools/wheel"
+          body: |
+            Automated upgrade of embedded pip, setuptools, and wheel dependencies.
+
+            This PR was created automatically by the [upgrade workflow](https://github.com/${{ github.repository }}/actions/workflows/upgrade.yaml).
+
+            > [!NOTE]
```
gaborbernat: We could setup a PAT to allow it to run and go through 🤔

gaborbernat (changes requested):
We need to implement a slight delay window here by default to follow our delay policy: https://virtualenv.pypa.io/en/latest/explanation.html#periodic-update-mechanism

Generally we wait at least a week before we pull in a new wheel as embedded.

.github/workflows/upgrade.yaml
gaborbernat: Could be also interesting to see if we can use https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys here 🤔

PR #3036, opened 2026-02-15
✨ feat(create): sync with upstream CPython/PyPy venv

gaborbernat:
virtualenv's builtin creators replicate what CPython's `Lib/venv/` and PyPy's `lib-python/3/venv/` do, but have drifted since ~2020. This PR brings them back in sync by addressing the most impactful gaps identified from reviewing ~84 upstream CPython commits and PyPy's py3.9-3.11 evolution.

The highest-priority changes focus on correctness: `pyvenv.cfg` now includes the `executable` key (CPython 3.11+) that tools like `uv` and IDEs rely on, the `include` directory is created per PEP 405, and free-threaded Python 3.13+ builds get proper `python3.Xt` / `pythonw3.Xt.exe` executable names plus `venvlaunchert.exe` launcher handling on Windows. PyPy environments also improve — `PYPY_PORTABLE_DEPS.txt` is respected for portable builds, and Windows DLL coverage expands beyond the previous narrow `libpypy*`/`libffi*` globs.

Activation scripts receive targeted fixes: bash now uses `uname`-based detection to catch MinGW/Git Bash (matching upstream), fish preserves `$status` across the prompt function and drops Fish <3 compatibility, and PowerShell gains `Get-PyVenvConfig` with `-VenvDir`/`-Prompt` parameters. All `major == 3` guards are removed since virtualenv requires Python 3.8+, and the single-use `resign` helper is inlined at its call site. The `venv` creator also learns to pass `--without-scm-ignore-files` on 3.13+ to avoid conflicts with virtualenv's own `.gitignore` handling.

Resolves #3021
Resolves #2090

PR #3161, opened 2026-06-07
Fish activator passes VIRTUAL_ENV through cygpath

### LINKED ISSUE #3160: Activation in fish inside msys2 (or similar) breaks PATH
**Issue**

When running `source .venv/Scripts/activate.fish` on windows from a fish shell inside msys2 ucrt64 it breaks the PATH. The template does https://github.com/pypa/virtualenv/blob/a423028aa00bccb071efb72469f4dc0f05b27cfa/src/virtualenv/activation/fish/activate.fish#L63
which inside MSYS2, cygwin and MinGW will result in a mismatched (unix vs windows style) absolute path. This was already addressed for bash in https://github.com/pypa/virtualenv/pull/1970 but the fish template stayed unchanged.

**Environment**

Provide at least:

- OS: Windows 11
- Shell: Fish (run inside MSYS2 UCRT64)
- Python version and path: bug concerns .fish script file and thus unrelated to python version or path
- `pip list` of the host python where `virtualenv` is installed: bug concerns .fish script file and thus unrelated to python version or path

The creation works fine, the template itself is broken.

LuNoX:
Closes #3160. Sourcing activate.fish from fish run inside MSYS2 (or similar) breaks the PATH because of a path style mismatch. This PR applies the changes from #1970 as well as the change to using uname from #3036 to the fish template as well.


LuNoX:
Concerning the PR checklist:
- As far as I understand the linter is for the python style correct? I only touched the fish script so should I still run the linter?
- For test coverage, I would appreciate some guidance as I'm not sure how to expand the fish activation unit tests. They currently do skip the test if the system IS_WIN. Do you want a completely new test, or just remove the skip, or add new test thats covering for MSYS2, Cygwin and MinGW fish shells?
- As for docs, I don't believe anything needs to be updated, as the change brings the repo in line with what the docs promise (activating inside fish regardless of Unix or windows just works)

gaborbernat (approved):
LGTM

PR #3162, opened 2026-06-08
Stop exporting PS1 in bash activator

### LINKED ISSUE #3158: bash activate.sh exports PS1
`src/virtualenv/activation/bash/activate.sh` exports `PS1` on activation and in `deactivate` ([L117-119](https://github.com/pypa/virtualenv/blob/230934ce669a1cfccd0697d1c4a2d88b2405e0c8/src/virtualenv/activation/bash/activate.sh#L117-L119), [L46-49](https://github.com/pypa/virtualenv/blob/230934ce669a1cfccd0697d1c4a2d88b2405e0c8/src/virtualenv/activation/bash/activate.sh#L46-L49)). `PS1` is a shell-local parameter; exporting it leaks the prompt into child processes, where another shell family renders it wrong (a bash `PS1` keeps the literal `\u` in a zsh child) and it persists after `deactivate`. Removing the two `export PS1` lines is enough — the sourcing shell's prompt is unaffected since assignment alone drives it.

Same issue as CPython venv: python/cpython#111501, fix in python/cpython#105279. Filing to keep the two aligned.

karlhillx:
Closes #3158.

`PS1` is shell-local prompt state. The bash activator now updates it without exporting it, so child processes do not inherit prompt text during activation or after deactivation.

Tests:
- `.tox/3.14/bin/python -m pytest tests/unit/activation/test_bash.py -q`

### DIFF
--- a/docs/changelog/3158.bugfix.rst
+++ b/docs/changelog/3158.bugfix.rst
@@ -1,0 +1,1 @@
+Stop exporting ``PS1`` from the bash activator so child processes do not inherit shell prompt state.
--- a/src/virtualenv/activation/bash/activate.sh
+++ b/src/virtualenv/activation/bash/activate.sh
@@ -45,7 +45,6 @@ deactivate () {
 
     if [ -n "${_OLD_VIRTUAL_PS1:-}" ] ; then
         PS1="$_OLD_VIRTUAL_PS1"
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):
LGTM

PR #3175, opened 2026-06-17
🐛 fix(seed): sync wheel regen script with seeder guard

gaborbernat:
The unsupported-target refusal in `#3173` taught the bundled seeder to decline Pythons below the oldest bundle, and that policy lives in `src/virtualenv/seed/wheels/embed/__init__.py` through `MIN`, `OLDEST_SUPPORTED`, the `_release_tuple` helper, and a `get_embed_wheel` that returns `None` instead of substituting the newest wheel. 🐛 That module is generated, though: `tasks/upgrade_wheels.py` rewrites it from a template on every wheel bump, and the template still emitted the old `get_embed_wheel`. The next routine wheel upgrade would have regenerated the file and silently reverted the guard, letting virtualenv seed dropped Python versions again.

This brings the `render_init` template in line with the committed module, so `MIN`, `OLDEST_SUPPORTED`, `_release_tuple`, and the below-oldest fallback removal survive regeneration. ✨ Running `tasks/upgrade_wheels.py --regen` now reproduces the current `__init__.py` byte-for-byte.

Reviewers can confirm by running `python tasks/upgrade_wheels.py --regen` against a clean tree and checking that `src/virtualenv/seed/wheels/embed/__init__.py` is unchanged.

### DIFF
--- a/tasks/upgrade_wheels.py
+++ b/tasks/upgrade_wheels.py
@@ -124,6 +124,15 @@ def render_init(folders: dict[Path, str] | None = None) -> None:
     BUNDLE_FOLDER = Path(__file__).absolute().parent
     BUNDLE_SUPPORT = {{ {bundle} }}
     MAX = next(reversed(BUNDLE_SUPPORT))
+    MIN = next(iter(BUNDLE_SUPPORT))
+
+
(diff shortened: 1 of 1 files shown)


albertoblue87-netizen (commented):
Done

PR #3193, opened 2026-07-21
Add `python-version` to `pyvenv.cfg` for PEP 838

konstin:
Draft PR accompanying [PEP 838](https://peps.python.org/pep-0838/)


gaborbernat:
Seems stalled and outdated, we can re-open when becomes again active.

PR #3217, opened 2026-08-20
🐛 fix(create): check every execute bit in ExePathRef.can_run

darrenhuai:

---

`ExePathRef.can_run` loops over the three execute bits, but the `break` sits outside the `if`:

```python
for key in [S_IXUSR, S_IXGRP, S_IXOTH]:
    if mode & key:
        self._can_run = True
    break
else:
    self._can_run = False
```

The loop stops after the first iteration, which costs three things:

- nothing reads `S_IXGRP` or `S_IXOTH`, so a binary that is group- or other-executable without the owner bit counts as unrunnable
- the unconditional `break` also skips the `for`/`else`, so a source without the owner bit leaves `_can_run` at `None`, and a property annotated `-> bool` returns `None`. The `# ty: ignore[invalid-return-type]` on the return line hid that
- the memoisation misses that case, so `stat()` runs again on each access

Windows shows it, since CPython sets the execute bits there only for `.exe`/`.bat`/`.cmd`/`.com`.

```
README.md                 0o100666   can_run=None   _can_run=None
.venv/Scripts/python.exe  0o100777   can_run=True   _can_run=True
```

Callers test truthiness and `None` is falsy, so today this stays latent. What changes for users is `can_run` honouring the group and other bits.

Replace the loop with a single mask test against all three bits and drop the `ty: ignore` that went with it; `ty check src/virtualenv` still passes. `tests/unit/create/via_global_ref/test_api.py` covers each execute bit, the no-bit case, and that a second read reuses the first `stat()`.


tests/unit/create/via_global_ref/builtin/test_ref.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
```
gaborbernat, line 1: Please do not create a new test file. Instead, update the existing tests or extend them where they are defined.

gaborbernat (changes requested):

tests/unit/create/via_global_ref/test_api.py
```diff
@@ -1,8 +1,38 @@
...
 from virtualenv.create.via_global_ref import api
+from virtualenv.create.via_global_ref.builtin.ref import ExePathRefToDest
 
 
 def test_can_symlink_when_symlinks_not_enabled(mocker) -> None:
     mocker.patch.object(api, "fs_supports_symlink", return_value=False)
     assert api.ViaGlobalRefMeta().can_symlink is False
+
+
+def _exe_ref(mocker, mode: int) -> ExePathRefToDest:
```
gaborbernat: Use fixtures and parametrized tests over helper methods.

gaborbernat (changes requested):

tests/unit/create/via_global_ref/test_api.py
```diff
@@ -1,8 +1,44 @@
...
+    ("exe_ref", "expected"),
+    [
+        pytest.param(0o644 | S_IXUSR, True, id="user"),
+        pytest.param(0o644 | S_IXGRP, True, id="group"),
+        pytest.param(0o644 | S_IXOTH, True, id="other"),
+        pytest.param(0o644, False, id="none"),
+    ],
+    indirect=["exe_ref"],
+)
+def test_can_run_honours_every_execute_bit(exe_ref, expected) -> None:
```
gaborbernat: Missing type annotations from all the test code 🤔

gaborbernat (changes requested):

PR #3233, opened 2026-09-10
🐛 fix(activation): restore PKG_CONFIG_PATH that was not set before

darrenhuai:
Activation prepends the environment's `lib/pkgconfig` to `PKG_CONFIG_PATH`. When the variable was unset, bash, fish and
PowerShell had no saved value to restore, so after `source env/bin/activate` and `deactivate` it stayed at
`/tmp/env/lib/pkgconfig`.

- bash and fish check whether `_OLD_PKG_CONFIG_PATH` exists (`${var+x}`, `set -q`) and unset `PKG_CONFIG_PATH` when the
  saved value is empty. PowerShell saves `$null` for an unset variable, and its existing restore removes it.
- csh ran `deactivate nondestructive` before saving, and its `test $?_OLD_PKG_CONFIG_PATH != 0 && ... || unsetenv` chain
  removed a user-set `PKG_CONFIG_PATH` on every activation. csh now saves an empty value for an unset variable, and
  `deactivate` unsets only when the saved value is empty.
- batch, fish, nushell and PowerShell add the separator only when there is a previous value, so an unset variable no
  longer becomes `C:\env\lib\pkgconfig;`.
- PowerShell hard-coded `\` and `;`, nushell `/` and `:`. Both now use the host's separators, which fixes pwsh on Linux
  and macOS and nushell on Windows.

`ActivationTester` now prints `PKG_CONFIG_PATH` before activation, while active and after deactivation, and the
`activation_tester` fixture runs every shell with the variable unset and set. xonsh does not manage the variable, so its
tester expects it unchanged. On `main` the unset cases fail for bash, fish, nushell and PowerShell, and the set cases
fail for csh and, on macOS, PowerShell.

`activate.bat` does not call `deactivate` first, so activating a second environment stacks `PKG_CONFIG_PATH`. This PR
leaves that as is.

### DIFF
--- a/docs/changelog/3233.bugfix.rst
+++ b/docs/changelog/3233.bugfix.rst
@@ -1,0 +1,4 @@
+Deactivation now unsets ``PKG_CONFIG_PATH`` when it was not set before activation, instead of leaving the environment's
+``lib/pkgconfig`` behind for good. The bash, fish and PowerShell scripts could not tell "nothing saved" apart from "was
+unset", so they restored nothing. Activation also no longer leaves a trailing separator in the batch, fish, nushell and
+PowerShell scripts when there was nothing to prepend to - by :user:`darrenhuai`.
--- a/src/virtualenv/activation/bash/activate.sh
+++ b/src/virtualenv/activation/bash/activate.sh
@@ -32,9 +32,13 @@ deactivate () {
         export TK_LIBRARY
         unset _OLD_VIRTUAL_TK_LIBRARY
     fi
-    if [ -n "${_OLD_PKG_CONFIG_PATH:-}" ]; then
-        PKG_CONFIG_PATH="$_OLD_PKG_CONFIG_PATH"
-        export PKG_CONFIG_PATH
+    if [ -n "${_OLD_PKG_CONFIG_PATH+x}" ]; then
+        if [ -n "$_OLD_PKG_CONFIG_PATH" ]; then
+            PKG_CONFIG_PATH="$_OLD_PKG_CONFIG_PATH"
+            export PKG_CONFIG_PATH
+        else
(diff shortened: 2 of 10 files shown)


darrenhuai:
The one red job, `test 3.13 - windows-2025-vs2026`, hit the 30 minute step timeout rather than failing an assertion, and I do not think it is this branch.

It stalled at 98%, in `tests/unit/seed/embed/test_bootstrap_link_via_app_data.py::test_seed_link_via_app_data[False]` — about 26 minutes between that test starting and the timeout killing the run. That file takes 25s here on this branch. Nothing in the diff outside `docs/changelog` touches seeding or app data; it is activation script templates and their tests.

The four new tests all ran in that same job before the stall:

```
[gw1] [ 60%] SKIPPED tests/unit/activation/test_bash.py::test_bash_deactivate_unsets_pkg_config_path_that_was_not_set
[gw1] [ 64%] PASSED  tests/unit/activation/test_batch.py::test_batch_deactivate_unsets_pkg_config_path_that_was_not_set
[gw1] [ 66%] SKIPPED tests/unit/activation/test_fish.py::test_fish_deactivate_unsets_pkg_config_path_that_was_not_set
[gw1] [ 73%] PASSED  tests/unit/activation/test_powershell.py::test_powershell_deactivate_unsets_pkg_config_path_that_was_not_set
```

1.1s and 1.2s for the two that run on Windows, the other two skipped there as intended. The same commit is green on 3.10, 3.11, 3.12 and 3.14 on the same Windows image, which is not what a real regression from this change would look like.

A re-run should clear it — I cannot trigger one from a fork. Happy to push a change if you would rather see it go green on a fresh run.

gaborbernat (approved):

PR #3250, opened 2026-09-17
🐛 fix(activation): escape batch quote() against injection

gaborbernat:
`BatchActivator.quote()` returned its input unchanged. A `--prompt` containing a double quote closes the `@set "VAR=value"` string early, so whatever follows runs as live cmd.exe syntax:

```console
$ virtualenv --prompt 'x" & calc & "' venv
$ venv\Scripts\activate.bat
```

runs `calc`. Confirmed on a real Windows runner, including that quoting alone isn't enough (`&` still splits the statement even with balanced quotes) and that a caller with delayed expansion enabled leaks an existing variable's value through `!NAME!` in the prompt.

`quote()` now neuters `" & | < > ^ !` and line boundaries. `(`/`)` are left alone - confirmed safe, and neutering them would break `C:\Program Files (x86)\...`. A destination or Tcl/Tk path containing `&` can't be represented at all, so batch activation is skipped for that one environment with a warning instead of silently breaking.

Tracked as [GHSA-x78j-v8h9-3j2q](https://github.com/pypa/virtualenv/security/advisories/GHSA-x78j-v8h9-3j2q).


sylvesterkaczmarek (commented):
`quote()` still leaves `!` untouched. If activation runs in a cmd.exe with delayed expansion enabled (for example `cmd /V:ON`), a prompt such as `foo!NAME!` is expanded or removed inside `@set "VIRTUAL_ENV_PROMPT=..."` instead of being preserved literally. The new property tests only cover normal cmd expansion, so this mode is missed. Please make the generated activation disable delayed expansion around these assignments or otherwise handle `!`, and add a delayed-expansion regression.

PR #3251, opened 2026-09-17
🐛 fix(seed): verify downloaded wheel digests

gaborbernat:
`download_wheel()` handed pip's output straight to the seeder with no integrity check. A compromised index or MITM'd download could substitute a different wheel under the same name.

`verify_wheel_digest()` checks the downloaded sha256 against PyPI's public JSON record for that filename. Missing record: logged, not fatal (private indexes are legitimate). Mismatch: fatal, wheel is never seeded.

Skipped entirely when `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, or `PIP_INDEX` is set - a private index can legitimately serve a different, intentionally rebuilt wheel under the same identity, and PyPI's digest isn't the right check for a wheel that never came from PyPI.

Tracked as [GHSA-94p9-xgh2-xp45](https://github.com/pypa/virtualenv/security/advisories/GHSA-94p9-xgh2-xp45).


sylvesterkaczmarek (commented):
The digest is always taken from public PyPI, even when `pip download` was deliberately configured to use a private index or mirror. A private repository can legitimately publish a rebuilt/patched wheel under the same distribution, version, and filename as PyPI; if PyPI has that filename, this code rejects the private artifact even though it came from the configured source. Please bind verification to the index that supplied the wheel (for example via index metadata/pip reporting), or only make the PyPI digest authoritative when PyPI was actually the source, and add a private-index same-filename regression.

sylvesterkaczmarek:
The environment-variable fix covers the common explicit custom-index path, but pip configuration files can still redirect the download while the new PyPI digest check remains active. I filed the remaining case as #3257.

PR #3277, opened 2026-09-21
fix(upgrade): check newly added wheel ages

gaborbernat:
The upgrade workflow checks the deleted wheel's release age and misses its untracked replacement. An upload from today can pass the seven-day gate because the old release is old enough.

Check added or modified tracked wheels and untracked wheel files. Use the selected artifact's upload timestamp rather than the first file in the release. Stop the workflow if PyPI cannot supply that artifact's metadata, so a network failure cannot bypass the gate.


Voyagerroc-Lab (commented):
Review finding (blocking)

The new parser still assumes the distribution name and version are the first two hyphen-separated fields:

    package, version, *_ = wheel.name.split("-")

Wheel filenames use the normalized project name, which may itself contain hyphens. For example, a valid artifact such as importlib-metadata-8.7.0-py3-none-any.whl is parsed as package=importlib and version=metadata, so the PyPI request becomes /pypi/importlib/metadata/json and the workflow fails (or, depending on the package, checks the wrong release). Please parse the wheel filename according to the wheel naming convention (e.g. identify the version from the right-hand tags / use a packaging parser) and add a regression test with a hyphenated distribution name.

PR #3278, opened 2026-09-21
fix(sbom): preserve format and reproducibility

gaborbernat:
The SBOM hook added in #3270 splits wheel `RECORD` rows on commas, which breaks valid CSV paths containing commas or quoted newlines. It also declares unresolved runtime dependencies as leaves, although [CycloneDX uses an empty dependsOn list to mean that a component has no dependencies](https://cyclonedx.org/guides/sbom/relationships/).

Parse `RECORD` as CSV and omit dependency entries where the transitive graph is unknown. Keep the root edges and incomplete composition, and allow those unknown relationships in the validator.

Use Hatchling's reproducible timestamp for the SBOM and recorded build input. Leave CI run identifiers in release attestations instead of embedding them in wheel content, and read the source revision from the checkout rather than a downstream builder's `GITHUB_SHA`.


Voyagerroc-Lab (approved):
Review summary

I checked the SBOM hook changes and the new fixture-based tests. CSV parsing now uses the standard CSV reader, unresolved runtime dependencies are represented without fabricated dependency entries, and Hatchling's reproducible timestamp is used consistently for both the SBOM timestamp and recorded build input. The tests cover comma, quote, newline, dependency, and timestamp cases.

The current ReadTheDocs status is failing, but it is a docs build status and the changed documentation is a one-line changelog entry; I found no implementation defect in the reviewed diff. Approving the code change while that external docs check is investigated separately.

Voyagerroc-Lab (approved):

